PR #3, opened 2015-07-21
Add data files (geographic and incumbent data)

gregbillock:
These data files come from the FCC and NOAA and reflect current
data on national boundaries, census tract designations, and
information regarding current 3.55GHz incumbent status.


gregbillock:
(reviewed by @KeyBridge)

PR #7, opened 2015-08-07
Fix coastal exclusion zone file.

gregbillock:
The KML file was actually a copy of the land-based file. Not sure how that happened, but this fixes it to the correct version.


KeyBridge:
Good catch.

PR #8, opened 2015-08-26
Add scripts and data file outputs for various SAS zones and borders.

gregbillock:
These scripts define the outlines of the SAS assignment authority
and normalized versions of NTIA-defined exclusion/protection zones
within that authority.


KeyBridge:
looks good

PR #9, opened 2015-09-29
Update README.md with code prerequisites

gregbillock:
Adds descriptions of the various platform, binary, and script configuration needed on a development system to use the scripts in the SAS repository.


KeyBridge:
add doc'n looks good

PR #10, opened 2015-10-01
Update border and protection zone scripts for accuracy.

gregbillock:
This change fixes up the border calculation to create a shapely-valid
polygon by removing a couple overlapping segments. It also uses a
better intersection method for deriving the coastal exclusion zones.
Data files are updated to correspond to the script output.


KeyBridge:
great progress

gregbillock:
Thanks. :-) For future reference, this was a @KeyBridge designed improvement using polygon intersect. :-)

PR #12, opened 2015-10-23
Add a sample certificate authority using openssl

gregbillock:
This change adds an openssl configuration file with example values
describing the various SAS certificate varieties. Example commands
are provided in the README for creating and managing such certificates.


KeyBridge:
Concur.

PR #13, opened 2015-10-28
Add scripts to retrieve files and translate FCC spreadsheet files to CSV

gregbillock:
This change adds scripts which retrieve source files from FCC, NOAA, and
NTIA. It also adds scripts to translate spreadsheet files to CSV and generate
the 3650-3700 KML file for NTIA-identified radar sites. The resulting data
files are included as part of the commit as well.


KeyBridge:
For radar site buffering: It looks like you're directly translating lat/lon between decimal and radian. We do this by shifting coordinate reference systems (WGS to Cartesian, buffer, then back). Will need to compare results and confirm match.

gregbillock:
Sounds good. I'm only doing degrees/minutes/seconds conversion and then doing everything based on decimal degrees after that. I don't think the R&O specifies that these are WGS84 coordinates, but I've been assuming we should interpret them that way.

The spherical trig isn't cartesian, but the comparison to cartesian should be quite close -- 80km circles have pretty low error at those latitudes.

I'm planning on lowering the angular discrimination a bit to make the files more wieldy. How about 96 points (5 degree intervals instead of 1 degree intervals)?

KeyBridge:
Concur that the total data size is small and doesn't warrant extreme optimization; 5-degree intervals sounds fine.

KeyBridge:
We covered this extensively in white space land….

The R&W doesn’t specify it but all FCC systems, forms and processes (except CDBS) specify and require NAD_83. If we ask that will be the answer: “NAD_83".

There exists a Federal directive stating NAD_83 and WGS_84 are interchangeable. Commission staff have generally let us use WGS_83 and NAD_83 notation interchangeably, recognizing the difference as a pedantic nit and also recognizing that commercial GPS units output WGS_84.

I have a copy of the directive somewhere if needed.

Cheers!

Jesse Caulfield, CEO
Key Bridge LLC
keybridgeglobal.com
Office: 703 414 3500
Mobile: 202 413 3635

gregbillock:
OK: I updated the data file generation to use 5-degree steps. All the coordinate math should be WGS84.

KeyBridge:
Looks good. Thank you for this!

PR #19, opened 2015-12-01
Add scripts for retrieval and extraction of FCC IBFS data.

gregbillock:
The scripts retrieve the most recent version of the IBFS database
as well as extracting the entries of interest from the dataset.
FSS locations, frequency ranges, and coordination data (pointing
data) is extracted and transformed into the resulting CSV file.


gregbillock:
@w4je can you look at this too?

PR #20, opened 2016-02-22
Add schema definitions for SAS-CBSD interface messages.

gregbillock:
This change adds schema definitions in machine-verifiable form
for the SAS-CBSD message definitions. It also includes examples
of those messages and a script to verify the correct syntax of
both the schema definitions as well as the validation of the
example messages.


jamesneel:
Good work.

PR #24, opened 2016-07-12
ITM Longley-Rice implementation with Python bindings

gregbillock:
This code is a sequence of commits which pull in the standard ITM implementation from NTIA/ITS. It then branches a copy of the C++ implementation and modifies it for more platform-independent compilation. The next commit adds a test which uses the fortran implementation test vectors. The following commits then add a python interface and a python file which illustrates and documents use of the C++ code.


w4je:
LGTM.

PR #26, opened 2016-08-18
OID Updates

clintwilson:
1. Changing ROLE OIDs
- Instead of storing the ROLE value as a separate, arbitrary x.509 extension with UTF-8 strings for each ROLE, this identifies specific OID values for each ROLE and places them in the Certificate Policies Extension.
- The OID values are now a sub of the previous ROLE OID. They use 1.3.6.1.4.1.46609.1.1.1 - 1.3.6.1.4.1.46609.1.1.6 for the current 6 ROLEs extant in the WInnForum ComSec doc.
2. Adds an OpenSSL config and script sections as necessary to create the CBSD OEM intermedate and EE cert.
3. Various text fixes in scripts and openssl.cnf.


cert/CACreationScript.sh
```diff
@@ -108,25 +108,33 @@ openssl ca -create_serial -cert $PI.crt -keyfile $PI.pkey -outdir $CSR -batch -o
...
 
-#TODO: manufacturer intermediate signed by $CBSD, and EE cert signed by manufacturer intermediate
-
+#Device OEM Intermediate
+CBSD_OEM=cbsd_oem_ca
+CBSD_OEM_SIGN=$CBSD_OEM$SIG
 
+openssl req -new -newkey rsa:$INT_RSA_KEY_SIZE -nodes -reqexts $CBSD_OEM -out $CBSD_OEM.csr -keyout $CBSD_OEM.pkey -subj "/C=US/ST=District of Columbia/L=Washington/O=Wireless Innovation Forum/OU=www.wirelessinnovation.org/CN=WInnForum RSA CBSD OEM CA-1" -config openssl.cnf
 
+openssl ca -create_serial -cert $CBSD.crt -keyfile $CBSD.pkey -outdir $CSR -batch -out $CBSD_OEM.crt -utf8 -days $INT_VALIDITY -policy policy_anything -md sha384 -extensions $CBSD_OEM_SIGN -config openssl.cnf -in $CBSD_OEM.csr
```
gregbillock: Should this be -keyfile $CBSD_OEM.pkey?

cert/CACreationScript.sh
```diff
@@ -108,25 +108,33 @@ openssl ca -create_serial -cert $PI.crt -keyfile $PI.pkey -outdir $CSR -batch -o
...
 CBSD_EE_SIGN=$CBSD_EE$SIG
 
-openssl req -new -newkey rsa:$EE_RSA_KEY_SIZE -nodes -reqexts $CBSD_EE -out $CBSD_EE.csr -keyout $CBSD_EE.pkey -subj "/C=US/ST=District of Columbia/L=Washington/O=Wireless Innovation Forum/OU=www.wirelessinnovation.org/CN=Device End-Entity Example" -config openssl.cnf
+openssl req -new -newkey rsa:$EE_RSA_KEY_SIZE -nodes -reqexts $CBSD_EE -out $CBSD_EE.csr -keyout $CBSD_EE.pkey -subj "/C=US/ST=District of Columbia/L=Washington/O=Wireless Innovation Forum/OU=www.wirelessinnovation.org/CN=CBSD End-Entity Example" -config openssl.cnf
 
 openssl ca -create_serial -cert $CBSD.crt -keyfile $CBSD.pkey -outdir $CSR -batch -out $CBSD_EE.crt -utf8 -days $EE_VALIDITY -policy policy_anything -md sha384 -extensions $CBSD_EE_SIGN -config openssl.cnf -in $CBSD_EE.csr
 
-#TODO: manufacturer intermediate signed by $CBSD, and EE cert signed by manufacturer intermediate
-
+#Device OEM Intermediate
```
gregbillock: Do you think we should have a separate intermediat e for OEM EE? Or just use the device intermediate? My mental image was to just use the cbsd_ca for this. So IOW there'd still be the OEM cert (which should have ca:true and keydepth=1 in the profile), but it would be off a cbsd_ca intermediate. You think that's sufficient separation?

cert/openssl.cnf
```diff
@@ -135,260 +140,366 @@ caIssuers;URI.0 = http://cacerts.digicert.com/WinnForumCA.crt
...
 keyUsage = critical, digitalSignature, cRLSign, keyCertSign
+certificatePolicies=@cps,ROLE_CA,ROLE_OPERATOR
 
-ROLE=ASN1:SEQUENCE:oper_seq
+[ cps ]
 
-[ oper_seq ]
-field1 = UTF8:OPERATOR
-field2 = UTF8:CA
+policyIdentifier = 2.16.840.1.114412.2.1
```
gregbillock: Is this a digicert OID? Is this a placeholder for a WINNF OID and policy link we should make?

cert/openssl.cnf
```diff
@@ -135,260 +140,366 @@ caIssuers;URI.0 = http://cacerts.digicert.com/WinnForumCA.crt
...
+[ sas_ca ]
 
 subjectKeyIdentifier = hash
 #authorityKeyIdentifier = keyid:always,issuer
 basicConstraints = critical, CA:true, pathlen:0
 keyUsage = critical, digitalSignature, cRLSign, keyCertSign
+certificatePolicies=@cps,ROLE_CA,ROLE_SAS
 
-ROLE=ASN1:SEQUENCE:sas_seq
+[ cps ]
```
gregbillock: I'm not sure I understand the repeated [cps] sections. Are they basically sub-sections per cert type?

cert/openssl.cnf
```diff
@@ -135,260 +140,366 @@ caIssuers;URI.0 = http://cacerts.digicert.com/WinnForumCA.crt
...
+ CPS.1="https://www.digicert.com/CPS"
 
-[ wif_cbsd_ca_sign ]
+[ cbsd_oem_ca_sign ]
 
 subjectKeyIdentifier = hash
 authorityKeyIdentifier = keyid:always,issuer
 basicConstraints = critical, CA:true, pathlen:0
 keyUsage = critical, digitalSignature, cRLSign, keyCertSign
+certificatePolicies=@cps,ROLE_CA,ROLE_CBSD
```
gregbillock: So is this putting the extension within the policy standard section? That seems like a great plan if there's a hook there, since that's exactly what these are.

cert/CACreationScript.sh
clintwilson: I think that's how I have it configured currently.
CBSD_CA is issued from the ROOT and has a pathlength of 1. It can issue either CBSD_EE certs or CBSD_OEM_CA certs.
CBSD_OEM_CA is issued from CBSD_CA and has a pathlength of 0. It can issue only CBSD_OEM_EE certs (which are functionally equivalent to CBSD_EE certs).
So, for devices, there's a common intermediate that can either directly issue Device certs _or_ issue an additional signing intermediate so OEMs can issue their own Device certs.

cert/CACreationScript.sh
clintwilson: This is a CA issued by the CBSD Intermediate, so it's being signed by the CBSD crt/pkey files (the only CA cert _not_ issued directly by the Root).

cert/openssl.cnf
clintwilson: Right now the different [cps] sections are superfluous, but if different CAs operated different parts of the infrastructure, each profile may have its own CPS/CP which applies to it so I put in a [cps] section for each profile so they could be edited independently in the future.

cert/openssl.cnf
clintwilson: Yeah, just a place holder using DigiCert's general CPS OID. 
If WINNF were to have a global policy OID, this could be replaced by that. Most likely each CA would also want to include their specific, applicable policy as well. 
For a WINNF global policy OID, I think I would propose using something like 1.3.6.1.4.1.46609.2.1. WINNF owns the entire "1.3.6.1.4.1.46609" structure, correct?

cert/openssl.cnf
clintwilson: Yes, exactly. Since the ROLE values really represent an external policy on the certs, I think including those in the CP extension fits best. Each of the intermediate CAs has the CA Role and a profile-specific Role. Each of the end-entities has just the profile-specific Role.

cert/CACreationScript.sh
gregbillock: I see. OK. So they're basically on the same path but just have a bit different profiles for configurability. Sounds good.

cert/CACreationScript.sh
gregbillock: Got it. Thanks.

cert/openssl.cnf
gregbillock: I'm not an openssl config expert; do the section names scope to the closest ancestor like this? My mental image was they'd need to be called "[cps_ca]" and stuff like that to distinguish.

cert/openssl.cnf
gregbillock: Yes, that's our org number we got from the IANA, so we should be good for the whole sub-space. Your proposal looks good to me.

gregbillock:
I think this is good to go. I still have a couple questions about the [cps] sections and I think we hope to change the policy OID later. We can figure that out in subsequent work though.

LGTM, ready to check in. Thanks, Clint!

PR #28, opened 2016-09-28
Modify mdvar parameter from 12 to 3.

gregbillock:
This change was discussed with NIST in Sept 2016. The value for
this parameter as used by NTIA in their model was 3 instead of 12.


w4je:
LGTM

PR #35, opened 2016-12-22
First version of test harness.

grao1991:


grao1991:
Fixed.

src/harness/fake_sas.py
```diff
@@ -0,0 +1,154 @@
+"""A fake implementation of SasInterface, based on v1.0 of the SAS-CBSD TS.
```
gregbillock: Can you include the winnforum licensing headers on the .py files? You can copy from other .py files in the repository.

grao1991: Done

PR #39, opened 2017-01-03
Only fetch US tiles of NED. Script to retile NLCD data files.

gregbillock:
Substantial filtering added to only retrieve needed US-intersecting and nearby NED tiles.
Added a script to retile the large NLCD datasets to more manageable tiled versions.


w4je:
Looks groovy

PR #41, opened 2017-01-04
NED extraction update to create fewer files

gregbillock:
Also moves .gitignore from main directory to data directory.


w4je:
LGTM

PR #55, opened 2017-01-19
Change mdvar to 13

gregbillock:
This follows discussions in Winnf Propagation TG on ITM parameters
and the handling of variability in the use we plan to make of the
model. See R2-SGN-03.


w4je:
Awesome.

PR #57, opened 2017-01-19
NLCD lookup code for finding land use codes for a lat/lng.

gregbillock:


w4je:
LGTM based on cursory inspection.

PR #58, opened 2017-01-23
Fix bug in NLCD indexer on coordinate transformation.

gregbillock:
Adds translation ability from NLCD code to WG1 Prop classifications.
Also adds the ability for the NED indexer to prepare elevation
profiles.


w4je:
LGTM

PR #64, opened 2017-01-30
Lookup functions for climate and refractivity data

gregbillock:
These scripts provide lookup functionality for the tropo climate and surface refractivity ITU data.


w4je:
LGTM

PR #65, opened 2017-02-01
Fix NLCD indexer tile containment checks.

gregbillock:
Also make changes to decrease printouts from other lookup
utilities and add Python module indicator files.


w4je:
LGTM

PR #67, opened 2017-02-02
Patch 1

frbg69:


grao1991:
It seems that there are many things in this pull request. Can you split it into small pieces and send out the test harness part first?

frbg69:
The submitted code can be viewed in 4 parts:
1) **harness/cbas_sas_harness/** contains 3 files: **sas_interface.py**, **sas_connector.p**y and **cbsd_to_sas.py** this is the test harness itself that allows communication with SAS SUT.
2) **common/cbsd.py** is the cbsd to sas response message library for verifying and validating the incoming test harness response provided by the SAS SUT.
3) **tests/** contains the tests script **test_cbsd_registration.py** where the test cases have been written.
4) **test/testdata** contains the cbsd folder for each of the protocol messages and under each folder the corresponding json files used for building the cbsd request messages.

liguanghua:
Hi, 
It would be great if you can split this pull request: one for the first and second part (related to test harness itself); another for the third and fourth part (test script and data) but it doesn't need to contain all test cases at once.
The smaller amount of changes would be much easier for review.

Thanks!

frbg69:
I have split the PR into patch-1 and patch-2
**patch-1** contains parts 1 and 2
**patch-2** contains parts 3 and 4

src/harness/cbsd_sas_harnes/sas_interface.py
```diff
@@ -0,0 +1,422 @@
...
+    'RSA-AES256-GCM-SHA384',
+    'ECDHE-ECDSA-AES128-GCM-SHA256',
+    'ECDHE-ECDSA-AES256-GCM-SHA384',
+    'ECDHE-RSA-AES128-GCM-SHA256'])
+
+
+# Classes
+
+
+class CConfigParser(ConfigParser.ConfigParser):
```
grao1991: what's the difference between this class and using ConfigParser directly (like https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/sas.py line 32)?

src/harness/cbsd_sas_harnes/sas_interface.py
```diff
@@ -0,0 +1,422 @@
...
+        conn.setopt(conn.URL, url.encode('utf8'))
+        conn.setopt(conn.WRITEFUNCTION, response.write)
+        conn.setopt(conn.HTTPHEADER, header)
+
+        # Diagnostics settings
+        conn.setopt(conn.VERBOSE, 1)
+        #conn.setopt(conn.DEBUGFUNCTION, self.debug_function)
+
+        # Setup TLS 1.2
+        if(tls):
```
grao1991: TLS is required by TS, I think we might not need this flag.

src/harness/cbsd_sas_harnes/sas_interface.py
```diff
@@ -0,0 +1,422 @@
...
+
+        self.logger.debug('Http:     {}'.format(self.GetHttp()))
+        self.logger.debug('Https:    {}'.format(self.GetHttps()))
+        self.logger.debug('Config:   {}'.format(self.cfgFile))
+        self.logger.debug('CA Cert:  {}'.format(self.caCert))
+        self.logger.debug('Cert:     {}'.format(self.cert))
+        self.logger.debug('Key:      {}'.format(self.key))
+        self.logger.debug('Password: {}'.format(self.keypasswd))
+
+    def Request(self, pull, tls, url, request, ssl_caCert, ssl_cert, ssl_key):
```
grao1991: we already have code to do the similar thing checked in.

src/common/cbsd.py
```diff
@@ -0,0 +1,934 @@
...
+    'measReportConfig') + PARAMETER_RESPONSE + PARAMETER_OPERATIONPARAM
+MESSAGE_PARAMETERS_RELINQUISHMENT = (
+    'relinquishmentResponse',
+    'cbsdId',
+    'grantId') + PARAMETER_RESPONSE
+
+# Classes
+
+
+class CBSD(object):
```
grao1991: What are the benefits to define a lot of assert functions comparing to using assertEqual directly?

src/tests/test_cbsd_registration.py
```diff
@@ -0,0 +1,1087 @@
...
+            response = self.req.RegistrationRequest(devices)
+
+            # Check registration response
+            self.res.assertRegistrationResponse(
+                response,
+                [{"response": {"responseCode": 0}},
+                 {"response": {"responseCode": 0}},
+                 {"response": {"responseCode": 0}}])
+
+        except Exception as e:
```
grao1991: It would be better if we don't repeat try...except in every test cases.

src/harness/cbsd_sas_harnes/cbsd_to_sas.py
```diff
@@ -0,0 +1,383 @@
...
+CATEGORY_ERROR = 202
+UNSUPPORTED_SPECTRUM = 300
+INTERFERENCE = 400
+GRANT_CONFLICT = 401
+TOO_MANY_GRANTS = 402
+TERMINATED_GRANT = 500
+SUSPENDED_GRANT = 501
+UNSYNC_OP_PARAM = 502
+# Response Code resources
+RESPONSE_CODE = {
```
grao1991: Maybe we can get rid of this dict.

src/harness/cbsd_sas_harnes/cbsd_to_sas.py
```diff
@@ -0,0 +1,383 @@
...
+#INI_FILE_SECTION = 'CBSD-SAS'
+#INI_FILE_OPTIONS = {
+#    'CAT_A': None,
+#    'CAT_B': None,
+#    'SPECTRUM_GAA': None,
+#    'SPECTRUM_PAL': None,
+#    'GRANT': None,
+#    'HEARTBEAT': None}
+# CBSD to SAS Resources
+RESOURCES = {
```
grao1991: Is it possible to not define this dict?

src/harness/cbsd_sas_harnes/cbsd_to_sas.py
```diff
@@ -0,0 +1,383 @@
...
+        'res': 'heartbeatResponse',
+        'ini': 'HEARTBEAT'},
+    'rel': {
+        'name': 'RELINQUISHMENT',
+        'path': '/relinquishment',
+        'req': 'relinquishmentRequest',
+        'res': 'relinquishmentResponse',
+        'ini': 'RELINQUISHMENT'}}
+# Response Code values
+SUCCESS = 0
```
frbg69: 1. Please look at new branch 'patch-3' where I have made the 'merge' in terms of not affecting our already running test scripts
2. Removed code never used

Ok on the constant values as mentioned

src/harness/cbsd_sas_harnes/sas_interface.py
frbg69: I had modified it to interpolate options defined both within a section and between sections, since we are now using the new config based upon the agreement, I have discarded the iterpolation/read code and left it smaller, so as to preserv the option 'case format' which allows me to not to have to move more code than what is needed.

src/harness/cbsd_sas_harnes/sas_interface.py
frbg69: Good observation, have made the needed changes.

src/harness/cbsd_sas_harnes/sas_interface.py
frbg69: I have simplified it accordingly, please look at the patch-3 branch.

src/common/cbsd.py
frbg69: Answering your question: it is not enough to only check that the correct value comes in, we need to verify that other parameters that should also accompany or should not come at all, are present or not and thus comply with the protocol spec exactly as it is written, no more nor less.
To also complement the previous answer: there is only one place where we know we verify each message, and follow the protocol as to: verify that we receive the required parameters, and the values and types are correct, then the same for the conditional parameters and at the end the optional parameters, if any.
We also make sure, no additional parameters arrive, and if they do so we raise a warning.
Also, if per each combination of expected responseCode value we were to check all the parameter combinatios as per the spec, and so and so.... it would make each test case both to large and messy and prone to error.
And if you look inside the code, you would see we do use the assertEqual that comes with unittest, as well as various other unittest provided assertions, accordingly.

src/harness/cbsd_sas_harnes/cbsd_to_sas.py
frbg69: I use it extensibly in all the request messages within this module, I rather not have to remove it.

src/harness/cbsd_sas_harnes/cbsd_to_sas.py
frbg69: Removed!

PR #68, opened 2017-02-02
Fix bug in eHata rolling hills correction; add py ITM wrapper

gregbillock:
Also add changes to ITM wrapper to do better bounds checking
and bad values detection, and accept output from cases where
warnings are found with parameter values (ITM marks a pretty
wide variety of inputs as resulting in less accurate output.)


w4je:
LGTM

PR #70, opened 2017-02-07
Rename retrieve_ned_files.py to retrieve_ned_nlcd_files.py

w4je:
Add nlcd to file name to make clear that it downloads both NED and NLCD data sets.


gregbillock:
lgtm

PR #71, opened 2017-02-07
Add a model for the integrated NTIA prop model.

gregbillock:
This model incorporates adjustments to the ITM and eHata models
from WG1 Propagation task group, and combines the models using
the methods in NTIA TR 15-517.


w4je:
Line 137 of model.py, the two '>' should be '>='.

Otherwise this looks OK, but a lot of little places to go wrong.

gregbillock:
OK, fixed inequalities

w4je:
LGTM

PR #72, opened 2017-02-07
Add support for gdal version 1.x

w4je:


src/geo/nlcd_indexer.py
```diff
@@ -108,8 +109,17 @@ def IndexCoords(self, lat, lng):
     #print '  coord_bounds=', self.coord_bounds
     #print '  inv_txf=', self.inv_txf
     #print '  txf=', self.txf
-    x = self.inv_txf[0] + self.inv_txf[1] * coord[0] + self.inv_txf[2] * coord[1]
-    y = self.inv_txf[3] + self.inv_txf[4] * coord[0] + self.inv_txf[5] * coord[1]
+    
+    # The format of the return values from inv_txf (gdal.InvGeoTransform) changed from
+    # gdal version 1 to version 2
+    gdal_version = osgeo.gdal.__version__
```
gregbillock: Can you move this to line 34 instead? That is, if version == 1: self.inv_txf = self.inv_txf[1] or something.

PR #73, opened 2017-02-08
Patch 3

grao1991:


src/harness/cbsd_to_sas.py
```diff
@@ -0,0 +1,271 @@
...
+            utc = timezone('UTC')
+            return(utc.localize(datetime.utcnow()))
+
+        except Exception as e:
+            self.logger.error('Time Zone error: {}'.format(e))
+            sys.exit(1)
+
+    # Message Requests
+
+    def RegistrationRequest(self, reg_obj):
```
grao1991: can you reuse code which has already been submitted into github?

src/harness/sas_connector.py
```diff
@@ -0,0 +1,68 @@
...
+# Log format
+LOG_FORMAT = logging.Formatter(
+    '%(asctime)-15s - SasConnector - '
+    '%(funcName)s - %(levelname)s - %(message)s')
+
+
+# Classes
+
+
+class SasConnector(object):
```
grao1991: I think we can also get rid of this class and put these function into another class or in config file.

src/harness/testcases/registration_testcase.py
```diff
@@ -1,92 +0,0 @@
...
-#    See the License for the specific language governing permissions and
-#    limitations under the License.
-import json
-import os
-import unittest
-
-import sas
-
-
-class RegistrationTestcase(unittest.TestCase):
```
grao1991: can you not delete this file, but adding your test cases into this file?

src/harness/sas_connector.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env python
+# -- encoding: utf-8 --
+
+__author__ = "Francisco Benavides (francisco.benavides@federatedwireless.com)"
```
grao1991: We might not want to put author information here since other people might also work on this file.

We might need to put apache license text in every file.

src/harness/cbsd_to_sas.py
```diff
@@ -0,0 +1,271 @@
...
+        if(not isinstance(reg_obj, list)):
+            reg_obj = [reg_obj]
+        req_array[RESOURCES['dereg']['req']] = [
+            {'cbsdId': r['cbsdId']} if 'cbsdId' in r else {} for r in reg_obj]
+
+        # Send request, get response
+        results = self.SendMessage(PUSH, RESOURCES['dereg'], req_array)
+        return(results[RESOURCES['dereg']['res']])
+
+    def SpectrumInquiryRequest(self, reg_obj, spec_obj):
```
grao1991: Is it possible to follow what we have in src/harness/sas_interface.py, which seems to be easier to read.

src/tests/registration_test.py
```diff
@@ -0,0 +1,801 @@
...
+        devices = json.load(
+            open(os.path.join(
+                os.getcwd(),
+                'tests', 'testdata', 'registration', 'device1_b.json')))
+
+        # Register
+        response = self.req.RegistrationRequest(devices)
+
+        # Check registration response
+        self.res.assertRegistrationResponse(
```
grao1991: I think we agreed to not introduce extra assert function in today's meeting.

frbg69: I missed it!

src/harness/cbsd_to_sas.py
frbg69: just changing the name impacts all our already implemented test scripts, so its best to keep this.

src/harness/sas_connector.py
frbg69: Agreed, changing all files with this

src/harness/cbsd_to_sas.py
frbg69: Also, in this way, we can add a sas_to_sas.py module for the SAS to SAS testing, reusing the sas_interface module

src/harness/testcases/registration_testcase.py
frbg69: We already have all the test cases, prefer to keep them since they were already working

src/harness/sas_connector.py
frbg69: The original idea came from my predecessor, and I kept it since it works just as it is without adding any additional complexity. I think we can keep.

src/harness/cbsd_to_sas.py
frbg69: seem my previous response further down on the same.

src/harness/sas_connector.py
frbg69: We have made further analysis and will be merging more code to yours in order to re-use what you already put in place.

PR #76, opened 2017-02-14
Support fcc id injection.

grao1991:


grao1991:
Added the missing on in grant_testcase.py

qingwu123:
Sorry I still don't understand. 

If the goal is to inject fcc id only after _sas_admin.Reset() is called, then it should be called only once for every testcase file, that's not the case.

If the goal is to inject fcc id before each _sas.Registration() call, then that's not the case for registration_testcase.py and heartbeat_testcase.py(missing one).

If the goal is to inject fcc id before _sas.Registration() call for successful test cases only, then that's not the case for heartbeat_testcase.py and grant_testcase.py.

grao1991:
Should be the second case. Fixed now, please check.

src/harness/testcases/registration_testcase.py
```diff
@@ -67,6 +69,7 @@ def test_10_3_4_2_5_1(self):
     # Register the device, make sure at least one required parameter is invalid
     device_a = json.load(
         open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
```
qingwu123: I think you want to switch line 72 and line 73. I realized it after I approved the pull request. Can submit the fix with next pull request.

PR #79, opened 2017-02-21
Add CBSD-SAS registration test 10.3.4.1.1.2

janetliuboston:


grao1991:
Are we working on this test case (text copied from winnforum document)?

1. Make sure no CBSD ID exists in SAS for the CBSD being tested. All
Conditional parameters for the CBSD B should already be pre-loaded
into SAS. The measCapability contains no value. Ensure that CBSD
Test Harness or certified CBSD sends correct Registration request to
SAS: valid userId, fccId, and cbsdSerialNumber.
2. SAS approves the request by sending a CBSD Registration Responseas follows:
 SAS response includes a valid CBSD ID.
 measReportConfig shall not be included
 The responseCode in Response Object should be 0, concluding an approved registration

If so, I think your code is different from winnforum document. The document says in step 1 we want to pre-loaded all conditional parameters into SAS instead of providing them in registration request.

janetliuboston:
Agreed. 10.3.4.1.1.1 will also need to be modified to pre-load all conditional parameters into SAS.

grao1991:
I will propose some testing interface to pre-load conditional parameters.

grao1991:
Can you resolve conflicts?

janetliuboston:
Merges have been resolved.

PR #82, opened 2017-02-25
Add rural vegetative loss

w4je:
Added 0-15 dB random loss to ITM in rural areas per 15-517, A.2.

### DIFF
--- a/src/prop/model.py
+++ b/src/prop/model.py
@@ -19,6 +19,7 @@
 import math
 import sys
 import os
+import random
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
 from ehata import ehata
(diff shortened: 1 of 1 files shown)


src/prop/model.py
```diff
@@ -136,6 +137,8 @@ def PropagationLoss(self, f, lat1, lng1, h1, lat2, lng2, h2, land_cat=''):
       
     if land_cat == 'RURAL' or h1 >= 200 or h2 >= 200:
       itm_loss = self.ITM_AdjustedPropagationLoss(lat1, lng1, h1, lat2, lng2, h2, f, 0.5)
+      if land_cat == 'RURAL':
```
gregbillock: We don't want losses to be random -- that means different applications will get different answers. We set vegetative loss to 0 in prop TG, as with other clutter losses.

gregbillock: (If this didn't make it into the requirements we should add it.)

w4je: Okeydokey. Although, we don't actually set other clutter losses to zero, they're built deterministically into eHata, but I agree, if we add random losses it's impossible to compare on a point-to-point basis.

gregbillock:
Do you want to put this behind like a "simulate exclusion zone" flag or something? That same flag would need things like randomized indoor penetration loss rather than a fixed value as well.

w4je:
We will not implement this factor. A proposed note should be added to R2-SGN-04 to this effect.

PR #83, opened 2017-02-25
Added effective heights calculation

w4je:
The comparison to 200 m is based on effective heights instead of actual heights. Modified code accordingly. ***Note: profile[] is now calculated as many as three times in this code for each path. Should we just declare it as a global variable and calculate once at the beginning, or implement another way to do it so we're not creating the same profile three times?

### DIFF
--- a/src/prop/model.py
+++ b/src/prop/model.py
@@ -133,8 +133,12 @@ class PropagationLossModel:
         code = self.nlcdIndx.NlcdCode(lat1, lng1)
       land_cat = land_use.NlcdLandCategory(code)
     print 'Using land_cat =', land_cat
-      
-    if land_cat == 'RURAL' or h1 >= 200 or h2 >= 200:
+
+#   Calculate effective heights of tx and rx:
+    profile = self.nedIndx.Profile(lat1, lng1, lat2, lng2)
+    h1eff, h2eff = EffectiveHeights(h1, h2, profile)
+    
+    if land_cat == 'RURAL' or h1eff >= 200 or h2eff >= 200:
       itm_loss = self.ITM_AdjustedPropagationLoss(lat1, lng1, h1, lat2, lng2, h2, f, 0.5)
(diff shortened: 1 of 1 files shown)


src/prop/model.py
```diff
@@ -133,8 +133,12 @@ def PropagationLoss(self, f, lat1, lng1, h1, lat2, lng2, h2, land_cat=''):
...
       land_cat = land_use.NlcdLandCategory(code)
     print 'Using land_cat =', land_cat
-      
-    if land_cat == 'RURAL' or h1 >= 200 or h2 >= 200:
+
+#   Calculate effective heights of tx and rx:
+    profile = self.nedIndx.Profile(lat1, lng1, lat2, lng2)
+    h1eff, h2eff = EffectiveHeights(h1, h2, profile)
+    
+    if land_cat == 'RURAL' or h1eff >= 200 or h2eff >= 200:
```
gregbillock: It is true there's a lot of profile recalc. The most painful is the 80km thing, but the code doesn't have all the profile-passing you might want, either, so it ends up getting recalculated.

We could do a couple things about this. One is instead of 80km we could use profile sub-sets to the nearest 80km, which would save some recalc.

Is there a spot lower where you can eliminate a profile recalc by pulling it out here?

src/prop/model.py
```diff
@@ -133,8 +133,12 @@ def PropagationLoss(self, f, lat1, lng1, h1, lat2, lng2, h2, land_cat=''):
         code = self.nlcdIndx.NlcdCode(lat1, lng1)
       land_cat = land_use.NlcdLandCategory(code)
     print 'Using land_cat =', land_cat
-      
-    if land_cat == 'RURAL' or h1 >= 200 or h2 >= 200:
+
```
gregbillock: Do you want to add the min-height=20 above? Should it go around line 89? Put a TODO at least...

PR #86, opened 2017-02-25
Implement minimum CBSD height of 20 m

w4je:
Applies to all eHata calculations per R2-SGN-04

### DIFF
--- a/src/prop/model.py
+++ b/src/prop/model.py
@@ -127,6 +127,8 @@ class PropagationLossModel:
 
   # This is the oracle for propagation loss from point 1 to point 2 at frequency f (Mhz).
   def PropagationLoss(self, f, lat1, lng1, h1, lat2, lng2, h2, land_cat=''):
+  # Note: lat1/lng1/h1 _must_ correspond to the CBSD for proper implementation of R2-SGN-04
+
     if land_cat == '':
       code = self.nlcdIndx.NlcdCode(lat2, lng2)
       if code == 11:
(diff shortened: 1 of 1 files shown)


src/prop/model.py
```diff
@@ -127,24 +127,28 @@ def LandClassification(self, lat, lng):
...
       land_cat = land_use.NlcdLandCategory(code)
     print 'Using land_cat =', land_cat
 
-#   Calculate effective heights of tx and rx:
+    # Note: R2-SGN-04 calls for a minimum CBSD height of 20 m for all eHata calculations.
+
+    # Calculate effective heights of CBSD and rx:
     profile = self.nedIndx.Profile(lat1, lng1, lat2, lng2)
-    h1eff, h2eff = EffectiveHeights(h1, h2, profile)
+    h1eff, h2eff = EffectiveHeights(max(h1,20.), h2, profile)
```
gregbillock: Is the structure height min=20, or is the effective height min=20? 0112 says h_b is the effective height, so I think this should be

h1eff = max(h1eff, 20.0)

w4je: It's not clear. I honestly don't think SGN-04 meant to say that h_b should be h_effective. Maybe they did, but look at it this way -- it then says that if you have two devices separated by 50 m, then the CBSD side used effective height and the other side uses terrain site, in the slant-range calculation. That makes no sense. There's no way that "effective height" on one end could have any influence on FSL between two devices on the opposite side of a street from each other. That's silly.

gregbillock:
I don't see any slant-range calculation there. It's just "CBSD-receiver distance". So there's no accounting for either a 1km tower or a weird effective-height situation.

My take is that this is because these short ranges aren't really that well managed by ehata, so we're basically putting a low end on.

gregbillock:
Do you want to commit this?

src/prop/model.py
w4je: According to Paul it's both. So I think it should be h1eff = max(EffectiveHeights(max(h1,20.)),20.)

PR #87, opened 2017-02-27
Define PreloadRegistrationData interface for multi step registration testing

grao1991:


janetliuboston:
How is this pre-load registration interface different from a registration? Will we be creating a test path separate from actual running path?

grao1991:
They can not share the same url because the behavior is slightly different. But I try to reuse the SAS-CBSD message which seems to be enough.

janetliuboston:
Sounds good.

PR #90, opened 2017-03-01
Add deregistration test case WINFF.FT.S.DER.5

qingwu123:


src/harness/testcases/deregistration_testcase.py
```diff
@@ -75,3 +75,41 @@ def test_10_15_4_2_1(self):
...
+
+  def test_10_15_4_3_4_1(self):
+    """CBSD ID value invalid: single request object.
+
+    CBSD sends deregistration request to SAS in which the CBSD ID is not a
+    string, but an arbitrary number. The response should be FAIL.
+    """
+
+    # Register the device
+    device_a = json.load(
```
qingwu123: Agreed. But looks like the test case was removed from the latest test plan, so I removed the test case. Please review. Thanks.

src/harness/testcases/deregistration_testcase.py
```diff
@@ -77,4 +77,21 @@ def test_10_15_4_2_1(self):
...
+    CBSD sends deregistration request when its CBSD ID is not in SAS,
+    the response should be FAIL.
+    """
+
+    # Deregister the device before registration, thus the CBSD ID does not exist in SAS
+    request = {'deregistrationRequest': [{'cbsdId': 'A nonexistent cbsd id'}]}
+    response = self._sas.Deregistration(request)['deregistrationResponse'][0]
+    # Check the deregistration response
+    self.assertFalse('cbsdId' in response)
+    self.assertEqual(response['response']['responseCode'], 103)
```
grao1991: Is it allowed to return 105?

qingwu123: Forgot to verify against the latest version. Will fix this. Thanks.

qingwu123: Actually 105 doesn't make sense to me though for both test cases(DER.3 and DER.5). For DER.3, cbsdId is not even provided, so how could SAS return 105. For DER.5, cbsdId is nonexistent, so SAS should not return 105. Should we request test plan be modified?

axsahoo:
It is possible that SAS can return 105 in DER_3 and DER_5. From the TLS connection (https), SAS may be able to identify the CBSD (e.g., by looking at the IP address of the connection) and deregisters the CBSD by returning 105. A CBSD which is sending such a bad request (DER_3 or 5) is probably in a bad state and hence the SAS may send 105 if it can identify the CBSD (as explained above).

thanks and regards

-Anirud

Anirudha (Anirud)  Sahoo
National Institute of Standards and Technology
Wireless Networks Division
Communications Technology Lab
100 Bureau Drive, Stop 6730
Gaithersburg, MD  20899-8920
Ph: 301-975-4439

qingwu123:
IP address is not permanent. Does SAS really want to associate a CBSD with an ip address?

axsahoo:
This part was discussed in WinnForum. IP address is one way of identifying the CBSD. There could be other ways. The bottom line is that if the SAS could identify the CBSD, then it should be allowed to send 105. So the test case description in the TS is correct.

thanks and regards

-Anirud

Anirudha (Anirud)  Sahoo
National Institute of Standards and Technology
Wireless Networks Division
Communications Technology Lab
100 Bureau Drive, Stop 6730
Gaithersburg, MD  20899-8920
Ph: 301-975-4439

qingwu123:
OK. Thanks for clarifying Anirud.

src/harness/testcases/deregistration_testcase.py
qingwu123: Fixed. Please review

PR #93, opened 2017-03-08
add SAS-CBSD registration test 10.3.4.1.2.1

janetliuboston:
Single-Step registration (Cat A CBSD with no existing CBSD ID)

### DIFF
--- a/src/harness/testcases/registration_testcase.py
+++ b/src/harness/testcases/registration_testcase.py
@@ -43,6 +43,24 @@ class RegistrationTestcase(unittest.TestCase):
     response = self._sas.Registration(request)['registrationResponse'][0]
     # Check registration response
     self.assertEqual(response['response']['responseCode'], 0)
+    
+  @winnforum_testcase
+  def test_10_3_4_1_2_1(self):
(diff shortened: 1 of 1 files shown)


src/harness/testcases/registration_testcase.py
```diff
@@ -43,6 +43,24 @@ def test_10_3_4_1_1_1(self):
     response = self._sas.Registration(request)['registrationResponse'][0]
     # Check registration response
     self.assertEqual(response['response']['responseCode'], 0)
+    
+  @winnforum_testcase
+  def test_10_3_4_1_2_1(self):
```
janetliuboston: I think this is correctly named 10_3_4_1_2_1 Single-Step registration (Cat A with no existing CBSD ID)?

kate-harrison: Shall we just switch to using the new names now? I think this one should be WINFF.FT.S.REG.6.

janetliuboston: Yes, thanks.

janetliuboston:
Updated test case name to test_WINFF_FT_S_REG_6.

PR #96, opened 2017-03-17
added code in "fake_sas.py" to handle the deregistration missing "cbs…

axsahoo:
…dId" parameter

test case. Added a new file "__init__.py" in the "testcases" subdirectory.
Without this file any individual testcase (using "-m unittest" option) could
not be run.

### DIFF
--- a/src/harness/fake_sas.py
+++ b/src/harness/fake_sas.py
@@ -11,6 +11,36 @@
 #    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
+
+
+# Some parts of this software was developed by employees of 
+# the National Institute of Standards and Technology (NIST), 
+# an agency of the Federal Government. 
+# Pursuant to title 17 United States Code Section 105, works of NIST employees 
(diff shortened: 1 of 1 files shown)


src/harness/fake_sas.py
```diff
@@ -35,6 +65,8 @@
     'AES128-GCM-SHA256', 'AES256-GCM-SHA384', 'ECDHE-RSA-AES128-GCM-SHA256'
 ]
 
+MISSING_PARAM = 102
+INVALID_PARAM = 103
```
kate-harrison: FYI in other files, we've been using the number directly rather than defining a constant. I think either is fine, though (no change needed).

src/harness/fake_sas.py
```diff
@@ -110,15 +142,23 @@ def Relinquishment(self, request, ssl_cert=None, ssl_key=None):
   def Deregistration(self, request, ssl_cert=None, ssl_key=None):
     response = {'deregistrationResponse': []}
     for req in request['deregistrationRequest']:
-      response['deregistrationResponse'].append({
+      if (req.get('cbsdId') == None) :
```
kate-harrison: I think this can be written as:

if 'cbsdId' not in req

PR #97, opened 2017-03-17
add relinquishment test case WINFF_FT_S_RLQ_4

qingwu123:


src/harness/testcases/relinquishment_testcase.py
```diff
@@ -123,3 +124,25 @@ def test_10_13_4_1_2(self):
...
+
+    CBSD Harness sends Relinquishment Request to SAS including CBSD ID and
+    Grant ID in correct format but the CBSD ID does not exist in SAS.
+    The response should be FAIL.
+    """
+
+    # Relinquish the grant
+    request = {
+        'relinquishmentRequest': [{
+            'cbsdId': 'A non-exist cbsd id',
```
kate-harrison: Please change "non-exist" to "nonexistent".

src/harness/testcases/relinquishment_testcase.py
```diff
@@ -123,3 +124,25 @@ def test_10_13_4_1_2(self):
...
+            'grantId': 'A non-exist grant id'
+        }]
+    }
+    response = self._sas.Relinquishment(request)['relinquishmentResponse'][0]
+    # Check the relinquishment response
```
kate-harrison: The text of the test says we need to check that grantId matches the grantId in the request but the SAS-CBSD TS says grantId should only be included if it's a valid grantId. It seems that your code is correct and that the test text needs to be updated; do you agree?

src/harness/testcases/relinquishment_testcase.py
qingwu123: Will do.

src/harness/testcases/relinquishment_testcase.py
qingwu123: Agreed. I will add this to the list of issues we maintain with Masoud.

PR #99, opened 2017-03-19
add registration testcase WINFF.FT.S.REG.23, 24, 25

janetliuboston:


src/harness/testcases/registration_testcase.py
```diff
@@ -128,3 +128,61 @@ def test_10_3_4_2_5_3(self):
...
+    device_a['installationParam']['heightType'] = 'AGL'
+    device_a['installationParam']['indoorDeployment'] = False
+    request = {'registrationRequest': [device_a]}
+    response = self._sas.Registration(request)['registrationResponse'][0]
+    # Check registration response
+    self.assertEqual(response['response']['responseCode'], 202)
+    self.assertTrue('cbsdId' in response)
+
+  @winnforum_testcase
+  def test_WINFF_FT_S_REG_24(self):
```
grao1991: I think "INVALID_VALUE" is also allowed?

janetliuboston: The WINNF-15-s-0061 v0.6.0 no longer has INVALID_VALUE (103) in the test case description, although it was in v0.5.x. v0.6.0: 

SAS rejects the request by sending a CBSD Registration Response as follows:
* SAS response does not include cbsdId
* The responseCode in Response Object should be 202, concluding a failed registration.

Have sent an email to the spec author to clarify or update spec.

janetliuboston:
Spoke with Masoud and Kambiz, and only CATEGORY_ERROR (202) is expected here, as is written in the latest test spec v0.6.

kate-harrison:
That seems like really strange behavior: it's easier for a device to change its height than its category, right? I'm having trouble understanding when this kind of error message would be useful to the CPI/operator.

I guess the intended behavior is for the SAS to validate params with cat A and, if they fail, check if the params would be valid if the CBSD had declared cat B instead. If yes, return CATEGORY_ERROR; otherwise return INVALID_VALUE? I'm not seeing this described in the spec anywhere (but I haven't looked carefully); can you point me to it?

janetliuboston:
Since there is no definition in the TS about this, and INVALID_VALUE may make more sense, shall we ask WF WG 3 about this?

kate-harrison:
Yes, let's ask WG3 about this. I think Kambiz said he'd take that action item?

I propose that we do one of the following with these tests in the meantime:
1) Submit with code which exactly matches the current test spec (even if we're hoping it will change)
2) Wait to submit until we receive clarification/notice of changes

kate-harrison:
FYI WG3 just approved the removal of CATEGORY_ERROR. I don't know when WG4 will update the test cases but you might want to change this code to expect INVALID_VALUE and add a note in each place that we're not exactly following v0.6.4 (and why). Submitting sooner will help you avoid merge conflicts :)

janetliuboston:
Great. Added note and updated from 202 to 103. Thank you.

src/harness/testcases/registration_testcase.py
```diff
@@ -940,12 +940,86 @@ def test_WINNF_FT_S_REG_22(self):
...
+    # Register the device
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
+    device_a['installationParam']['eirpCapability'] = 31
+    request = {'registrationRequest': [device_a]}
+    response = self._sas.Registration(request)['registrationResponse'][0]
+    # Check registration response
+    self.assertEqual(response['response']['responseCode'], 103)
+    self.assertTrue('cbsdId' in response)
```
kate-harrison: assertFalse?

src/harness/testcases/registration_testcase.py
```diff
@@ -940,12 +940,86 @@ def test_WINNF_FT_S_REG_22(self):
...
+    device_a['installationParam']['latitude'] = 38.882162
+    device_a['installationParam']['longitude'] = 77.113755
+    device_a['installationParam']['height'] = 8
+    device_a['installationParam']['heightType'] = 'AGL'
+    device_a['installationParam']['indoorDeployment'] = False
+    request = {'registrationRequest': [device_a]}
+    response = self._sas.Registration(request)['registrationResponse'][0]
+    # Check registration response
+    self.assertEqual(response['response']['responseCode'], 103)
+    self.assertTrue('cbsdId' in response)
```
kate-harrison: assertFalse?

src/harness/testcases/registration_testcase.py
```diff
@@ -940,12 +940,86 @@ def test_WINNF_FT_S_REG_22(self):
...
+    # Register the device
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+    self._sas_admin.InjectFccId({'fccId': device_b['fccId']})
+    device_b['installationParam']['indoorDeployment'] = True
+    request = {'registrationRequest': [device_b]}
+    response = self._sas.Registration(request)['registrationResponse'][0]
+    # Check registration response
+    self.assertEqual(response['response']['responseCode'], 103)
+    self.assertTrue('cbsdId' in response)
```
kate-harrison: assertFalse?

src/harness/testcases/registration_testcase.py
janetliuboston: Yes, thanks! Fixed all three.

PR #102, opened 2017-03-21
Add relinquishment testcase WINFF.FT.S.RLQ.5

qingwu123:


src/harness/testcases/relinquishment_testcase.py
```diff
@@ -146,3 +146,33 @@ def test_WINFF_FT_S_RLQ_4(self):
...
+
+    # Register the device
+    device = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
+    request = {'registrationRequest': [device]}
+    response = self._sas.Registration(request)['registrationResponse'][0]
+    # Check registration response
+    self.assertEqual(response['response']['responseCode'], 0)
+    self.assertTrue('cbsdId' in response)
```
kate-harrison: I don't think this is strictly necessary. If 'cbsdId' is not in response, line 168 will cause an error that's almost identical to the one you will see from line 167.

src/harness/testcases/relinquishment_testcase.py
```diff
@@ -146,3 +146,33 @@ def test_WINFF_FT_S_RLQ_4(self):
...
+    del request, response
+
+    # Relinquish grant
+    request = {'relinquishmentRequest': [
+        {'cbsdId': cbsd_id, 'grantId': 'A nonexistent grant id'}]}
+    response = self._sas.Relinquishment(request)['relinquishmentResponse'][0]
+    # Check the relinquishment response
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertFalse('grantId' in response)
+    self.assertEqual(response['response']['responseCode'], 103)
```
kate-harrison: Do we also need to check that responseData contains 'grantId'?

src/harness/testcases/relinquishment_testcase.py
qingwu123: Will do.

src/harness/testcases/relinquishment_testcase.py
qingwu123: I thought about this. ResponseData is optional according to TS. We don't check it for other test cases.

src/harness/testcases/relinquishment_testcase.py
kate-harrison: Thanks!

src/harness/testcases/relinquishment_testcase.py
kate-harrison: Good point! In that case, we should probably ask for the test doc to be updated to match the TS: either remove this line or change it to "if responseData is not null, it should contain...". For now I agree that leaving it out of the test code is the best option.

qingwu123: Will add it to the test doc comment spreadsheet.

PR #103, opened 2017-03-21
Implement WINFF.FT.S.SIQ.14

krlinus:
Parameters in inquiredSpectrum mutually invalid in spectrum inquiry


janetliuboston:
Optional: Do you want to add a check for cbsdId exists like in SIQ.11 for consistency?
self.assertTrue('cbsdId' in response)

PR #105, opened 2017-03-22
add registration test case WINFF.FT.S.REG.26

janetliuboston:


src/harness/testcases/registration_testcase.py
```diff
@@ -147,3 +147,69 @@ def test_10_3_4_2_5_3(self):
...
+    device_1 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_2 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_3 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+    device_4 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+
+    # Set up unique FCC IDs
```
kate-harrison: I wonder if it makes sense to define device_c.json and device_d.json so that we don't have to jump through these hoops? We will probably need them for other tests anyway.

janetliuboston: That makes sense. I'll add device_c.json and device_d.json.

janetliuboston: Test case WINFF.FT.S.REG.15 requires an array of six CBSDs, five Cat A CBSDs and one Cat B CBSD. Should I create four more Cat A CBSDs:
device_c.json
device_d.json
device_e.json
device_f.json

and one more for Cat B?
device_g.json

kate-harrison: That seems reasonable. Isn't device_b.json already Cat B though? Then we might not need to create device_g.json.

janetliuboston: REG.14 needs two device B's. Shall we group all the A's together (a, b, c, d, e), and all the B's together (f, g)?

-	valid userId, fccId, and cbsdSerialNumber plus all Conditional parameters for Cat B CBSD shall be included for the third CBSD (Cat B)
-	valid userId, fccId, and cbsdSerialNumber shall be included for the fourth CBSD (Cat B).

janetliuboston: Or perhaps create a name that is clearer about what types of devices they are...

kate-harrison: Grouping is OK although we may not be able to maintain the scheme if we add more devices in the future.

I'm not opposed to changing the name (and if we did, I'd suggest using registration_* instead of device_* to match the other data files) although I worry a bit about trying to capture the parameters in the file name (why stop at category? why not also include location, EIRP, etc.?). It's not obvious to me what the best naming scheme is but I'm open to suggestions.

janetliuboston: I'm not sure either. It seems other test cases use the device_x.json files when registering CBSDs. I'm fine with renaming to registration_x.json or keeping device_x.json. 

How about naming the devices by order needed in test case order for now?
device_a.json: cat A (reg.1)
device_b.json: cat B (reg.2)
device_c.json: cat A (reg.14)
device_d.json: cat B (reg.14)
device_e.json: cat A (reg.15)
device_f.json: cat A (reg.15)
device_g.json: cat A (reg.15)

kate-harrison: I have no strong opinions on the ordering so I'm happy with your proposal. Let's only rename to registration_* if we have to change the names for another reason (e.g. if we decided to re-order the files later); renaming at any point will cause headache for anyone who currently has a PR under review and I'd prefer to minimize that :)

Another thing to consider is how much we want to diversify the parameters for the devices: different locations, max EIRPs, etc. I don't have any strong opinions on it but it might be worth thinking about.

janetliuboston:
Sounds good. I agree about keeping the name to minimize changes.

I also agree about thinking about how much we want to diversify the parameters. For now, I've tried to make as many different parameters as possible and kept all conditional parameters in the files. If we need to change them later, we can do so.

janetliuboston:
Updated with device_c.json (Cat A), device_d.json (Cat B), device_e.json (Cat A), device_f.json (Cat A).

src/harness/testcases/registration_testcase.py
```diff
@@ -162,3 +162,66 @@ def test_10_3_4_2_5_3(self):
...
+
+    device_1 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_2 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_c.json')))
+    device_3 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_e.json')))
+    device_4 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+
```
kate-harrison: Does it make sense to use

self.assertEqual(device_1['cbsdCategory'], 'A')

to match the test spec?

src/harness/testcases/registration_testcase.py
```diff
@@ -162,3 +162,66 @@ def test_10_3_4_2_5_3(self):
...
+    # First device success
+    self.assertTrue('cbsdId' in response['registrationResponse'][0])
+    self.assertFalse('measReportConfig' in response['registrationResponse'][0])
+    self.assertEqual(response['registrationResponse'][0]['response']['responseCode'], 0)
+
+    # Second, third, fourth devices failure 202
+    for x in range (1,4):
+        self.assertFalse('cbsdId' in response['registrationResponse'][x])
+        self.assertFalse('measReportConfig' in response['registrationResponse'][x])
```
kate-harrison: I don't think we need to check this?

src/harness/testcases/registration_testcase.py
```diff
@@ -162,3 +162,66 @@ def test_10_3_4_2_5_3(self):
...
+    response = self._sas.Registration(request)
+
+    # First device success
+    self.assertTrue('cbsdId' in response['registrationResponse'][0])
+    self.assertFalse('measReportConfig' in response['registrationResponse'][0])
+    self.assertEqual(response['registrationResponse'][0]['response']['responseCode'], 0)
+
+    # Second, third, fourth devices failure 202
+    for x in range (1,4):
+        self.assertFalse('cbsdId' in response['registrationResponse'][x])
```
kate-harrison: Technically this is not required by the test spec (at least the version I'm looking at). Should we suggest a change to the doc?

src/harness/testcases/testdata/device_c.json
```diff
@@ -0,0 +1,20 @@
+{
+  "fccId": "test_fcc_id_a2",
```
kate-harrison: Is there a reason you're not using "test_fcc_id_c"? If you're using this to reference the category, maybe we should change device_{a,b}.json to match ("test_fcc_id_a1")?

src/harness/testcases/registration_testcase.py
janetliuboston: Yes.

src/harness/testcases/testdata/device_c.json
janetliuboston: I will update them to use test_fcc_id_x", where x is a, b, c, d, e, f, etc.

src/harness/testcases/registration_testcase.py
janetliuboston: Test spec does not have this check. Will remove it.

src/harness/testcases/registration_testcase.py
janetliuboston: Agree. Will remove these.

src/harness/testcases/registration_testcase.py
janetliuboston: I'll add checking for all four devices for the category type.

src/harness/testcases/registration_testcase.py
janetliuboston: We can suggest a change to the doc as well. I'll bring this up.

src/harness/testcases/registration_testcase.py
```diff
@@ -181,3 +181,69 @@ def test_10_3_4_2_5_3(self):
...
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_2 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_c.json')))
+    device_3 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_e.json')))
+    device_4 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+
+    self.assertEqual(device_1['cbsdCategory'], 'A')
+    self.assertEqual(device_2['cbsdCategory'], 'A')
```
kate-harrison: Consider putting this in after line 207 where it's clearer that device_2 needs to be cat. A?

janetliuboston: Good idea.

PR #116, opened 2017-03-23
add registration test WINFF.FT.S.REG.4

janetliuboston:


janetliuboston:
The SAS does not seem to have interfaces to check for Step 3. Any thoughts on how to verify this?
"If the SAS had any existing grants assigned to the CBSD, upon receiving the RegistrationRequest, all such grants should be revoked and removed. If the SAS had no existing grants for the CBSD, this step is skipped."

kate-harrison:
I see two options:
1) We know that the SAS doesn't have any grants assigned to this CBSD (because we called Reset()), so we can skip this step.
2) As part of test step 1, we could also request a grant. For step 3, we would try to heartbeat the grant and verify that it's been revoked via the response.

I think #1 is the easier approach; #2 seems like it will be a lot of work.

janetliuboston:
Thanks. Option #1 sounds perfect.

PR #122, opened 2017-03-23
add registration test WINFF.FT.S.REG.8

janetliuboston:


janetliuboston:
Removed check for CBSD ID on re-registration to be the same as the original. The CBSD IDs could be different.

PR #124, opened 2017-03-23
add registration test WINFF.FT.S.REG.9

janetliuboston:
The test case does not compare the recorded CBSD IDs with the re-registered CBSD IDs, but I think it should. I will propose this change in the spec.

    self.assertTrue(cbsdId1 == response['registrationResponse'][0]['cbsdId'])
    self.assertTrue(cbsdId2 == response['registrationResponse'][1]['cbsdId'])

### DIFF
--- a/src/harness/testcases/registration_testcase.py
+++ b/src/harness/testcases/registration_testcase.py
@@ -142,9 +142,6 @@ class RegistrationTestcase(unittest.TestCase):
         self.assertTrue('cbsdId' in response['registrationResponse'][x])
         self.assertEqual(response['registrationResponse'][x]['response']['responseCode'], 0)
 
-    # Save CBSD IDs
-    cbsdId1 = response['registrationResponse'][0]['cbsdId']
-    cbsdId2 = response['registrationResponse'][1]['cbsdId']
     del devices, request, response
 
     # Re-register two devices, register third device
@@ -157,9 +154,6 @@ class RegistrationTestcase(unittest.TestCase):
         self.assertTrue('cbsdId' in response['registrationResponse'][x])
         self.assertFalse('measReportConfig' in response['registrationResponse'][x])
         self.assertEqual(response['registrationResponse'][x]['response']['responseCode'], 0)
-
-    self.assertTrue(cbsdId1 == response['registrationResponse'][0]['cbsdId'])
-    self.assertTrue(cbsdId2 == response['registrationResponse'][1]['cbsdId'])
 
   @winnforum_testcase
   def test_10_3_4_2_1(self):
--- a/src/harness/testcases/registration_testcase.py
+++ b/src/harness/testcases/registration_testcase.py
@@ -108,6 +108,61 @@ class RegistrationTestcase(unittest.TestCase):
     self.assertEqual(response['response']['responseCode'], 0)
 
   @winnforum_testcase
+  def test_WINFF_FT_S_REG_9(self):
+    """ Array Re-registration of Single-step-registered CBSD (CBSD ID exists)
+
+    The response should be SUCCESS.
+    """
+
+    # Register the device
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_b = json.load(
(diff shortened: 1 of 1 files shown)


kate-harrison:
Does the spec require that the CBSD ID issued during re-registration is the same as the original CBSD ID? If so, I agree with your proposed changes (although it might be worth including a comment that we are deviating from the test spec in this specific place).

janetliuboston:
I'll check with Masoud on the correct behavior tomorrow.

axsahoo:
The end result of this case should be same as CBSD unregistering and registering again. So, in this case, if the SAS had any grants when the re-registration was sent, then those grants should be revoked by the SAS. Since this is like CBSD unregistering and registering again, there is no guarantee that the CBSD will receive the same cbsdId as before. But one critical component of this test is to test the scenario that the CBSD has a successful grant before sending re-registration. And after re-registration is done, it should check that the grant is indeed revoked by the SAS (by sending a heartbeat with the grantId?).

janetliuboston:
CBSD ID does not have to be the same as the original CBSD ID on re-registration. I have removed the checks.

kate-harrison:
Anirud: I agree that this is a good case to check. I think the best strategy here is to submit a comment asking for a test which explicitly tests this behavior (or a modification to this test). Our policy has been to follow what is written as closely as possible and unfortunately testing this behavior would require us to deviate substantially from the test spec (as currently written). Does this seem like a reasonable plan?

src/harness/testcases/registration_testcase.py
```diff
@@ -108,6 +108,54 @@ def test_WINFF_FT_S_REG_6(self):
...
+    device_b['measCapability'] = []
+    device_c['measCapability'] = []
+
+    # Register two devices
+    devices = [device_a, device_b]
+    request = {'registrationRequest': devices}
+    response = self._sas.Registration(request)
+    
+    # Check registration response
+    for x in range (0,2):
```
kate-harrison: Consider changing for readability:

```
for resp in response['registrationResponse']:
  self.assertTrue('cbsdId' in resp)
  self.assertEqual(resp['response']['responseCode'], 0)
```

src/harness/testcases/registration_testcase.py
```diff
@@ -108,6 +108,54 @@ def test_WINFF_FT_S_REG_6(self):
...
+    self._sas_admin.InjectFccId({'fccId': device_c['fccId']})
+
+    # Make sure measCapability contains no value for all array elements
+    device_a['measCapability'] = []
+    device_b['measCapability'] = []
+    device_c['measCapability'] = []
+
+    # Register two devices
+    devices = [device_a, device_b]
+    request = {'registrationRequest': devices}
```
kate-harrison: Since you only use `devices` once, you could just do `request = {'registrationRequest', [device_a, device_b]}` and save yourself the delete later. It's not really a big deal either way though.

src/harness/testcases/registration_testcase.py
janetliuboston: I do like that for loops where all the elements are enumerated.

src/harness/testcases/registration_testcase.py
janetliuboston: That is better.

janetliuboston:
Updated. Thank you for the comments!

axsahoo:
Kate: The scenario I am suggesting regarding checking that existing grants get revoked by SAS is actually mentioned in the test case (Step 3), but some how the steps are not enumerated for that scenario.

kate-harrison:
Anirud: yes, I think it would be reasonable to amend step 1 of this test case so that we explicitly test for that scenario. Would you be willing to file a comment on that to WinnForum? Is it OK to use the current implementation until the test case is officially amended? I'm sure we'll have to look through all (and update some) of the tests once the final version is approved anyway.

axsahoo:
ok. I will take it to WG4 of WinnForum.

kate-harrison:
Great -- thank you!

janetliuboston:
Thank you, Anirud and Kate!

PR #126, opened 2017-03-24
Implement WINFF.FT.S.SIQ.19

krlinus:
Two spectrum inquiry requests - One request successful and the other
unsuccessful

### DIFF
--- a/src/harness/testcases/spectrum_inquiry_testcase.py
+++ b/src/harness/testcases/spectrum_inquiry_testcase.py
@@ -206,3 +206,53 @@ class SpectrumInquiryTestcase(unittest.TestCase):
     # Check Spectrum Inquiry Response
     response = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse'][0]
     self.assertEqual(response['response']['responseCode'], 300)
+
+  @winnforum_testcase
+  def test_WINFF_FT_S_SIQ_19(self):
+    """Send Spectrum Inquiry with dual requests #1 successful #2 unsuccessful.
+
+    The response should be NO_ERROR (code 0) and INVALID_PARAM, code 103
+    """
+    # Register the devices
+    device_a = json.load(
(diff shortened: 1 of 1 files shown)


janetliuboston:
Please resolve conflicts.

krlinus:
Conflicts are resolved.

src/harness/testcases/spectrum_inquiry_testcase.py
```diff
@@ -206,3 +206,53 @@ def test_WINFF_FT_S_SIQ_15(self):
...
+    }
+    request = {'spectrumInquiryRequest': [
+        spectrum_inquiry_0, spectrum_inquiry_1]}
+    # Send spectrumInquiryRequest and get response
+    response = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse']
+    self.assertEqual(len(response), 2)
+    # Check Spectrum Inquiry Response
+    self.assertEqual(response[0]['cbsdId'], cbsd_ids[0])
+    self.assertTrue('availableChannel' in response[0])
+    for available_channel in response[0]['availableChannel']:
```
janetliuboston: If response[0]['availableChannel'] is null, will this fail?

Add a check?
    if response[0]['availableChannel'] is not None:
        for available_channel in response[0]['availableChannel']:

krlinus: This failure will mean that the test failed. TS says:
"This parameter is an array of zero or more data objects, AvailableChannel, which
describes a channel that is available for the CBSD, see Table 19.
Included: If and only if the Spectrum Inquiry is successful."
It should be an array of objects, even if it contains no elements.

janetliuboston: Agree with this. The test spec needs to be updated to reflect the availableChannel will always be there for successful spectrum inquiry.

PR #127, opened 2017-03-24
Implement WINFF.FT.S.SIQ.20

krlinus:
Unsupported Spectrum (Error Code 300) when requesting spectrum inquiry


kate-harrison:
Can someone from FW please take a look at this?

PR #139, opened 2017-03-24
move reg.9 to correct place; no other changes

janetliuboston:


src/harness/testcases/registration_testcase.py
```diff
@@ -219,6 +182,44 @@ def test_WINNF_FT_S_REG_9(self):
         self.assertEqual(resp['response']['responseCode'], 0)
 
   @winnforum_testcase
+  def test_10_3_4_2_1(self):
```
kate-harrison: May as well rename it now?

janetliuboston: Yes

janetliuboston:
Also:
- Verified that the latest spec will fix the naming from WINFF to WINNF, so replaced that here.
- Removed two test cases that are no longer in v0.6.0

PR #145, opened 2017-03-28
Implement WINFF.FT.S.GRA.22

krlinus:
maxEirp specified in the grant for Cat B CBSD is not supported


krlinus:
This is included in Pull request #147

PR #147, opened 2017-03-28
Implement WINNF.FT.S.GRA.21, WINNF.FT.S.GRA.22 and WINNF.FT.S.GRA.23

krlinus:
Both requests are for GAA channels, no incumbent present in the GAA frequency ranges used in the two requests.


kate-harrison:
Could someone from FW please take a look at this? Thanks!

janetliuboston:
Yes, I'll take a look. Thanks!

PR #153, opened 2017-03-28
add relinquishment test WINFF.FT.S.RLQ.9

qingwu123:


src/harness/testcases/relinquishment_testcase.py
```diff
@@ -229,6 +229,106 @@ def test_WINFF_FT_S_RLQ_7(self):
...
+    The response should be FAIL.
+    """
+
+    # Register the device
+    device_2 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_2['fccId'] = "test_fcc_id_2"
+    device_4 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_4['fccId'] = "test_fcc_id_4"
```
kate-harrison: Why not just load device_b.json and avoid having to create new FCC IDs?

src/harness/testcases/relinquishment_testcase.py
```diff
@@ -229,6 +229,106 @@ def test_WINFF_FT_S_RLQ_7(self):
...
+    self.assertFalse('grantId' in response[1])
+    self.assertEqual(response[1]['response']['responseCode'], 103)
+    self.assertFalse('cbsdId' in response[2])
+    self.assertFalse('grantId' in response[2])
+    self.assertIn(response[2]['response']['responseCode'], [103, 105])
+    self.assertEqual(response[3]['cbsdId'], cbsd_id_2)
+    self.assertEqual(response[3]['grantId'], grant_id_4)
+    self.assertEqual(response[3]['response']['responseCode'], 103)
+    self.assertEqual(response[4]['cbsdId'], cbsd_id_2)
+    self.assertFalse('grantId' in response[4])
```
kate-harrison: I think this is supposed to be equal to grant_id_3?

src/harness/testcases/relinquishment_testcase.py
qingwu123: Sure, I can do that.

src/harness/testcases/relinquishment_testcase.py
qingwu123: grand_id_3 is relinquished earlier so it's no longer a valid grant. TS says invalid grant id should not be included in the response. I have requested Masoud/Kambiz to make correct in the test plan.

qingwu123:
It definitely felt like tongue twister equivalent of a test case.

PR #161, opened 2017-04-03
Fixed an issue on test_10_7_4_1_3_1_1_2 and renamed it 

redtechnologies:
As suggested I modified test_10_7_4_1_3_1_1_2 for it to follow the spec and I renamed it test_WINFF_FT_S_GRA_8


kate-harrison:
Red: I think you should be able to merge this now. Please let us know if you're having any difficulties with the process.

kate-harrison:
I just looked through the details a bit more and it looks like what happened was that the original PR was approved but then another commit was added post-approval (which I didn't notice when I made my comment two days ago).

I think these changes are good but not complete: we also need to update sas_interface.py to include the new methods and add the associated documentation so it's clear how to use these functions (especially if the reader only has access to the code, not our API doc). We should also make changes to fake_sas.py (I think only line 200 needs to be modified) so that the fake SAS will respond if someone runs a test including these methods. Can someone at Red take the action item to finish off this work? Thanks!

redtechnologies:
Apologies for this unapproved change to the master. 
Ok, we will update sas_interface.py and fake_sas.py as you suggest.

Christophe

PR #162, opened 2017-04-05
fixes to REG.4, REG.10, REG.12, REG.22

janetliuboston:
REG.4: test spec does not require us to check that it's the same CBSD ID (line 199)
    REG.10: should also check that there is no CBSD ID in the response
    REG.12: should the error code be 200 instead of 103?
    REG.22: according to the test spec, we don't need to check that there is no CBSD ID in the third response

### DIFF
--- a/src/harness/testcases/registration_testcase.py
+++ b/src/harness/testcases/registration_testcase.py
@@ -182,6 +182,7 @@ class RegistrationTestcase(unittest.TestCase):
     request = {'registrationRequest': [device_b]}
     response = self._sas.Registration(request)['registrationResponse'][0]
     # Check registration response
+    self.assertTrue('cbsdId' in response)
     self.assertFalse('measReportConfig' in response)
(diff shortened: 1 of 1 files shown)


src/harness/testcases/registration_testcase.py
```diff
@@ -182,7 +182,6 @@ def test_WINNF_FT_S_REG_4(self):
     request = {'registrationRequest': [device_b]}
     response = self._sas.Registration(request)['registrationResponse'][0]
     # Check registration response
-    self.assertTrue('cbsdId' in response)
```
kate-harrison: Actually I think we want to remove line 198.

This line can also be removed (line 188 will fail if 'cbsdId' is not in response) but it wasn't the goal of my comment.

janetliuboston: I removed the wrong line. Fixed it. Thanks!

PR #167, opened 2017-04-10
ITS eHata reference code plus python bindings

gregbillock:
This branch brings in ITS eHata code with bindings to make it available as-is through python.


w4je:
LGTM

Issue #171, opened 2017-04-12
DER_8 - missing cbsd_id update after second registration


redtechnologies:
It seems cbsd_id is not updated after step 4 (second registration).
As a result step 5 (heartbeat request) and step 6 (second deregistration) are performed using the wrong cbsd_id.

janetliuboston:
Thank you for the comment. Will add the cbsd_id to this test case.
In practice, we hadn't added a different cbsd_id because it is a combination of FCC ID and CBSD serial number, and ends up being the same cbsd_id, but this is not specified in the TS, so it could be a different CBSD ID.

janetliuboston:
For reference, about the CBSD ID, from WINNF-15-S-0112-V1 5 0 CBRS Operational and Functional Requirements, R2-SRR-08:
The CBSD ID shall have a one-to-one correspondence with the combination FCC ID + CBSD S/N and may be identical to that combination.

janetliuboston:
Merged the fix in.

PR #174, opened 2017-04-13
Fix issue #171 DER.8

janetliuboston:
Fixed issue #171.

Also noticed that the conditional parameter measCapability was missing from the device json files. Added measCapability as a blank array for all devices.


src/harness/testcases/deregistration_testcase.py
```diff
@@ -30,7 +30,7 @@ def tearDown(self):
     pass
 
   @winnforum_testcase
-  def test_WINFF_FT_S_DER_1(self):
+  def test_WINNF_FT_S_DER_1(self):
```
kate-harrison: Consider updating to use "DRG" since you're changing these lines anyway?

PR #177, opened 2017-04-14
Implement WINNF.FT.S.HBT.16

krlinus:


src/harness/testcases/heartbeat_testcase.py
```diff
@@ -400,3 +401,100 @@ def test_10_9_4_2_3_1_3(self):
...
+            'grantId': grant_id
+        }]
+    }
+    response = self._sas.Relinquishment(request)['relinquishmentResponse'][0]
+    # Check the relinquishment response
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertEqual(response['grantId'], grant_id)
+    self.assertEqual(response['response']['responseCode'], 0)
+
+    # use relinquished grantId in new heartbeat request after transmitExpireTime
```
janetliuboston: How is this test case ensuring that the transmitExpireTime has passed?

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -400,3 +401,100 @@ def test_10_9_4_2_3_1_3(self):
...
+    transmit_expire_time = datetime.strptime(response['transmitExpireTime'],
+                                             '%Y-%m-%dT%H:%M:%SZ')
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertEqual(response['grantId'], grant_id)
+    self.assertLess(datetime.utcnow(), transmit_expire_time)
+    self.assertEqual(response['response']['responseCode'], 0)
+    del request, response
+
+    # Relinquish the grant
+    # Note: The Testcase requires grant to be terminated at this point, but
```
janetliuboston: Has this been submitted in WG4 or agreed by the author of the spec?

kate-harrison: We discussed with Yi Hsuan (one of the Google members of WG4) and he agreed that the test case should be changed since the more complex behavior specified in the test spec is not required to achieve the desired effect. We have not formally discussed with WG4 or test spec authors but plan to submit a comment when the spec is under ballot.

I see three options:
1) Submit the current code (knowing that we'll need to look carefully through all test cases once the spec is finalized anyway, and we can update at that time if needed)
2) Wait to submit until we get more formal agreement from WG4 (pre-ballot)
3) Write the code according to the test spec

(3) requires a lot more work for no gain; I expect we'll be successful in changing the test spec so I'd rather avoid (3). (2) isn't a bad option, especially if we think they will need to change other tests in a similar way. (1) also seems fine to me. Which do you prefer?

janetliuboston: Thanks. #1 sounds good.

src/harness/testcases/heartbeat_testcase.py
krlinus: On lines 492/493 I calculate how long I have to wait for transmitExpireTime to elapse, by subtracting the current time from it
On line 494 I call sleep() for the calculated amount of time + 1 second
On lines 495 and 496 I am ensuring that the transmitExpireTime has elapsed and the grantExpireTime is still in the future.

PR #183, opened 2017-04-19
Readme edits for pycurl openssl support requirement

jcereced:
Comments explaining that openssl support is required for the pycurl package


janetliuboston:
Prefer updating the pre-requisites in the README to avoid confusion with multiple paths.

jcereced:
I agree making the requirements more clear is preferable.  I wanted to document the issue we encountered and the workaround in case anybody were to run into the problem in the future.  We have tried recompiling PycURL with OpenSSL and are having issues (yum stopped working) so for the time being we left the changes in sas.py so we could use the test harness.  I agree that the best way to proceed is to simply update the pre-requisites

jcereced:
closing and opening a different simpler mod to the readme

PR #185, opened 2017-04-19
Added WINNF.FT.S.HBT.17

sushantparkhi:


src/harness/testcases/heartbeat_testcase.py
```diff
@@ -742,6 +744,68 @@ def test_WINNF_FT_S_HBT_15(self):
...
+    self.assertTrue(response['response']['responseCode'] in (103, 500))
+    del request, response
+
+  @winnforum_testcase
+  def test_WINNF_FT_S_HBT_17(self):
+    """Heartbeat Request from CBSD in Registered state (immediately after CBSD's grant is expired)
+    Response Code should  be 103 or 500"""
+
+    # Register the device
+    device_a = json.load(
```
kate-harrison: Do you mean "device_e"?

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -742,6 +744,68 @@ def test_WINNF_FT_S_HBT_15(self):
...
+    request = {'grantRequest': [grant_0]}
+
+    # Check grant response
+    response = self._sas.Grant(request)['grantResponse'][0]
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertEqual(response['response']['responseCode'], 0)
+    grant_id = response['grantId']
+
+    # Calculate the Difference Between Current Time and the GrantExpireTime
+    grant_expire_time = time.strptime(response['grantExpireTime'], '%Y-%m-%dT%H:%M:%SZ')
```
kate-harrison: I think you can simplify some of this by using the datetime module instead (see HBT_2 for an example).

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -742,6 +744,68 @@ def test_WINNF_FT_S_HBT_15(self):
...
+    current_time = time.strptime(datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ'), '%Y-%m-%dT%H:%M:%SZ')
+    difference_time = time.mktime(grant_expire_time) - time.mktime(current_time)
+    del request, response
+
+    # Log the Wait Time in seconds
+    print "Difference between grantExpireTime and CurrentTime (in seconds)", difference_time
```
kate-harrison: Please use the logging module rather than print statements.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -742,6 +744,68 @@ def test_WINNF_FT_S_HBT_15(self):
...
+
+    # Ensure that GrantExpireTime is Greater than CurrentTime
+    self.assertGreaterEqual(difference_time, 0)
```
kate-harrison: I think this will be easier to read if you do something like this: https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/heartbeat_testcase.py#L146

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -742,6 +744,68 @@ def test_WINNF_FT_S_HBT_15(self):
...
+    # Start the Timer to request the Heartbeat after the calculated seconds delay from the current time
+    Timer(difference_time, lambda: self.heartbeatRequest(cbsd_id, grant_id)).start()
```
kate-harrison: Isn't it simpler to just do:

```
time.sleep(grant_expire_time - current_time + 1)
# send heartbeat request and check response
```

I don't think setting up a Timer is necessary and I'm worried that it might cause problems.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -742,6 +744,68 @@ def test_WINNF_FT_S_HBT_15(self):
         datetime.strptime(response['transmitExpireTime'], '%Y-%m-%dT%H:%M:%SZ'),
         transmit_expire_time_1)
 
+  def heartbeatRequest(self, cbsd_id, grant_id):
```
kate-harrison: Let's not create helper functions, especially when they have generic names but test-specific behavior (e.g. acceptable response codes). I don't think you gain anything from this helper function, especially if you remove the use of Timer() as suggested below.

sushantparkhi:
I made the changes requested by you and is ready for review.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -789,23 +771,33 @@ def test_WINNF_FT_S_HBT_17(self):
...
-
-    # Calculate the Difference Between Current Time and the GrantExpireTime
-    grant_expire_time = time.strptime(response['grantExpireTime'], '%Y-%m-%dT%H:%M:%SZ')
-    current_time = time.strptime(datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ'), '%Y-%m-%dT%H:%M:%SZ')
-    difference_time = time.mktime(grant_expire_time) - time.mktime(current_time)
+    grant_expire_time = datetime.strptime(response['grantExpireTime'], '%Y-%m-%dT%H:%M:%SZ')
     del request, response
 
+    # Calculate the Difference Between Current Time and the GrantExpireTime
+    difference_time = (grant_expire_time - datetime.utcnow()).total_seconds() + 1
```
kate-harrison: I think you want to "+ 1" on line 784 instead.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -789,23 +771,33 @@ def test_WINNF_FT_S_HBT_17(self):
...
     # Log the Wait Time in seconds
-    print "Difference between grantExpireTime and CurrentTime (in seconds)", difference_time
-
+    logging.debug('Difference between grantExpireTime and CurrentTime (in seconds) ', difference_time)
     # Ensure that GrantExpireTime is Greater than CurrentTime
     self.assertGreaterEqual(difference_time, 0)
```
kate-harrison: Why not use
`self.assertGreaterEqual(grant_expire_time, datetime.utcnow())`

Then the code is so clear that you don't need the comment :)

PR #187, opened 2017-04-21
add WINNF.FT.S.HBT.5 & WINNF.FT.S.HBT.6

janetliuboston:


src/harness/testcases/heartbeat_testcase.py
```diff
@@ -272,6 +272,136 @@ def test_WINNF_FT_S_HBT_4(self):
...
+    grant_0 = json.load(
+        open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_0['cbsdId'] = cbsd_id
+    request = {'grantRequest': [grant_0]}
+    # Check grant response
+    response = self._sas.Grant(request)['grantResponse'][0]
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertTrue(response['grantId'])
+    self.assertEqual(response['response']['responseCode'], 0)
+    self.assertTrue(response['measReportConfig'])
```
kate-harrison: I think we might need a SAS trigger to make sure the SAS includes measReportConfig in the grantResponse. Page 17 of v1.0.0 of the SAS-CBSD TS says "If the measReportConfig parameter is included..." and I can't find any text saying that inclusion is required. What do you think?

kate-harrison: It feels like we should be checking the value of this field (not just its existence) but the test spec isn't very clear. Thoughts?

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -272,6 +272,136 @@ def test_WINNF_FT_S_HBT_4(self):
...
+    grant_id = response['grantId']
+    grant_expire_time = datetime.strptime(response['grantExpireTime'],
+                                          '%Y-%m-%dT%H:%M:%SZ')
+    del request, response
+
+    # Heartbeat
+    request = {
```
kate-harrison: I think the test spec is wrong and that we need to include the measReport field here.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -272,6 +272,136 @@ def test_WINNF_FT_S_HBT_4(self):
...
+        'heartbeatRequest': [{
+            'cbsdId': cbsd_id,
+            'grantId': grant_id,
+            'operationState': 'GRANTED'
+        }]
+    }
+    response = self._sas.Heartbeat(request)['heartbeatResponse'][0]
+    # Check the heartbeat response
+    self.assertTrue(response['measReportConfig'])
```
kate-harrison: I think the test spec is wrong and we don't need to check this.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -272,6 +272,136 @@ def test_WINNF_FT_S_HBT_4(self):
...
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertEqual(response['grantId'], grant_id)
+    transmit_expire_time = datetime.strptime(response['transmitExpireTime'],
+                                             '%Y-%m-%dT%H:%M:%SZ')
+    self.assertLess(datetime.utcnow(), transmit_expire_time)
```
kate-harrison: "transmitExpireTime ... shall be not later than 240 seconds in the future"

sushantparkhi:
I have made changes and is ready for review.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -317,6 +320,35 @@ def test_WINNF_FT_S_HBT_5(self):
         }]
     }
     response = self._sas.Heartbeat(request)['heartbeatResponse'][0]
+
+    # Check the heartbeat response
+    self.assertTrue(response['measReportConfig'])
```
kate-harrison: Should we also check the value of measReportConfig? The spec isn't very clear.

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -273,6 +273,207 @@ def test_WINNF_FT_S_HBT_4(self):
...
+  def test_WINNF_FT_S_HBT_6(self):
+    """Array request: SAS-directed Heartbeat Req: SAS has requested CBSD to 
+    perform measurement through measReportConfig in Grant response or a 
+    previous Heartbeat Response.
+
+    The response should be SUCCESS.
+    """
+
+    # Register the device
+    device_a = json.load(
```
kate-harrison: Maybe device_e to match the file? Or device1 so that it's not implied that this is device_a.json?

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -273,6 +273,207 @@ def test_WINNF_FT_S_HBT_4(self):
...
+        self.assertEqual(resp['cbsdId'], cbsd_ids[resp_number])
+        self.assertEqual(resp['grantId'], grant_ids[resp_number])
+        transmit_expire_time = datetime.strptime(resp['transmitExpireTime'],
+                                                 '%Y-%m-%dT%H:%M:%SZ')
+        self.assertLess(datetime.utcnow(), transmit_expire_time)
+        self.assertLessEqual(
+          (transmit_expire_time - datetime.utcnow()).total_seconds(), 240)
+        self.assertLessEqual(transmit_expire_time, grant_expire_times[resp_number])
+        self.assertEqual(resp['response']['responseCode'], 0)
+    self.assertTrue(response[0]['measReportConfig'])
```
kate-harrison: Check value of measReportConfig?

src/harness/testcases/heartbeat_testcase.py
```diff
@@ -273,6 +273,207 @@ def test_WINNF_FT_S_HBT_4(self):
...
+    grant_0 = json.load(
+        open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_0['cbsdId'] = cbsd_id
+    request = {'grantRequest': [grant_0]}
+    # Check grant response
+    response = self._sas.Grant(request)['grantResponse'][0]
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertTrue(response['grantId'])
+    self.assertEqual(response['response']['responseCode'], 0)
+    self.assertTrue(response['measReportConfig'])
```
kate-harrison: I think the measReportConfig is still optional here? I think it's only required during heartbeat for this test (but I could be wrong -- I don't know this part of the spec very well).

PR #194, opened 2017-05-02
Deprecated

w4je:


gregbillock:
Why are you removing this wrapper?

w4je:
Because it calls an implementation of eHata that we're removing?

PR #195, opened 2017-05-03
Added Get and Post Request Methods

sushantparkhi:
Few SAS - SAS Test Cases requires GET Request to Pull the Data. Created separate methods to either make a GET or POST Request.

### DIFF
--- a/src/harness/sas.py
+++ b/src/harness/sas.py
@@ -36,9 +36,8 @@ def GetTestingSas():
   version = config_parser.get('SasConfig', 'Version')
   return SasImpl(base_url, version), SasAdminImpl(base_url)
 
-
-def _Request(url, request, ssl_cert, ssl_key):
-  """Sends HTTPS request.
+def _RequestPost(url, request, ssl_cert, ssl_key):
+  """Sends HTTPS POST request.
 
   Args:
     url: Destination of the HTTPS request.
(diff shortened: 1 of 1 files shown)


src/harness/sas.py
```diff
@@ -76,6 +75,40 @@ def _Request(url, request, ssl_cert, ssl_key):
   logging.debug('Response:\n' + body)
   return json.loads(body)
 
+def _RequestGet(url, ssl_cert, ssl_key):
```
kate-harrison: I think this implementation is actually POST, not GET. All of the SAS-CBSD and data injection RPCs are HTTP POST.

src/harness/sas.py
```diff
@@ -103,7 +136,7 @@ def Deregistration(self, request, ssl_cert=None, ssl_key=None):
     return self._CbsdRequest('deregistration', request, ssl_cert, ssl_key)
 
   def _CbsdRequest(self, method_name, request, ssl_cert=None, ssl_key=None):
-    return _Request('https://%s/%s/%s' %
+    return _RequestPost('https://%s/%s/%s' %
```
kate-harrison: Please re-align the arguments (indent 4 spaces) here and below.

src/harness/sas.py
sushantparkhi: Yes, all the Injection methods are HTTP POST. But SAS - SAS Testcases like ARE.1, ARE.3, SIR.1, SIR.3 and so on requires HTTP GET Request.

kate-harrison: I'll ask a different way: what's the difference in implementation between _RequestPost() and _RequestGet()? They look identical to me.

sushantparkhi: There is no conn.setopt(conn.POST, True) in the GET Request. If this is not set to true then by default it will make a GET Request. Also, GET Request doesn't have any body in it, I also removed the conn.setopt(conn.POSTFIELDS, request) line.

kate-harrison: This makes sense -- sorry I didn't see it the first time.

PR #198, opened 2017-05-09
Added SAS Implementation Record and ESC Sensor Data Injection

sushantparkhi:


src/harness/sas_interface.py
```diff
@@ -287,4 +287,28 @@ def InjectSasAdministratorRecord(self, request):
...
+    pass
+
+  @abc.abstractmethod
+  def InjectSasImplementationRecord(self, request):
+    """SAS admin interface to inject SAS Implementation Record into SAS under test.
+
+    Args:
+      request: A dictionary with a single key-value pair where the key is
+        "record" and the value is a SAS Implementation information (which is 
+        itself a dictionary). The dictionary is an SASImplementation object
```
kate-harrison: The language here is a little redundant. Maybe:

> A dictionary with a single key-value pair where the key is "record" and the value is a SasImplementation object (which is itself a dictionary specified in the SAS-SAS TS, WINNF-16-S-0096).

src/harness/sas_interface.py
```diff
@@ -287,4 +287,28 @@ def InjectSasAdministratorRecord(self, request):
...
+  @abc.abstractmethod
+  def InjectEscSensorDataRecord(self, request):
+    """SAS admin interface to inject ESC Sensor Data Record into SAS under test.
+
+    Args:
+      request: A dictionary with a single key-value pair where the key is
+        "record" and the value is a ESC Sensor Data information (which is 
+        itself a dictionary). The dictionary is an EscSensorData object 
+        (Specified in SAS-SAS TS WINNF-16-S-0096)
+    """
```
kate-harrison: Consider adding a note about the behavior here:

> Behavior: SAS should act as if it is connected to an ESC sensor with the provided parameters.

PR #199, opened 2017-05-10
Added Measurement Report Trigger for Heartbeat and Registration

sushantparkhi:


src/harness/fake_sas.py
```diff
@@ -208,7 +208,9 @@ def do_POST(self):
                        '/admin/injectdata/fss', '/admin/injectdata/wisp',
                        '/admin/injectdata/cluster_list',
                        '/admin/injectdata/pal_database_record',
-                       '/admin/injectdata/sas_admin'):
+                       '/admin/injectdata/sas_admin',
+                       '/admin/trigger/meas_report_in_registration',
```
kate-harrison: Janet pointed out the naming inconsistency in the Google Doc. I like her suggestion to use "_response" at the end of both. Would you mind updating to use that naming convention?

src/harness/sas_interface.py
```diff
@@ -287,4 +287,36 @@ def InjectSasAdministratorRecord(self, request):
...
+    pass
+
+  @abc.abstractmethod
+  def TriggerMeasurementReportRegistration(self, request):
+    """SAS admin interface to trigger measurement report request for all subsequent
+    registration request 
+
+    Args:
+      request: A dictionary with a single key-value pair where the key is
+        "meas_report_config" and the value is a string of permitted enumerations
```
kate-harrison: array of string?

src/harness/sas_interface.py
```diff
@@ -287,4 +287,36 @@ def InjectSasAdministratorRecord(self, request):
...
+        EUTRA_CARRIER_RSSI_NON_TX or EUTRA_CARRIER_RSSI_ALWAYS
```
kate-harrison: We might want to reference the docs instead of putting these values here -- otherwise we will have old values in the code when the set of permitted enumerations changes.

src/harness/sas_interface.py
```diff
@@ -287,4 +287,36 @@ def InjectSasAdministratorRecord(self, request):
...
+        
+    Note: The SAS will request a measurement report in the RegistrationResponse
```
kate-harrison: "will" -> "should"?

PR #202, opened 2017-05-15
Fixed Issues in Sas Interface

sushantparkhi:
* Added decorator in InjectSasImplementationRecord  and InjectEscSensorDataRecord 
* Changed order of functions

### DIFF
--- a/src/harness/sas_interface.py
+++ b/src/harness/sas_interface.py
@@ -322,7 +322,6 @@ class SasAdminInterface(object):
       request: A dictionary with a single key-value pair where the key is
         "meas_report_config" and the value is an array of string of permitted 
         enumerations specified in WINNF-16-S-0016
-  
     Note: The SAS should request a measurement report in the RegistrationResponse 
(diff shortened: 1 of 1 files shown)


src/harness/sas_interface.py
```diff
@@ -290,15 +290,39 @@ def InjectSasAdministratorRecord(self, request):
...
+    the provided parameters.
+    """
+    pass
+
+  @abc.abstractmethod
   def TriggerMeasurementReportRegistration(self, request):
     """SAS admin interface to trigger measurement report request for all subsequent
     registration request 
-
+
```
kate-harrison: Do we really want this extra whitespace? (here and below)

sushantparkhi:
Yes, that's a good point that we can match the order in the API Doc or put Injection and Trigger functions together because at some point there will be more triggers which will be required for SAS-SAS testcases.
For now, I have kept the Measurement Trigger functions together and injections required for SAS -SAS testcases above that.

PR #203, opened 2017-05-15
Added Trigger for SAS Implementation Record Pull

sushantparkhi:


sushantparkhi:
I have implemented SIR.2 which uses this trigger and HTTP Server required for execution of SIR.2. You can check the code in branch "sushant_server_sir_2". The HTTP Server Class has few helper methods to setup the server for Push/Pull Request and then wait for the response from SAS Under Test.
Is the feasibility in regards to an implementation or ITS requiring Public IP Address for Push Requests?

kate-harrison:
Correct: the feasibility problem relates to ITS's ability to accept *inbound* requests -- they already have some IT trouble with *outbound* requests so I'm worried they would not be able to handle inbound requests. I have an email thread with Kambiz about this and I'll add you to it so we can continue the discussion there.

PR #206, opened 2017-05-16
Added WINNF.FT.S.GRA.15

sushantparkhi:


src/harness/testcases/grant_testcase.py
```diff
@@ -314,6 +314,67 @@ def test_WINNF_FT_S_GRA_14(self):
...
+
+    # Create Grant Request with frequency range containing both
+    # PAL(3550000000 - 3560000000) and GAA(3560000001 - 3580000000) Channel
+    grant_0 = json.load(
+      open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_0['cbsdId'] = cbsd_id
+    grant_0['operationParam']['operationFrequencyRange'][
+      'lowFrequency'] = 3550000000.0
+    grant_0['operationParam']['operationFrequencyRange'][
+      'highFrequency'] = 3580000000.0
```
RedaHossino: do you think  we could get the frequencies of the PPA(FR1) and then we add another channel not in the PPA (FR2), so grant frequency will be FR = FR1 + FR2 ?
(of course that depends if they consider the frequency of the PPA as a configurable parameter) then they can change it from JSon files without the need to change the code.

src/harness/testcases/grant_testcase.py
```diff
@@ -314,6 +314,67 @@ def test_WINNF_FT_S_GRA_14(self):
...
+    request = {'grantRequest': [grant_0]}
+
+    # Send grant request and get response
+    response = self._sas.Grant(request)['grantResponse'][0]
+    # Check grant response
+    self.assertEqual(response['cbsdId'], cbsd_id)
+    self.assertFalse('grantId' in response)
+    self.assertTrue(response['response']['responseCode'] in 103)
```
redtechnologies: do you think this line: self.assertTrue(response['response']['responseCode'] in 103) work?
 maybe it should be  self.assertEqual(response['response']['responseCode'], 103) ?

src/harness/testcases/grant_testcase.py
```diff
@@ -287,63 +288,62 @@ def test_WINNF_FT_S_GRA_14(self):
...
+    # Create Grant Request containing PAL and GAA frequency
     grant_0 = json.load(
       open(os.path.join('testcases', 'testdata', 'grant_0.json')))
     grant_0['cbsdId'] = cbsd_id
     grant_0['operationParam']['operationFrequencyRange'][
-      'lowFrequency'] = 3550000000.0
+      'lowFrequency'] = pal_low_frequency
     grant_0['operationParam']['operationFrequencyRange'][
-      'highFrequency'] = 3580000000.0
+      'highFrequency'] = pal_high_frequency + 100000000.0
```
RedaHossino: 100_000_000 =100 MHz I think we need only one additional GAA channel for the grant.

src/harness/testcases/grant_testcase.py
```diff
@@ -285,6 +288,72 @@ def test_WINNF_FT_S_GRA_14(self):
...
+    cbsd_id = response['cbsdId']
+    del request, response
+
+    # Inject PAL and PPA Database Record
+    ppa_record['ppaInfo']['cbsdReferenceId'] = [cbsd_id]
+    self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+    self._sas_admin.InjectZoneData({'record': ppa_record})
+
+    # Trigger daily activities and wait for it to get it complete
+    self._sas_admin.TriggerDailyActivitiesImmediately()
```
kate-harrison: This seems like something that would be well-suited to a helper function :)

sushantparkhi: Yep. Should I write the helper function just to check the status or with a trigger as well? I see two ways of writing the helper function:
1. Helper Function with TriggerDailyActivitiesImmediately or TriggerDailyActivitiesScheduled with a Status Check
2. Helper Function just for the Status Check and Trigger will be from Test Case.

Do you have any preferences?

kate-harrison: My thought was something like "TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()" (kind of your #1 but without the Scheduled option).

sushantparkhi: I've committed code, let me know if that works.

src/harness/util.py
```diff
@@ -125,3 +135,14 @@ def makePpaAndPalRecordsConsistent(ppa_record, pal_records, low_frequency,
   # Converting from defaultdict to dict
   ppa_record = json.loads(json.dumps(ppa_record))
   return ppa_record, pal_records
+
+
+def TriggerDailyActivitiesImmediatelyAndWaitUntilComplete(_sas_admin):
```
kate-harrison: It could also be in sas_testcase.py so you don't have to pass in the SAS admin instance (I think it'll be less prone to error). If you keep it here, the param name should probably change to "sas_admin" to match Python naming conventions.

sushantparkhi: I think we are planning to have only assertions in the sas_testcase.py?

kate-harrison: Oh, was that the plan? :) I made my suggestions based on whether the function needed access to member variables/functions (asserts, _sas_admin) or not. Clearly we need to pick one and document it :)

sushantparkhi: I think I agree with you in a way that if the functions need an access to member variables we can have that function whether it's utility or assertion in the sas_testcase.py. But if it doesn't require any access to member variables then we can keep that in the util.py?

kate-harrison: Sounds good to me!

src/harness/sas_testcase.py
```diff
@@ -100,3 +103,13 @@ def assertRegisteredAndGranted(self, registration_request, grant_request,
 
     # Return cbsd_ids and grant_ids
     return list_of_cbsd_ids_and_grant_ids
+
+  def TriggerDailyActivitiesImmediatelyAndWaitUntilComplete(self):
+    self._sas_admin.TriggerDailyActivitiesImmediately()
+    signal.signal(signal.SIGALRM, timeout_error)
```
kate-harrison: It might be easier to read the code if you just used a lambda here to raise the exception.

sushantparkhi: I tried that! :) I think lambda doesn't support statements we can only have expressions in it?

src/harness/util.py
```diff
@@ -22,6 +22,11 @@
 import uuid
 
 
+class TimeoutException(Exception):
```
kate-harrison: Is it actually important to have a custom exception type? If you're going to try-except around the helper function then it makes sense, and otherwise you may as well just raise a generic error with a custom error message.

src/harness/sas_testcase.py
kate-harrison: Huh! Looks like you technically *can* do it but it's so unreadable that it's probably not worth it: https://stackoverflow.com/questions/8294618/define-a-lambda-expression-that-raises-an-exception

That said, consider moving the Timeout stuff to this file if you're not going to use it anywhere else.

sushantparkhi: Thanks for the link! Removed the timeout_error and used Lambda!

src/harness/sas_testcase.py
```diff
@@ -19,6 +19,9 @@
 import unittest
 import sas_interface
 import sas
+import signal
```
kate-harrison: Do we need to add this to dependency list in the README?

src/harness/sas_testcase.py
kate-harrison: Looks great :) Thanks!

src/harness/sas_testcase.py
sushantparkhi: I think signal is a part of Python Standard Library https://docs.python.org/2/library/signal.html

kate-harrison: Great!

sushantparkhi: Should I merge this branch to master as Sony already approved the other part?

kate-harrison: Sounds good!

PR #209, opened 2017-05-18
Added WINNF.FT.S.SIR.1

sushantparkhi:


src/harness/sas.py
```diff
@@ -135,6 +135,15 @@ def Relinquishment(self, request, ssl_cert=None, ssl_key=None):
   def Deregistration(self, request, ssl_cert=None, ssl_key=None):
     return self._CbsdRequest('deregistration', request, ssl_cert, ssl_key)
 
+  def ImplementationRecordExchange(self, request, ssl_cert=None, ssl_key=None):
```
kate-harrison: Maybe "GetSasImplementationRecord" would be more readable within the test case?

src/harness/sas.py
```diff
@@ -135,6 +135,15 @@ def Relinquishment(self, request, ssl_cert=None, ssl_key=None):
   def Deregistration(self, request, ssl_cert=None, ssl_key=None):
     return self._CbsdRequest('deregistration', request, ssl_cert, ssl_key)
```
kate-harrison: Don't forget to update sas_interface.py and fake_sas.py.

src/harness/sas.py
```diff
@@ -135,6 +135,15 @@ def Relinquishment(self, request, ssl_cert=None, ssl_key=None):
...
 
+  def ImplementationRecordExchange(self, request, ssl_cert=None, ssl_key=None):
+    return self._SasRequest('sas', request, ssl_cert, ssl_key)
```
kate-harrison: 'sas' has changed to 'sas_impl' in the latest version of the spec (and our API doc).

src/harness/testcases/testdata/sas_impl_record_0.json
```diff
@@ -0,0 +1,23 @@
+{
+  "id": "sas/0/0",
```
kate-harrison: Maybe we want to make this a little more realistic?

Also, the prefix has changed to "sas_impl/" (see the latest version of the SAS-SAS doc that Greg just sent out).

src/harness/testcases/testdata/sas_impl_record_0.json
```diff
@@ -0,0 +1,23 @@
...
+    "note": ["Test Data for Harness"]
+  },
+  "publicKey": "Key",
+  "fccInformation": {
+    "certificationId": "0",
+    "certificationDate": "2017-05-02",
+    "certificationExpiration": "2019-05-01",
+    "certificationConditions": "Condition",
+    "frn":"frn",
+    "sasPhase":"sas"
```
kate-harrison: I think this has to be "1" or "2".

src/harness/testcases/WINNF_FT_S_SIR_testcase.py
```diff
@@ -0,0 +1,62 @@
...
+    response = self._sas.ImplementationRecordExchange(impl_record['id'])
+
+    # Verify the response using SasImplementationMessage Object schema
+    schema_path = os.path.join('..', '..', 'schema', 'SasImplementationMessage.schema.json')
+    schema = json.load(open(schema_path))
+    Draft4Validator.check_schema(schema)
+    schema_directory = os.path.dirname(os.path.realpath(schema_path))
+    resolver = RefResolver(referrer=schema, base_uri='file://' + schema_directory + '/')
+    # Validate with schema
+    self.assertRaises(ValidationError, validate(response, schema, resolver=resolver))
```
kate-harrison: Shouldn't it *not* raise an error?

src/harness/testcases/WINNF_FT_S_SIR_testcase.py
```diff
@@ -0,0 +1,62 @@
...
+    # Inject the SAS Implementation Record
+    impl_record = json.load(
+      open(os.path.join('testcases', 'testdata', 'sas_impl_record_0.json')))
+    self._sas_admin.InjectSasImplementationRecord({'record': impl_record})
+
+    # Get the SAS Implementation Record using Pull Command
+    response = self._sas.ImplementationRecordExchange(impl_record['id'])
+
+    # Verify the response using SasImplementationMessage Object schema
+    schema_path = os.path.join('..', '..', 'schema', 'SasImplementationMessage.schema.json')
```
kate-harrison: Hmm, this is an interesting and efficient approach. I wonder if it makes sense to package it into a helper function (in sas.py, I guess?) called AssertContainsRequiredFields() which takes just the schema filename ("SasImplementationMessage.schema.json") and the response as input. It seems that we will use this for all SAS-SAS tests and I think it will make the test script more readable. What do you think?

src/harness/testcases/WINNF_FT_S_SIR_testcase.py
```diff
@@ -0,0 +1,62 @@
...
+    # Validate with schema
+    self.assertRaises(ValidationError, validate(response, schema, resolver=resolver))
+
+
+
+
+
+
+
+
```
kate-harrison: Lots of extra whitespace?

src/harness/testcases/WINNF_FT_S_SIR_testcase.py
sushantparkhi: I thought of doing the same thing! Should I create a new class for Helper function in sas.py? We can later add all the helper functions in the same class for other testcases as well.

kate-harrison: Actually (contrary to what I first suggested) I think the best plan is as follows:
 - Create `class SasTestCase(unittest.TestCase)` in sas_testcase.py (new file) and put the helper function in there.
 - Derive other test classes from this new class, e.g. `class SpectrumInquiryTestcase(sas_testcase.SasTestCase)`

This way all test cases have access to the same helper functions *and* the helper functions can call things like `self.assertTrue()`. The downside is that it may become cluttered if we have too many helper functions... but the upside is that they'll all be in one location (and maybe that will motivate us to keep the list small) :)

What do you think?

sushantparkhi:
I have made the changes requested by you. Thanks for your suggestions!

src/harness/fake_sas.py
```diff
@@ -222,6 +237,18 @@ def do_POST(self):
     self.end_headers()
     self.wfile.write(json.dumps(response))
 
+  def do_GET(self):
+    """Handles GET requests."""
+    path, value = self._parseUrl(self.path)
+    if path == "/v1.0/sas_impl":
```
kate-harrison: Do we really need to include the "/v1.0" part? The POST version doesn't have it.

src/harness/sas_testcase.py
```diff
@@ -0,0 +1,32 @@
...
+import json
+from jsonschema import validate, Draft4Validator, RefResolver
+import os
+import unittest
+import sas_interface
+
+
+class SasTestCase(sas_interface.SasTestcaseInterface, unittest.TestCase):
+
+  def AssertContainsRequiredFields(self, schema_path=None, response=None):
```
kate-harrison: You probably want to remove "=None" -- these args aren't optional.

src/harness/sas_interface.py
```diff
@@ -339,4 +354,21 @@ def TriggerMeasurementReportHeartbeat(self, request):
...
+  """Includes Helper Function interface for SAS-CBSD and SAS-SAS Testcases"""
+
+  __metaclass__ = abc.ABCMeta
+
+  @abc.abstractmethod
+  def AssertContainsRequiredFields(self, schema_path=None, response=None):
+    """Assertion of Required Fields in Response validating it with Schema
+
+    Args:
+      schema_path: A string containing the filename of the schema to be used
```
kate-harrison: Maybe it's better to call it "schema_filename"? Otherwise they might think they need to provide the full path.

Do we need to tell the user where this function will look for the schema files?

src/harness/testcases/testdata/sas_impl_record_0.json
```diff
@@ -0,0 +1,23 @@
+{
+  "id": "sas_impl/0/0",
```
kate-harrison: Shouldn't this contain the admin's ID as the second token?

src/harness/testcases/testdata/sas_impl_record_0.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "Sas Implementation Record",
+  "administratorId": "sas_admin/sas/0",
```
kate-harrison: Updated in SAS-SAS TS: I think "sas_admin/test_admin_name" would be correct given the new spec.

src/harness/fake_sas.py
sushantparkhi: The HTTP GET and POST methods shall be used for all SAS-SAS requests. The URL endpoints for SAS-SAS requests are described in Table 2. Requests are sent to the SAS URL path which includes the string “/{sas_version_number}/{sas_record_type}”...

src/harness/testcases/testdata/sas_impl_record_0.json
sushantparkhi: I think the format is sas_impl/$ADMINISTRATION/$SAS_IMPLEMENTATION and the $ADMINISTRATOR is the SAS-CA Certified unique SAS Administrator identifier

src/harness/testcases/testdata/sas_impl_record_0.json
sushantparkhi: SAS-SAS V1.0.0 and V1.5.1 has a format sas_admin/sas/$ADMINISTRATOR. Is this the recent change which is not yet reflected in the Spec?

src/harness/fake_sas.py
kate-harrison: The URL path is basically the same for SAS-CBSD, right? So the approach which works for SAS-CBSD should also work for SAS-SAS, I think... Am I missing something?

src/harness/testcases/testdata/sas_impl_record_0.json
kate-harrison: Yup, so you'd use something like "sas_impl/test_admin_name0/unique_impl_id0" as test data, right?

src/harness/testcases/testdata/sas_impl_record_0.json
kate-harrison: I believe this has been updated -- please see the email I just forwarded to you.

src/harness/fake_sas.py
sushantparkhi: SAS-SAS makes a GET Request that's why I have added go_GET method to handle that and then splitting the path into two parts one actual path of the request "v1.0/sas_impl" and another with the ID which is passed in the URL.

src/harness/testcases/testdata/sas_impl_record_0.json
sushantparkhi: I've committed new changes. Let me know if that works :)

src/harness/testcases/testdata/sas_impl_record_0.json
sushantparkhi: Thank you for forwarding the email! It seems that it has been updated in v1.5.3 Spec. Should we follow the latest version v1.5.3 while creating the test cases?

src/harness/fake_sas.py
kate-harrison: Sorry, you're completely right -- I was misreading do_POST().

I wonder if someday we'll want to version the admin API... but that's a question for later.

src/harness/testcases/testdata/sas_impl_record_0.json
kate-harrison: Looks good!

src/harness/testcases/testdata/sas_impl_record_0.json
kate-harrison: *sigh* Versioning is such a headache. Let's try to keep to the latest protocol specs since the changes are minimal, they'll be finalized "soon", and cert will be based on those anyway. Once they are actually finalized we might want to do a quick sweep of the code to check for parts that need to be updated (although changes in the test spec will probably tell us where to look).

sushantparkhi:
I've tested fake_sas with SIR.1 and it's working!

src/harness/sas_testcase.py
```diff
@@ -0,0 +1,32 @@
...
+class SasTestCase(sas_interface.SasTestcaseInterface, unittest.TestCase):
+
+  def AssertContainsRequiredFields(self, schema_filename, response):
+    schema_filename = os.path.join('..', '..', 'schema', schema_filename)
+    schema = json.load(open(schema_filename))
+    Draft4Validator.check_schema(schema)
+    schema_dir = os.path.dirname(os.path.realpath(schema_filename))
+    resolver = RefResolver(referrer=schema, base_uri='file://' + schema_dir + '/')
+    # Raises ValidationError when incorrect response
+    validate(response, schema, resolver=resolver)
```
RedaHossino: maybe we need to have more validation constraints on the properties according to the SAS-SAS spec (e.x datetime with specific formats, email, phoneNumber... etc).
another solution would be to compare the response with the injected data of SAS Implementation Json (this solution is not the exact way they specified in the test spec of [WINNF.FT.S.SIR.1]).
_what do you think?_

src/harness/testcases/testdata/sas_impl_record_0.json
```diff
@@ -0,0 +1,23 @@
...
+  "contactInformation": {
+    "contactType": "operation",
+    "name": "Contact Name",
+    "phoneNumber": ["123 456 7890"],
+    "email": ["admin@sas.com"],
+    "address": ["Address"],
+    "note": ["Test Data for Harness"]
+  },
```
assaneRed: comment on "sas_impl_record_0.json" object using by "WINNF_FT_S_SIR_testcase.py" in master branch
ContactInformation should be a array 
Refer to the last WINNF version for SAS-SAS protocol
https://workspace.winnforum.org/higherlogic/ws/public/download/3403/WINNF-16-S-0096-V1.5.8%20SAS-SAS%20Protocol%20Technical%20Specification.pdf

assaneRed:
comment on "sas_impl_record_0.json" object using by "WINNF_FT_S_SIR_testcase.py" in master branch
ContactInformation should be a array 
Refer to the last WINNF version for SAS-SAS protocol
https://workspace.winnforum.org/higherlogic/ws/public/download/3403/WINNF-16-S-0096-V1.5.8%20SAS-SAS%20Protocol%20Technical%20Specification.pdf

sushantparkhi:
Fixed in https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/255

PR #210, opened 2017-05-22
Added WINNF.FT.S.SIQ.17

sushantparkhi:


src/harness/testcases/spectrum_inquiry_testcase.py
```diff
@@ -391,20 +391,40 @@ def test_WINNF_FT_S_SIQ_16(self):
...
     """This test ensures that when a message with two requests, 
     both for PAL channels, is sent with valid parameters, 
     the SAS sends a successful response for both the requests.
     Response Code must be 0 for all CBSDs
     """
 
-    # Register the devices
+    # Load the Data
+    pal_low_frequency = 3550000000.0
+    pal_high_frequency = 3700000000.0
```
RedaHossino: according to the spec "WINNF-15-S-011" the pal license may aggregate up to **four** PAL channels, so I think giving15 channels is not coherent with that.
and the PAL Id is by census tract by frequency channel so one pal record for one census tract on one channel.

src/harness/testcases/spectrum_inquiry_testcase.py
```diff
@@ -391,6 +391,87 @@ def test_WINNF_FT_S_SIQ_16(self):
...
+      self.assertEqual(resp['response']['responseCode'], 0)
+      cbsd_ids.append(resp['cbsdId'])
+    del request, response
+
+    # Inject Cluster List
+    for ppa_id, cbsd_id in zip(ppa_ids,cbsd_ids):
+      cluster_list = {'zoneId': ppa_id,
+                      'cbsdIds': [cbsd_id]}
+      self._sas_admin.InjectClusterList(cluster_list)
+
```
RedaHossino: the PPA and PAL record are not injected into the SAS under test?

src/harness/testcases/spectrum_inquiry_testcase.py
```diff
@@ -391,6 +391,90 @@ def test_WINNF_FT_S_SIQ_16(self):
...
+        open(os.path.join('testcases', 'testdata', 'pal_record_0.json')))
+      ppa_record = json.load(
+        open(os.path.join('testcases', 'testdata', ppa_filename)))
+      ppa_record, pal_record = makePpaAndPalRecordsConsistent(ppa_record,
+                                                                  [pal_record],
+                                                                  pal_low_frequency,
+                                                                  pal_high_frequency,
+                                                                  device['userId'])
+
+      self._sas_admin.InjectZoneData({"record": ppa_record})
```
RedaHossino: "record" should be "zoneData" right?

sushantparkhi: Thanks for all your comments! I recently created PR to change the zoneData to record for consistency https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/sas_interface.py#L254 and to match with the Test API Doc https://docs.google.com/document/d/19YbeQsSzm5kCIOeVzsF8piqI0Lizy3p345aiVGf-Hlg/edit#heading=h.fbgibud28jgj

RedaHossino: thank you, I wasn't aware of this change in the document.

src/harness/testcases/spectrum_inquiry_testcase.py
```diff
@@ -391,6 +391,95 @@ def test_WINNF_FT_S_SIQ_16(self):
...
+    # Check registration response
+    cbsd_ids = []
+    for resp in response:
+      self.assertEqual(resp['response']['responseCode'], 0)
+      cbsd_ids.append(resp['cbsdId'])
+    del request, response
+
+    for ppa_record, pal_record, cbsd_id in zip(ppa_records, pal_records, cbsd_ids):
+      # Update PPA Record with CBSD ID and Inject Data
+      ppa_record['ppaInfo']['cbsdReferenceId'] = [cbsd_id]
```
RedaHossino: the cbsd_id(which is unique by one SAS) doesn't seem to be the same as the cbsdReferenceId (globally unique), and actually I don't know why we need both of them!

kate-harrison: You are correct that they're not the same. In this case, we are injecting a PPA into the SAS which manages exactly those CBSDs, so I think we should clarify in the test spec that we are injecting the raw CBSD ID and *not* the "CBSD reference ID" (as defined in the SAS-SAS spec) in this specific case. It will reduce implementation burden in both the test harness and the SAS under test. Do you agree?

RedaHossino: yeah it is easier to do that, but I think the bsdReferenceId in ppa should always mean the same thing in the cbsd-sas protocol, sas-sas protocol and their test specification. so I think in the tests (or in the helper function) we need to build the cbsdReferenceId in the ppa to be fccId/sha1(cbsdserialnumber), otherwise the specifications should allow to excahnge the CBSD-IDs ,and not the reference ids, in the PPA record of sas-sas protocol.
what do you think? 
we can discuss that further on the call if you prefer.

kate-harrison: The specs do not allow exchange of CBSD IDs directly because (for reasons I don't know) we are not supposed to exchange CBSD serial numbers with other SASes. So the CBSD Reference ID must be as described in the SAS-SAS TS when we are exchanging information between SASes (i.e. there is no potential to change the SAS-SAS spec at this time) and it is not necessarily the same as the CBSD ID (technically a SAS could choose to create the CBSD ID using the same methodology, but there is no requirement).

tl;dr: already there's a difference between the CBSD ID in the SAS-CBSD exchange and the CBSD Reference ID in the SAS-SAS exchange (and we can't change either of those decisions).

For the testing API *only*, I think it makes sense to use the CBSD ID instead of the CBSD Reference ID. If we use the CBSD Reference ID, the SAS would need to maintain mapping information so it can immediately translate the CBSD Reference ID to the CBSD ID (not a simple function since SHA-1 is not reversible). There's no other part of the SAS that requires this mapping information/functionality so we would add it only for testing purposes, which is something I generally oppose :) If we use the CBSD ID, the SAS can use the ID directly -- the same as it would do in normal SAS-SAS exchange.

I do agree that if we decide to use the CBSD ID in this case that we should document it clearly so that there's no confusion in the future.

RedaHossino: I agree with that, there is still one problem if the SAS check the format of the reference ID in the ppa record and if it doesn't generate the CBSD ID the same way as the reference id then this proposal will cause some validation problems for the SAS.
if the SASs agree to generate the CBSD IDs the same way as cbsd reference id the issue will be solved and I think this is the best situation, especially they describe the CBSD ID as "**globally unique**" so if the SASs doesn't agree on an algorithm to generate them(like cbsd reference id) it is difficult to assure the uniqueness of CBSD IDs

kate-harrison: I don't think it's necessary to agree on a format for the CBSD ID (see the email I just sent to Christophe).

My assumption is that your code for handling "inject PPA" will be at least slightly different from your code for handling "receive PPA from another SAS". If I'm correct, then it's just a matter of incorporating this small difference as well.

Consider also that *your* implementation is free to set the CBSD ID to be the same as the CBSD reference ID, with or without agreement from other SASes. In that case, you don't need to worry about what the test code is doing since it will always be the same for you :)

jcereced:
More of a comment/question on the TC itself than the implementation

The CBSD is making a SIQ for the 3550-3700 range and it's expecting a response to include a "PAL" channel only in the response.  Shouldn't the SAS also respond with the available (if any) "GAA" channel.  It's quite possible I'm basing this on an understanding of SIQ that may be obsolete...hence the question

sushantparkhi:
I think in the testcase I implemented in a way that the available channel can have GAA `self.assertTrue(any(channel['channelType'] == 'PAL'
                          for channel in resp['availableChannel']))` but in the whole list it should at least have one PAL as a channelType

src/harness/testcases/spectrum_inquiry_testcase.py
```diff
@@ -391,6 +391,95 @@ def test_WINNF_FT_S_SIQ_16(self):
...
+  def test_WINNF_FT_S_SIQ_17(self):
+    """This test ensures that when a message with two requests, 
+    both for PAL channels, is sent with valid parameters, 
+    the SAS sends a successful response for both the requests.
+    Response Code must be 0 for all CBSDs
+    """
+
+    # Load the Data
+    pal_low_frequency = 3550000000.0
+    pal_high_frequency = 3560000000.0
```
RedaHossino: according to the new decision in the spec of SAS-CBSD the frequncy will be integer, but maybe we can do this modification later when we will change it everywhere on the master branch.

jcereced:
I agree the way is implemented will work fine and not fail if the SAS responds with GAA channels as well.  I added that as a comment on the WG4 test case review doc so I think for now the TC is good!

kate-harrison:
It looks like some of the required fields were omitted from ppa_record_1.json (I think it should only be fipsCode and censusYear which are missing, but best to compare with ppa_record_0.json which we know is correct). Sushant, can you please create a quick PR to add these in?

PR #212, opened 2017-05-30
Added WINNF.FT.S.SIR.3

sushantparkhi:


src/harness/testcases/WINNF_FT_S_SIR_testcase.py
```diff
@@ -43,3 +43,19 @@ def test_WINNF_FT_S_SIR_1(self):
...
+  @winnforum_testcase
+  def test_WINNF_FT_S_SIR_3(self):
+    """This test verifies that a SAS Under Test can handle the Unknown Implementation 
+     Id 
+    Response Code must be 200 with empty JSON Body"""
+
+    # Inject the SAS Implementation Record
+    impl_record = json.load(
+      open(os.path.join('testcases', 'testdata', 'sas_impl_record_0.json')))
+    self._sas_admin.InjectSasImplementationRecord({'record': impl_record})
```
kate-harrison: Do we even need to inject this? I don't think it's necessary for the test.

sushantparkhi: Yes, it's not necessary for the test. I just thought to have some valid data in SAS UUT and then make an invalid request. I can remove the injection from there.

kate-harrison: I prefer removing it (less code to maintain) but it's also fine to leave it in.

PR #217, opened 2017-06-06
Fix WinnForum test case REG_14.

harinineelu:
Check for:
- registration requests [0], [2] to succeed with code 0
- registration requests [1], [3] to fail with code 200

Miscellaneous formatting changes.


kate-harrison:
FYI Harini is at Google so someone other than me should review this change :)

harinineelu:
The third device should fail due to lack of CPI information and hence the current test code is correct. This change is not required, so deleting.

PR #219, opened 2017-06-08
Added Helper Function to Make PPA, PAL and Device Consistent

sushantparkhi:


src/harness/util.py
```diff
@@ -25,3 +29,92 @@ def decorated_testcase(*args, **kwargs):
...
+      pal_record['channelAssignment'][assign]['lowFrequency'] = low_frequency
+      pal_record['channelAssignment'][assign]['highFrequency'] = high_frequency
+  # Make the date consistent in Ppa Record
+  ppa_record = _changeDateInPpaAndPal(ppa_record,
+                                      previous_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'),
+                                      next_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'))
+  # Add Pal Ids into the Ppa Record
+  ppa_record['ppaInfo']['palId'] = [pal_record['palId'] for pal_record
+                                    in pal_records]
+  device['installationParam']['latitude'], device['installationParam']['longitude'] = \
```
kate-harrison: If you really want to do this in this function, you should probably take an array of devices to suit the general case.

That said, I think it's better to keep this function focused (make PPA and PAL records consistent) and let the test case move the devices if it so desires -- it's only one line of code anyway.

src/harness/util.py
```diff
@@ -25,3 +29,92 @@ def decorated_testcase(*args, **kwargs):
...
+      A tuple consisting of PPA Record, PAL Record List and Device with 
+      modified Latitude and Longitude
+  """
+
+  previous_year_date = datetime.now().replace(year=datetime.now().year - 1)
+  next_year_date = datetime.now().replace(year=datetime.now().year + 1)
+
+  for pal_record in pal_records:
+    # Change the FIPS Code and Registration Date-Year in Pal Id
+    pal_record['palId'] = '/'.join([str(ppa_fips_code) if index == 2 else
```
kate-harrison: This feels overly complicated. Why not just use:

```
fcc_channel_id = "1"   # you will need to use this later to fill licenseFrequencyChannelId
pal_record['palId'] = "/".join(["pal", #same date code, fcc_channel_id])
```

The current implementation looks like it might break in interesting ways if the PAL ID in the input is malformed.

src/harness/testcases/testdata/pal_record_0.json
```diff
@@ -0,0 +1,28 @@
+{
+  "palId": "pal/05-2017/20041084200/A",
```
kate-harrison: I think we probably want to only fill in the fields that will *not* be replaced so that it's clear what the behavior is. What do you think?

src/harness/util.py
```diff
@@ -25,3 +29,92 @@ def decorated_testcase(*args, **kwargs):
...
+    if Point([lng, lat]).within(ppa_polygon):
+      return lat, lng
+    else:
+      return _getRandomLatLongInPolygon(polygon)
+  except NotImplementedError:
+    # If there's no intercept again call the function
+    return _getRandomLatLongInPolygon(polygon)
+
+
+def _changeDateInPpaAndPal(record, prev_date, next_date):
```
kate-harrison: This implementation is clever but it assumes that the keys exist (correctly) in the record. If you follow my suggestion above, they won't. If you prefer to specify-and-replace the fields, I think you leave yourself open to the possibility of someone forgetting to set the field ("oh, the code sets it anyway so I don't need to specify it).

src/harness/util.py
```diff
@@ -25,3 +29,92 @@ def decorated_testcase(*args, **kwargs):
...
+    # Make the date consistent in Pal Record for Registration and License
+    pal_record = _changeDateInPpaAndPal(pal_record,
+                                        previous_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'),
+                                        next_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'))
+    # Change License Information in Pal
+    pal_record['license']['licenseAreaIdentifier'] = str(ppa_fips_code)
+    pal_record['license']['licenseAreaExtent'] = \
+      'zone/census_tract/census/%d/%d' % (ppa_census_year, ppa_fips_code)
+    # Change Frequency Information in Pal
+    for assign in ('primaryAssignment', 'secondaryAssignment'):
```
kate-harrison: Is it OK to give them the same value? Maybe better is to omit secondaryAssignment for now (I'm not sure if they'll actually use it anyway).

src/harness/util.py
```diff
@@ -25,3 +29,92 @@ def decorated_testcase(*args, **kwargs):
...
+      pal_record['channelAssignment'][assign]['lowFrequency'] = low_frequency
+      pal_record['channelAssignment'][assign]['highFrequency'] = high_frequency
+  # Make the date consistent in Ppa Record
+  ppa_record = _changeDateInPpaAndPal(ppa_record,
+                                      previous_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'),
+                                      next_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'))
+  # Add Pal Ids into the Ppa Record
+  ppa_record['ppaInfo']['palId'] = [pal_record['palId'] for pal_record
```
kate-harrison: I think you also need to set the 'id' field in ppa_record.

src/harness/util.py
```diff
@@ -25,3 +29,92 @@ def decorated_testcase(*args, **kwargs):
...
+                                    enumerate(pal_record['palId'].split('/'))])
+    pal_record['userId'] = device['userId']
+    # Make the date consistent in Pal Record for Registration and License
+    pal_record = _changeDateInPpaAndPal(pal_record,
+                                        previous_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'),
+                                        next_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'))
+    # Change License Information in Pal
+    pal_record['license']['licenseAreaIdentifier'] = str(ppa_fips_code)
+    pal_record['license']['licenseAreaExtent'] = \
+      'zone/census_tract/census/%d/%d' % (ppa_census_year, ppa_fips_code)
```
kate-harrison: Set licenseFrequencyChannelId too? (although I'm not sure how it's used so maybe we can ignore it)

src/harness/util.py
```diff
@@ -25,3 +29,92 @@ def decorated_testcase(*args, **kwargs):
...
+      ppa_fips_code: (number) The FIPS Code of the Census Tract where 
+      the PPA belongs.
+      ppa_census_year: (number) The census year in which the census 
+      tract was defined.
+      low_frequency: (number) The Primary Low Frequency for PAL
+      high_frequency: (number) The Primary High Frequency for PAL
+      device: (dictionary) An object containing device which has to be
+      moved into the PPA Polygon.
+    Returns:
+      A tuple consisting of PPA Record, PAL Record List and Device with
```
kate-harrison: Are the dicts passed by value or by ref? If by ref, there's no need to return them at the end.

src/harness/testcases/testdata/ppa_record_0.json
```diff
@@ -0,0 +1,50 @@
+{
```
kate-harrison: Do you want to specify census year and FIPS code somewhere in here?

src/harness/util.py
```diff
@@ -45,76 +52,67 @@ def _getRandomLatLongInPolygon(polygon):
...
   # Add Pal Ids into the Ppa Record
-  ppa_record['ppaInfo']['palId'] = [pal_record['palId'] for pal_record
-                                    in pal_records]
-  device['installationParam']['latitude'], device['installationParam']['longitude'] = \
-    _getRandomLatLongInPolygon(ppa_record['zone'])
+  ppa_record = defaultdict(lambda: defaultdict(dict), ppa_record)
 
-  return ppa_record, pal_records, device
+  ppa_record['ppaInfo']['palId'] = [pal['palId'] for pal in pal_records]
+  ppa_record['id'] = 'zone/ppa/admin_0/%s' % (ppa_record['ppaInfo']['palId'][0])
```
kate-harrison: You probably want to add a random number (or other identifier here) -- otherwise injecting two PPAs which have the same PAL license (which is totally fine based on the specs) would cause the first to be overwritten by the second.

kate-harrison: Technically "admin_0" should come instead from "creator" in the PPA record.

src/harness/util.py
```diff
@@ -45,76 +52,67 @@ def _getRandomLatLongInPolygon(polygon):
...
 
+  # Make the date consistent in Ppa Record
+  ppa_record['ppaInfo']['ppaBeginDate'] = previous_year_date.strftime('%Y-%m-%dT%H:%M:%SZ')
+  ppa_record['ppaInfo']['ppaExpirationDate'] = next_year_date.strftime('%Y-%m-%dT%H:%M:%SZ')
+  return pal_records, ppa_record
```
kate-harrison: Return in the same order as the function signature?

src/harness/util.py
```diff
@@ -45,76 +52,67 @@ def _getRandomLatLongInPolygon(polygon):
...
-    for assign in ('primaryAssignment', 'secondaryAssignment'):
-      pal_record['channelAssignment'][assign]['lowFrequency'] = low_frequency
-      pal_record['channelAssignment'][assign]['highFrequency'] = high_frequency
-  # Make the date consistent in Ppa Record
-  ppa_record = _changeDateInPpaAndPal(ppa_record,
-                                      previous_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'),
-                                      next_year_date.strftime('%Y-%m-%dT%H:%M:%SZ'))
+    pal_rec['channelAssignment']['primaryAssignment']['lowFrequency'] = low_frequency
+    pal_rec['channelAssignment']['primaryAssignment']['highFrequency'] = high_frequency
+    pal_records[index] = json.loads(json.dumps(pal_rec))
```
kate-harrison: What does this do? (can you add a comment to the code to explain?)

src/harness/util.py
```diff
@@ -45,76 +52,67 @@ def _getRandomLatLongInPolygon(polygon):
...
   previous_year_date = datetime.now().replace(year=datetime.now().year - 1)
   next_year_date = datetime.now().replace(year=datetime.now().year + 1)
+  fcc_channel_id = '1'
+  ppa_fips_code = ppa_record['fipsCode']
+  ppa_census_year = ppa_record['censusYear']
+  del ppa_record['censusYear'], ppa_record['fipsCode']
 
-  for pal_record in pal_records:
+  for index, pal_rec in enumerate(pal_records):
+    pal_rec = defaultdict(lambda: defaultdict(dict), pal_rec)
```
kate-harrison: I'm admittedly not familiar with defaultdict() but I can't quite figure out what benefit this provides.

sushantparkhi: This helps to add a key into the dictionary without raising the KeyError of some default type. For example in case of **pal_rec['channelAssignment']['primaryAssignment']['lowFrequency']** it will raise a KeyError as channelAssignment does not exist in the dictionary. To get rid of a lot of ifs to check if Key is there or not we can use defaultdict which will create a key in the dictionary with some default type if it's not there.

src/harness/util.py
sushantparkhi: This line is again converting the defaultdict to dict type.

src/harness/util.py
```diff
@@ -25,3 +31,91 @@ def decorated_testcase(*args, **kwargs):
...
+    Args:
+      ppa_record: (dictionary) A dictionary containing PPA Record.
+      pal_records: (list) A list of PAL Records in the form of dictionary
+      which has to be associated with the PPA.
+      low_frequency: (number) The Primary Low Frequency for PAL.
+      high_frequency: (number) The Primary High Frequency for PAL.
+      user_id: (string) The userId from the CBSD.
+
+    Returns:
+      A tuple containing pal records list which contains individual pal records
```
kate-harrison: Update order here?

src/harness/util.py
kate-harrison: Very nice! (I will have to remember this for future projects...)

src/harness/util.py
kate-harrison: Can't you just use `pal_records[index] = dict(pal_rec)`?

Don't you need to do this for the PPA record as well?

src/harness/util.py
```diff
@@ -106,7 +106,7 @@ def makePpaAndPalRecordsConsistent(ppa_record, pal_records, low_frequency,
     # Change Frequency Information in Pal
     pal_rec['channelAssignment']['primaryAssignment']['lowFrequency'] = low_frequency
     pal_rec['channelAssignment']['primaryAssignment']['highFrequency'] = high_frequency
-    pal_records[index] = json.loads(json.dumps(pal_rec))
+    pal_records[index] = dict(pal_rec)
```
kate-harrison: Same for ppa_record?

src/harness/util.py
sushantparkhi: Thanks for that suggestion! I think the defaultdict is not required for PPA Record because we are making changes in the root level. But in the case of PAL Record the addition of key has the nesting that's why I have used defaultdict over there.

kate-harrison: In that case, I think you want to remove this line?

```
ppa_record = defaultdict(lambda: defaultdict(dict), ppa_record)
```

sushantparkhi: My Bad! Probably I should convert it to dict. Removing the line will throw an error on ppaInfo. :)

PR #228, opened 2017-06-13
Make REG_14 configurable.

kate-harrison:
Following our discussion this morning, here's a more complete example of a configurable test.


src/harness/testcases/registration_testcase.py
```diff
@@ -569,30 +564,50 @@ def test_WINNF_FT_S_REG_14(self):
...
+    }
+    with open(config_filename, 'w') as f:
+      f.write(json.dumps(config, indent=2, sort_keys=False))
+
+  @winnforum_testcase
+  def test_WINNF_FT_S_REG_14(self):
+    """Pending registration in Array request (responseCode 200)
+    The response should be FAILURE.
+    """
+    config_filename = os.path.join('testcases', 'config',
```
sushantparkhi: How about using something like this 
`config_filename = os.path.join('testcases', 'config', '{0}.config'.format(self.id().split('.')[3]))` instead of manually putting the filename. Also when this is used we can also move the Line:589 to Line:592 to generate_REG_14_config() definition and directly returning the config?

kate-harrison:
This has been superseded by PR #348 and hence can be closed.

PR #234, opened 2017-06-14
Added WINNF.FT.S.SIR.2, HTTP Request Listener, Fake SAS Trigger for Pull

sushantparkhi:
* Added HTTP Request Listener for Pull and Push Request (sas.py and sas_interface.py)
* Added WINNF.FT.S.SIR.2
* Added Support for Pull Trigger in Fake SAS (fake_sas.py)
* Moved server.cert and server.key into cert directory
* Formatting Changes
* Updated Config File to Support HTTP Request Listener (sas.cfg)


kate-harrison:
I expect that the SIR.* test cases will be removed or made optional. However, we still need to support similar functionality so that the test harness can act as a peer SAS for the SAS under test. Given that, it's still important to work with ITS to resolve any IT issues they have with this approach.

I don't think it's necessary to review this PR but it seems good to leave it on GitHub, hence this comment so that anyone looking at the PR can understand what's going on.

kate-harrison:
I think we can safely close this PR now. James/Sushant, do you agree?

jni2000:
I agree if we do not need them any more.

sushantparkhi:
Yes, I think we can close this PR without removing Branch!

PR #237, opened 2017-06-16
Implement helper functions to 1. Register devices 2. Register devices…

harinineelu:
… and get grants.

Functions are meant to be used with config files as proposed in https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/228

### DIFF
--- a/src/harness/sas_interface.py
+++ b/src/harness/sas_interface.py
@@ -182,7 +182,7 @@ class SasInterface(object):
       ssl_cert: Path to SSL cert file, if None, will use default cert file.
       ssl_key: Path to SSL key file, if None, will use default key file.
     Returns:
-      A dictionary of Esc Sensor Data Message object specified in 
+      A dictionary of Esc Sensor Data Message object specified in
       WINNF-16-S-0096
     """
     pass
@@ -403,5 +403,45 @@ class SasTestcaseInterface(object):
     Returns:
       Nothing. It asserts if something about the response is broken/not per
       specs. Assumes it is dealing with an approved request.
+    """
+    pass
+
+  @abc.abstractmethod
+  def assertRegistered(self, registration_request,
+                       conditional_registration_data=None):
+    """Register a list of devices.
+
+    Quickly register N devices, assert registration SUCCESS, get CBSD IDs.
+    Includes injection of FCC IDs and conditional registration data.
+
+    Args:
+      registration_request: A complete registration request containing N
+        individual requests (format according to the SAS-CBSD TS).
+      conditional_registration_data: Conditional parameters (format
+        according to SasAdminInterface.PreloadRegistrationData()).
+
+    Returns:
+      A list of cbsd_ids.
+    """
+    pass
+
+  @abc.abstractmethod
+  def assertRegisteredAndGranted(self, registration_request, grant_request,
+                                 conditional_registration_data=None):
+    """Register and get grants for a list of devices.
+
+    Quickly register and grant N devices; assert SUCCESS for each step and
+    return corresponding CBSD and grant IDs.
+    Args:
+      registration_request: A complete registration request containing N
+        individual requests (format according to the SAS-CBSD TS).
+      grant_request: A complete grant request containing N individual requests
+        (format according to the SAS-CBSD TS). Note: cbsdId field should not be
+        specified.
+      conditional_registration_data: Conditional parameters, if any. (format
+        according to the SAS-CBSD TS).
+
+    Returns:
+      A list of  dictionaries, each with keys "cbsdId" and "grantId".
     """
     pass
--- a/src/harness/sas_interface.py
+++ b/src/harness/sas_interface.py
@@ -251,9 +251,9 @@ class SasAdminInterface(object):
 
     Args:
       request: A dictionary with a single key-value pair where the key is
-        "zoneData": ZoneData object to be injected into SAS under test.
-        For more information about ZoneData please see the SAS-SAS TS
-        (WINNF-16-S-0096).
+        "record" and the value is ZoneData object to be injected into 
+        SAS under test. For more information about ZoneData please see 
+        the SAS-SAS TS (WINNF-16-S-0096).
     """
     pass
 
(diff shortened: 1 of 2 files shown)


src/harness/sas_interface.py
```diff
@@ -404,4 +404,44 @@ def assertValidResponseFormatForApprovedGrant(self, grant_response):
...
+
+  @abc.abstractmethod
+  def assertRegistered(self, registration_request,
+                       conditional_registration_data=None):
+    """Register a list of devices.
+
+    Quickly register N devices, assert registration SUCCESS, get CBSD IDs.
+    Includes injection of FCC IDs and conditional registration data.
+
+    Args:
```
sushantparkhi: Can we pass a whole config file instead of individual arguments? Because if we are planning to use this with what has been proposed in https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/228, it will be difficult to know when to use this assertion as the responseCodes are part of the Config File which has some 0s and other responseCode as well.

src/harness/sas_testcase.py
```diff
@@ -45,3 +53,50 @@ def assertValidResponseFormatForApprovedGrant(self, grant_response):
...
+    list_of_cbsd_ids_and_grant_ids = []
+
+    for cbsd_id, grant_req in zip(cbsd_ids, grant_request):
+      grant_req['cbsdId'] = cbsd_id
+
+    request = {'grantRequest': grant_request}
+    grant_response = self._sas.Grant(request)['grantResponse']
+
+    # Check the grant response; collect CBSD and Grant IDs
+    for cbsd_id, grant_resp in zip(cbsd_ids, grant_response):
```
sushantparkhi: Can we use assertValidResponseFormatForApprovedGrant to validate the Grant Response?

src/harness/sas_testcase.py
```diff
@@ -45,3 +53,50 @@ def assertValidResponseFormatForApprovedGrant(self, grant_response):
...
+    if conditional_registration_data:
+      self._sas_admin.PreloadRegistrationData(conditional_registration_data)
+
+    request = {'registrationRequest': registration_request}
+    response = self._sas.Registration(request)['registrationResponse']
+
+    # Check the registration response; collect CBSD IDs
+    cbsd_ids = []
+    for resp in response:
+      self.assertEqual(resp['response']['responseCode'], 0)
```
sushantparkhi: Do we also need a check for measReportConfig ? We can add a check based on the Registration Request Dictionary; if the measCapability is empty then the measReportConfig should not exist in Response?

src/harness/sas_interface.py
kate-harrison: I don't think we want to pass the entire config file: each config file may have its own naming scheme for the parameters. For example, consider the PPA tests (recently proposed by Red and agreed to in WG4) in which they want to be able to specify a set of CBSDs that's 40 km away from the PPA and another set which is closer than 40 km. The easiest config for that would have two registration request variables, one for each category.

I think we want the helper functions to always assert/assume success as that gives us a nice tradeoff between complexity/easy of understanding and # of times the function can be reused. We could rename to something like "assertSuccessfullyRegistered" if you think that would be clearer?

src/harness/sas_testcase.py
kate-harrison: I think this is part of a larger philosophical question: do we want to do light or heavy checks of responses within the helper functions? I can see arguments in both directions:
Pro light: we've already validated the functionality more thoroughly in other tests; let's assume that this works and not double-check all of the time. If it *does* fail, it may increase debugging time. Heavy checking also means that if you have an error in e.g. requesting measurements you may fail tons of unrelated tests; it might be nice to (for the most part) only fail a test if you're missing the functionality that it's testing. Finally, heavy checking has the potential to complicate the implementation of the helper functions, possibly introducing subtle bugs (this argument is much weaker than the rest).
Pro heavy: the SAS should always be right, so the checks should always pass (and if they fail, that's super interesting). By testing more completely, we have higher confidence in SAS performance.

I think in general I'm in favor of light testing for now and possibly adding the heavy testing later. But if you feel strongly in favor of heavy testing now, we can do that instead. Let me know, and let's adjust the comments to reflect the decision so it's clear to the caller.

src/harness/sas_interface.py
sushantparkhi: Based on that, then we will only be using this Helper Function where we just need the Successful Registration before executing anything else? 

Because as mentioned in the comments by @harinineelu if we want to use this same assertRegistered like in the case of REG_14 #228 then it will get complicated because we have to make sure that we only call this assertion when the expectedResponseCode is 0.

src/harness/sas_testcase.py
sushantparkhi: As assertRegisteredAndGranted is kind of a black box where it only returns the cbsd id and the grant id list. But if we want to use assertRegisteredAndGranted and assertValidResponseFormatForApprovedGrant together like we can in the case of GRA_20, it won't be possible in the current implementation as it requires grantResponse as an argument. 

We can assume that if this Helper Function will successfully register and get a valid grant, then all the required fields in the grantResponse should exist and can be validated using assertValidResponseFormatForApprovedGrant.

src/harness/sas_interface.py
kate-harrison: Yeah, I think we only want to use these in the case where we expect successful registration (which I think should happen in 100% of cases except REG_*). It should work well with both configurable and non-configurable tests (since we just take the registrationRequest as a parameter). I believe her reference to REG_14 was to highlight that this will work with configurable tests, not that it should necessarily be used with REG_14 specifically.

src/harness/sas_testcase.py
kate-harrison: Similar to my comment above, we might not want to use this helper function in the GRA_* testcases since those will require more careful checking of the grant response. We could call assertValidResponseFormatForApprovedGrant() from within assertRegisteredAndGranted(), which may mean we can use it in a few more testcases (e.g. GRA_20); this approach has the pros and cons listed in my earlier message. I slightly prefer not calling it inside the helper function.

In general, I think we should assume that the helper functions will take care of the common case and that any uncommon case will not necessarily be able to use them.

PR #240, opened 2017-07-04
Fix LRU for NLCD indexer

gregbillock:


w4je:
Still getting a ~60% failure rate on NLCD lookups with the revised code.

PR #241, opened 2017-07-10
GerardRed_implementation_fss_cbds_gain

redtechnologies:
Reference implementation for FSS gain and CBSD gain computation

### DIFF
--- a/src/prop_current/antenna_gainsV1_0.py
+++ b/src/prop_current/antenna_gainsV1_0.py
@@ -1,0 +1,152 @@
+'''
+Created on 6 juil. 2017
+
+@author: Red Technologies
+Gerard Garnier
+'''
+import vincenty
+from ehata_its_wf import *
+from cmath import isnan
+
+def GetCBSDAntennaGain(antennaLat, antennaLon, receiverLat, receiverLon, antennaAzimut, antennaBeamWidth):
+    """
+    Returns the CBSD antenna gain (R2-SGN-20).
+    Inputs:
+        antennaLat, antennaLon    lat/lon of CBSD (deg)
+        receiverLat, receiverLon  lat/lon of Receiver (deg)
+        antennaAzimut             (deg : 0 - 359 )
+        antennaBeamWidth          3-dB antenna beamwidth (deg : 0 - 359 )
+    Output:
+        CBSD antenna gain
+    """
+    d, AZ, backaz = vincenty.dist_bear_vincenty(antennaLat, antennaLon, receiverLat, receiverLon)
+    teta = antennaAzimut - AZ
+    return -min(12 * pow(teta/antennaBeamWidth,2), 20)
+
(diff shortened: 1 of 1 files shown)


sushantparkhi:
Maybe the input parameters in CBSD gain computation can be changed to antennaAzimuth, antennaBeamWidth and aZ, instead of taking CBSD and Receiver Lat Lon? Because PPA Reference Model implementation already has the value of aZ and this will be a redundant call to vincenty.dist_bear_vincenty() function to again get the same value.

sushantparkhi:
Can you also move this to new directory harness/reference_models/${folder_name} ?
folder_name = antenna_model?

redtechnologies:
Hi Sushant,
Can you check again this new version.
Gerard

src/prop_current/antenna_gainsV1_0.py
```diff
@@ -0,0 +1,150 @@
...
+    Output:
+        theta : Elevation angles
+                 - theta[0] = mobile horizon distance, in radians
+                 - theta[1] = base station horizon distance, in radians
+    """
+    np = int(pfl[0])           #// number of points
+    xi = pfl[1]                #// step size of the profile points, in meter
+    d_meter = np * xi
+
+    h_gnd__meter = AverageTerrainHeight(pfl)
```
sbdt-google: Why not be explicit instead of referencing another module:
` h_gnd_meter = npy.mean(pfl[2:])`
Also please add on top the import
`import numpy as npy`

src/prop_current/antenna_gainsV1_0.py
```diff
@@ -0,0 +1,150 @@
...
+    
+    en0 = 301.0
+    ens = 0
+    if (h_gnd__meter == 0):
+        ens = en0
+    else:
+        ens = en0 * math.exp(-h_gnd__meter / 9460)
+    gma = 157e-9
+
+    gme = gma * (1 - 0.04665 * math.exp(ens / 179.3))
```
sbdt-google: Would add a small comment to explain these complex calculation:
`# Effective earth curvature per meter, based on refractivity ens at average altitude based on refractivity en0 at sea level`

src/prop_current/antenna_gainsV1_0.py
```diff
@@ -0,0 +1,150 @@
...
+    qc = 0.5 * gme
+    q = qc * d_meter
+    theta[1] = (zb - za) / d_meter
+    theta[0] = theta[1] - q
+    theta[1] = -theta[1] - q
+  
+    if np < 2:
+        return
+
+    sa = 0.
```
sbdt-google: I guess all the following is an extract and rewrite of the ITM horizon calculation routine hzns(), so it is fine.

redtechnologies:
Hi Serge,
I split PR 241 into two PR's (PR 250 and PR 251), one for CBSD antenna gain and one for FSS antenna gain wich your suggestions. 
Can you check again please ?
Best regard.
Gerard

sbdt-google:
Hi Gerard,
Sorry for the late answer, as I was out on vacation last week. 
Will look at the the new 2 PR shortly.
Thanks,
Serge

PR #245, opened 2017-07-12
Added WINNF.FT.S.GRA.24

sushantparkhi:


src/harness/testcases/grant_testcase.py
```diff
@@ -508,34 +508,39 @@ def test_WINNF_FT_S_GRA_24(self):
...
-    device_c['installationParam']['longitude'] = getRandomLatLongInPolygon(ppa_record)
-    device_a['userId'] = user_id
-    device_c['userId'] = user_id
+    registration_request = []
+    ppa_records = []
+    pal_records = []
+    pal_frequency_0 = [3550000000.0, 3560000000.0]
+    pal_frequency_1 = [3570000000.0, 3580000000.0]
+
+    for device_filename, pal_filename, ppa_filename, frequency in zip(
```
kate-harrison: Nice!

src/harness/testcases/grant_testcase.py
```diff
@@ -508,34 +508,39 @@ def test_WINNF_FT_S_GRA_24(self):
...
+                                                              [pal_record],
+                                                              frequency[0],
+                                                              frequency[1],
+                                                              device['userId'])
+
+      # Move the Device to a random location in PPA
+      device['installationParam']['latitude'], \
+      device['installationParam']['longitude'] = getRandomLatLongInPolygon(ppa_record)
+      ppa_records.append(ppa_record)
+      pal_records.append(pal_record[0])
```
kate-harrison: While the naming makes it look confusing, I think you actually want `pal_records = pal_records + pal_record` (so that it will work in the case when you have multiple PAL records... just in case this is one of the test cases that becomes configurable).

src/harness/testcases/grant_testcase.py
```diff
@@ -508,34 +508,39 @@ def test_WINNF_FT_S_GRA_24(self):
...
+      self._sas_admin.InjectFccId({'fccId': device['fccId']})
+      registration_request.append(device)
 
     # Register the devices
-    request = {'registrationRequest': [device_a, device_c]}
+    request = {'registrationRequest': registration_request}
```
kate-harrison: Use the helper function to do the registration + collect CBSD IDs for you?

src/harness/testcases/grant_testcase.py
```diff
@@ -544,10 +549,11 @@ def test_WINNF_FT_S_GRA_24(self):
...
     del request, response
 
-    # Update PPA Record with CBSD ID and Inject Data
-    ppa_record['ppaInfo']['cbsdReferenceId'] = cbsd_ids
-    self._sas_admin.InjectZoneData({"record": ppa_record})
-    self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+    for ppa_record, pal_record, cbsd_id in zip(ppa_records, pal_records, cbsd_ids):
+      # Update PPA Record with CBSD ID and Inject Data
+      ppa_record['ppaInfo']['cbsdReferenceId'] = [cbsd_id]
+      self._sas_admin.InjectZoneData({"record": ppa_record})
```
kate-harrison: Swap the order here (see my comment on one of your other PRs).

src/harness/testcases/testdata/pal_record_1.json
```diff
@@ -0,0 +1,12 @@
+{
+  "censusYear": 2010,
+  "fipsCode": 20063955100,
+  "licensee": "CBSD Operator No.1",
+  "registrationInformation": {
+    "frn": "test_fcc_frn_pal_a"
```
kate-harrison: "...pal_1"? I'm not sure why you're using 'a' here.

src/harness/testcases/grant_testcase.py
```diff
@@ -501,6 +501,88 @@ def test_WINNF_FT_S_GRA_23(self):
...
+    cbsd_ids = []
+    for resp in response:
+      self.assertEqual(resp['response']['responseCode'], 0)
+      cbsd_ids.append(resp['cbsdId'])
+    del request, response
+
+    for ppa_record, pal_record, cbsd_id in zip(ppa_records, pal_records, cbsd_ids):
+      # Update PPA Record with CBSD ID and Inject Data
+      ppa_record['ppaInfo']['cbsdReferenceId'] = [cbsd_id]
+      self._sas_admin.InjectZoneData({"record": ppa_record})
```
kate-harrison: Re-order here, as mentioned in a different PR.

src/harness/testcases/grant_testcase.py
sushantparkhi: I can use `pal_records = pal_records + pal_record` or `pal_records.extend(pal_record)` but I think it's safe to assume that L528 will return a list with one PAL Record as we are only passing one PAL in the list to make it consistent with PPA.

If we want to make it "configurable" then I think we need to change the whole loop structure here and below. For example, this loop might look like

`for device_filename, pal_filename, ppa_filename, frequency in zip(
      ('device_a.json', 'device_c.json'),
      ('ppa_record_0.json', ['pal_record_0.json', 'pal_record_x.json'...]),
      ('ppa_record_1.json', ['pal_record_1.json', 'pal_record_x.json'...]),
      (pal_frequency_0, pal_frequency_1)):...`

Let me know your thoughts

kate-harrison:
I believe this has been deprecated by WG4, but maybe it's worth keeping the PR up because the test scenario may have been merged into another test case.

janetliuboston:
Let's close this so it's not cluttering up work that needs to be done. We can refer to this closed PR if needed. Thanks.

PR #247, opened 2017-07-12
Added WINNF.FT.S.SIQ.18

sushantparkhi:

PR #248, opened 2017-07-18
Fixed ppa Helper function by adding censusInfo and channel in Pal Record

sushantparkhi:
Removed censusInformation (fipsCode and censusYear) from PPA and added that into Pal Record with fccChannelId to allow multiple PALs.

### DIFF
--- a/src/harness/util.py
+++ b/src/harness/util.py
@@ -107,7 +107,7 @@ def makePpaAndPalRecordsConsistent(ppa_record, pal_records, low_frequency,
     # Change Frequency Information in Pal
     pal_rec['channelAssignment']['primaryAssignment']['lowFrequency'] = low_frequency
     pal_rec['channelAssignment']['primaryAssignment']['highFrequency'] = high_frequency
-    # Convert from defaultDict to dict
+    # Converting from defaultdict to dict
     pal_records[index] = json.loads(json.dumps(pal_rec))
 
   # Add Pal Ids into the Ppa Record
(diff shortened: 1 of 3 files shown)


janetliuboston:
Red was concerned that (PR #247 ) PAL shouldn't have the same census tract (FIPS code) and frequency (channel ID) information for all PALs. Each PAL should have their own FIPS code and channel ID. Possibly the channel ID could be put back as a parameter, as the PPA can only operate on the one channel across all the PALs, but the census tract should be different.

src/harness/util.py
```diff
@@ -84,14 +85,13 @@ def makePpaAndPalRecordsConsistent(ppa_record, pal_records, low_frequency,
...
     pal_fips_code = pal_rec['fipsCode']
     pal_census_year = pal_rec['censusYear']
-    pal_fcc_channel_id = pal_rec['fccChannelId']
-    del pal_rec['fipsCode'], pal_rec['censusYear'], pal_rec['fccChannelId']
+    del pal_rec['fipsCode'], pal_rec['censusYear']
 
     pal_rec = defaultdict(lambda: defaultdict(dict), pal_rec)
     # Change the FIPS Code and Registration Date-Year in Pal Id
     pal_rec['palId'] = '/'.join(['pal', '%s-%d' % ('{:02d}'.format(previous_year_date.month),
                                                    previous_year_date.year), str(pal_fips_code),
```
kate-harrison: Consider reformatting (wrapping the line at 80 characters).

PR #251, opened 2017-07-21
GerardRed_fss_antenna_gain

redtechnologies:


src/prop_current/fss_antenna_gainsV1_1.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    Returns elevation angles of the horizons from each terminal.
+    extract and rewrite of the ITM horizon calculation routine hzns()
+    Inputs:
+        pfl          Terrain profile line
+        fssHeight   height of the mobile, in meters
+        cbsdHeight   height of the base station, in meters
+    Output:
+        theta : Elevation angles
+                 - theta[0] = mobile horizon distance, in radians
```
sbdt-google: It seems to me that the the theta calculated here are slopes not angles in radians.

src/prop_current/fss_antenna_gainsV1_1.py
```diff
@@ -0,0 +1,138 @@
...
+    
+    return FSSGain
+
+def GetElevationAngle(pfl, fssHeight, cbsdHeight, theta):
+    
+    """
+    Returns elevation angles of the horizons from each terminal.
+    extract and rewrite of the ITM horizon calculation routine hzns()
+    Inputs:
+        pfl          Terrain profile line
```
sbdt-google: add: "(from fss to cbsd)"

src/prop_current/fss_antenna_gainsV1_1.py
```diff
@@ -0,0 +1,138 @@
...
+    # If the FSS earth station registration data includes values for w1 and w2, 
+    # SAS shall use these values. Otherwise SAS shall assume w1=0 and w2=1.
+    if w1 is None or w2 is None:
+        FSSGain = GgsoP
+    else:
+        FSSGain = w1*Ggso + w2*GgsoP
+    
+    return FSSGain
+
+def GetElevationAngle(pfl, fssHeight, cbsdHeight, theta):
```
sbdt-google: Would be more pythonic if the function returns the theta tuple, ie no need to pass it as input. You can simply work with 2 variable theta0 and theta1 in the function, then do a "return theta0,theta1" at the end.

src/prop_current/fss_antenna_gainsV1_1.py
```diff
@@ -0,0 +1,138 @@
...
+        FSS Earth Station Antenna Gain
+    """
+    d, azstart, AZ = vincenty.dist_bear_vincenty(cbsdLat, cbsdLng, fssLat, fssLng)
+
+    theta = [0., 0.]
+    GetElevationAngle(pfl, fssHeight, cbsdHeight, theta)
+    
+            
+    # off-axis angle
+    ThetaAngle = (180 * (math.acos( (math.cos(theta[0]) * math.cos(math.radians(fssPointingElevation))\
```
sbdt-google: I double checked the formula. Seems good to me.

sbdt-google:
Marking as deprecated as the code has been extended and integrated with reference propagation models in PR:
https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/283

PR #253, opened 2017-07-24
Added PPA Reference Model

sushantparkhi:


src/harness/reference_models/ppa_model/utils.py
```diff
@@ -0,0 +1,154 @@
...
+from src.harness.util import getRandomLatLongInPolygon
+from collections import defaultdict
+from datetime import datetime
+import random
+import json
+import os
+import glob
+
+
+def to_dist_bear_vincenty(**kwargs):
```
gregbillock: Andy checked in a vincenty module in geo/vincenty.py

There is also the pygc package which is a third party implementation.

src/harness/reference_models/ppa_model/utils.py
```diff
@@ -0,0 +1,154 @@
...
+  """Placeholder to Call Vincenty, will be removed when the Vincenty is moved to reference models"""
+  pass
+
+
+def great_circle(**kwargs):
+  latitude = kwargs.pop('latitude')
+  longitude = kwargs.pop('longitude')
+  distance = kwargs.pop('distance') # In Km.
+  azimuth = kwargs.pop('azimuth')
+  calc = np.vectorize(to_dist_bear_vincenty)
```
gregbillock: What's the damage to taking out the curried functions and vectorization and just letting the python spin longer? I'm sure it's a big hit relatively speaking, but even if it goes from 0.1s to 5s or something its probably not that big a deal unless we're doing lots and lots of these.

sushantparkhi: In the current implementation, I am calling vincenty 360 times but if I remove the vectorization then the call will increase to 360*200 i.e. I need to call this for every point. If not then I can use pygc, it has the same implementation.

gregbillock: Andy's and pygc are virtually identical -- I think pygc has a vector method IIRC.

gregbillock: (Note Andy suggested to WG1 that we just not use vincenty and use haversine methods instead. I think that's the real solution to this vincenty craziness.)

src/harness/reference_models/ppa_model/utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def convert_shape(polygon):
+  ppa = shape(polygon).buffer(0)
+  return ppa
+
+
+def smooth(x, window_len=15):
+  s = np.r_[x[-1:-window_len / 2:-1], x, x[window_len / 2:0:-1]]
+  w = np.hamming(window_len)
```
jcereced: File: util.py  (https://raw.githubusercontent.com/Wireless-Innovation-Forum/Spectrum-Access-System/50cc119c7a585aeb5319de8736fed06a34941d06/src/harness/reference_models/ppa_model/utils.py)
Function: smooth
Line 61 s = np.r_[x[-1:-window_len / 2:-1], x, x[window_len / 2:0:-1]] appears to be wrong. It seems the intent is s = np.r_[x[-(window_len / 2):], x, x[0:window_len/ 2]] which is padding the sequence x with cyclically.
 
If X is an array 0:24
 
With an array 0 to 24, the current implementation gives
[24 23 22 21 20 19 18  0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24  7  6  5  4  3  2  1]
 
It should be cyclical
 
[18 19 20 21 22 23 24 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24 0 1 2 3 4 5 6]

sushantparkhi: Thanks for noting that! How about something like `s = np.r_[x[-window_len:],x,x[:window_len]]`?

src/harness/reference_models/ppa_model/utils.py
sbdt-google: Note that the new implementation of reference model has an update vincenty.py, which provides a "to_dist_bear" vectorized for a list of distances (for single bearing direction) - see function GeodesicPoints() - function names have been changed to follow the convention in harness code).
It is much faster than calling the dist_bear as a python loop (or np.vectorize which actually do a loop).

sbdt-google:
Can you adapt your code to use the reference_models geo and propagation stuff.
Then I will be glad to review your PR in more details.

xwu0823:
Hi,
We already implemented reference_models geo and propagation stuff. Could you help to do a review?

sbdt-google:
Ok I will start the review if you have updated the code.

jni2000:
Thanks, we have updated the codes a while before and there are no more update as far as we are aware of. Please continue the review. Thanks a lot.

James Ni
Director of SAS System
978.996.7480

jni2000:
please can we speedup this review, All the PCR test cases are depending on this. This has been here for a long while. Thanks,

sbdt-google:
I hope to finish this afternoon. Will let you know.

jni2000:
thanks.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,82 @@
...
+    lon.append(lo)
+  polygon = geojson.Polygon([zip(lon,lat)])
+  return polygon
+
+
+def ppa_creation_model(device_filenames, pal_record_filenames,
+                       pal_user_id, pal_low_frequency, pal_high_frequency):
+  """Create a PPA Polygon based on the PAL Records and Device Information
+  Args:
+    device_filenames: (List) A list containing device filenames.
```
kate-harrison: We need to take a list of registrationRequests and palRecords, not the filenames. Loading by filename (especially considering the post-processing that is done later) is not flexible enough to do what is needed.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,82 @@
...
+    pal_record_filenames: (List) A list containing pal record filenames.
+    pal_user_id: (String) Pal User Id to be used in all Pal Records.
```
kate-harrison: I'm not sure this (and the frequency inputs) is necessary.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,82 @@
...
+  Args:
+    device_filenames: (List) A list containing device filenames.
+    pal_record_filenames: (List) A list containing pal record filenames.
+    pal_user_id: (String) Pal User Id to be used in all Pal Records. 
+    pal_low_frequency: (Number) Pal Primary Low Frequency to be assigned 
+    to all Pal Records. 
+    pal_high_frequency: (Number) Pal Primary High Frequency to be assigned 
+    to all Pal Records. 
+  Returns:
+     A Tuple containing PPA Polygon in GeoJSON, a List of
```
kate-harrison: Output should only be the PPA polygon.

src/harness/reference_models/ppa_model/utils.py
```diff
@@ -0,0 +1,146 @@
...
+      census_tracts = get_census_tracts(pal_record['license']['licenseAreaIdentifier'])
+      device['installationParam']['latitude'], \
+      device['installationParam']['longitude'] = getRandomLatLongInPolygon(census_tracts)
+      devices.append(device)
+
+  return devices, pal_records
+
+
+def get_census_tracts(fips_code):
+  for filename in glob.glob(os.path.join('census_tracts', '*.json')):
```
kate-harrison: Are all of the census tracts checked into GitHub? We may need to add them to the [SAS-Data](https://github.com/Wireless-Innovation-Forum/SAS-Data) repo so that everything works correctly.

src/harness/reference_models/ppa_model/compute.py
```diff
@@ -0,0 +1,84 @@
...
+from math import log10
+from functools import partial
+import numpy as np
+from utils import great_circle
+from reference_models.antenna.antenna import GetStandardAntennaGains
+from reference_models.geo import nlcd
+from reference_models.propagation.wf_hybrid import CalcHybridPropagationLoss
+RX_HEIGHT = 6
+
+def prop_model(**kwargs):
```
kate-harrison: Is this still needed?

src/harness/reference_models/ppa_model/ppa_ref_model.py
kate-harrison: To be clear: we *cannot* be doing the "make PAL records consistent" kind of work in this reference model -- the behavior is totally unexpected and it will not match what the SAS UUT will do when given the same data (which is the definition of a reference model).

sushantparkhi: Instead of doing it in the reference model, do you think it will be a good idea to create a utility function as a part of a reference model, which will call the `makePpaAndPalRecordsConsistent` and `getRandomLatLongInPolygon`? Or leave it up to the Test Dev to call these functions if necessary? If that is the case then the reference model will always assume that the dictionaries provided are always correct(we can have some checks in place).

src/harness/reference_models/ppa_model/utils.py
jni2000: we have the census tract data, where should they go ?

src/harness/reference_models/ppa_model/ppa_ref_model.py
jni2000: Kate, based on your comment we will need some refactoring and some thinking as to how to return the updated Device Info. Since in the Reference Model itself there’s function makePpaAndPalRecordsConsistent (Janet and Sushant had a long discussion with you regarding the implementation of this function) which basically add all the required parameters for the PPA and PAL. If we don’t do it in reference model, then every time the PPA Reference Model is called we will have to call this function and then we will have to pass the PPA and PAL Record Dictionary as a parameter instead of file names. Separating the two can potentially cause misusing of the functions when someone new to the test code development and want to call the reference model without knowing the fact that the two must be together and the  makePpaAndPalRecordsConsistent function must be call ahead of the reference mode call.

We thought of having it more unified, but if your suggested way is a better solution for good reasons, we will modify the code then.

Thanks

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,82 @@
...
+import multiprocessing
+from utils import *
+from compute import load_partials
+from reference_models.geo.vincenty import GeodesicPoint
+
+THRESHOLD = -96
+
+
+def __compute_points_gain_prop(device):
+  pool = Pool(multiprocessing.cpu_count())
```
sbdt-google: The pool is a common resource and should not be created for each call as it is expensive.
It can be created either as a global, or passed to the function (and created at the caller).

Also note that the caller is itself calling __compute_points_gain_prop() with a pool, sending every device calculation in parallel. It is a bad idea to nest such pool use, and can create issues, contention and no performance gain in fact.

Also note that the python multiprocessing pool is using system multi process as multithreading does not give any advantage in Python. Multi process means that instance of Python process are created multiple time, and there is some heavy  communication to pass the data from the master process to the child process. It is thus interesting for some heavy calculation, and not very small calculation.
I am not sure some of the calculation below are big enough to require such an heavy artillery.

If the performance is not so much an issue for these tests, then we could simply live with the top level multiprocessing at the device stage, and keep simple `map` here.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,82 @@
...
+  install_param = device['installationParam']
+  # Get all the Function Partials
+  great_circle_partial, antenna_gain_partial, \
+  prop_partial, cnt_partial = load_partials(install_param, THRESHOLD)
+  # Compute all the Points in 0-359 every 200m upto 40km
+  points = pool.map(great_circle_partial, range(0, 360))
+  # Compute the Gain for all Direction
+  antenna_gain = pool.map(antenna_gain_partial, range(0, 360))
+  # Compute the Path Loss
+  path_loss = pool.map(prop_partial, enumerate(points))
```
sbdt-google: Only this one is a bit heavy in terms of calculation, other are very fast and the multiprocessing may in fact slow things down. See my previous comment.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,82 @@
...
+   Note: Device Records must contain eirpCapability in installationParam Object
+   and Pal Records must contain fipsCode
+  """
+  devices, pal_records = load_files(device_filenames, pal_record_filenames,
+                                   pal_user_id, pal_low_frequency, pal_high_frequency)
+  pool = Pool(multiprocessing.cpu_count())
+  polygon_device = \
+    pool.map(__compute_points_gain_prop, devices)
+  polygons = map(convert_shape, polygon_device)
+  contour_union = cascaded_union(polygons)
```
sbdt-google: See earlier comment: _compute_points_gain_prop should return a shapely polygon, instead converting it to geojson, then converting it back to shapely. Lets stay as much as possible with shapely polygon since the geometrical operation are done with it, and use geojson just as input/output format.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,82 @@
...
+    to all Pal Records. 
+  Returns:
+     A Tuple containing PPA Polygon in GeoJSON, a List of 
+     Consistent PAL Record and Device List
+   Note: Device Records must contain eirpCapability in installationParam Object
+   and Pal Records must contain fipsCode
+  """
+  devices, pal_records = load_files(device_filenames, pal_record_filenames,
+                                   pal_user_id, pal_low_frequency, pal_high_frequency)
+  pool = Pool(multiprocessing.cpu_count())
```
sbdt-google: I would keep the usage of the pool only here, where you parallelize per device. Doing further fine grain parallelization inside __compute_points_gain() should not give any additional speed gain (and actually probably some slow down due to contention and context switch), provided that number of CBSD  > number of CPU.
But in any case if only small number of CBSD, the calculation will be fast enough so no real need of fancy speed up.

src/harness/reference_models/ppa_model/utils.py
kate-harrison: You can add them to the repository I mentioned in my first comment.

src/harness/reference_models/ppa_model/ppa_ref_model.py
kate-harrison: James,

I agree that the function makePpaAndPalRecordsConsistent() is a very useful function and that it should be used in the test cases. However, it is inappropriate to use it in a reference model.

A reference model should, by definition, have the same behavior as a "model" SAS when given the same data. If you tried to claim a PPA with conflicting PAL records, a correct SAS should give an error rather than silently aligning the data and ignoring the problem.

It is the responsibility of the code/configuration which calls the reference model to make sure that the inputs make sense. If the inputs do not make sense, the reference model should return an error and exit. The reference model should do *exactly* what a SAS would do in the same situation, full stop. If it does anything else, it is no longer a reference model.

Based on his comment, I think Sushant understands what I'm trying to say (i.e. that it's test dev's responsibility to call these helper functions before calling the reference model, and that the reference model should do at least some error checking) so perhaps he can explain further if you still have questions.

Kate

src/harness/reference_models/ppa_model/compute.py
```diff
@@ -0,0 +1,84 @@
...
+					  longitude=install_param['longitude'],
+					  distance=np.arange(0.2, 40.2, 0.2),
+					  azimuth=azimuth)
+
+
+def __compute_cnt(install_param, threshold, gain_prop_tuple):
+  gain, path_loss = gain_prop_tuple
+  cnt = 0
+  for pl in path_loss:
+	if (install_param['eirpCapability'] - pl) + gain > threshold:
```
arhannan: install_param['eirpCapability']  should be adjusted by install_param['antennaGain']  before use in computing the interference power. otherwise, we'll be adding the antennaGain twice

jni2000:
sbdt-google: We have finished the revision of teh PPA reference model codes based on your comments and the interface we agreed on. Please give a review and approve it if there is no issues. Thanks.

sushantparkhi:
@sbdt-google @kate-harrison 
Sorry lot of commits/changes!
*Following are the high-level changes made to the Reference Model:*
1. Refactored the whole code, removed unnecessary multiprocessing Pool calls, and placeholders.
2. Added Census Tract Driver as a part of geo package (If required I can create separate PR for this)
3. Changed interface to support dictionary of devices and pal records instead of file names.
4. Used Shapely Geometry wherever possible instead of converting it to GeoJSON
5. Changed the way modules were imported.
6. Incorporated #PR392 (Area Function and Hole Removal) function.
7. Changed the Function names from Snake Case to CamelCase and variable names to make it more readable.
8. Changed Receiver Height to 1.5m.
9. Added Nlcd and Census Tract Driver Configuration Helper Function.
*Todo:*
1. Add Validation for input parameters.
2. Add example of how to use this reference model with some test data.
3. Add Census Tract data to SAS-Data Repository.
4. Add Documentation (README.md)

sbdt-google:
Thanks Sushant. I will review by this week-end. 
Meanwhile can you review the small PR #397 which has been pending for a while.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,145 @@
...
+nlcd_driver = nlcd.NlcdDriver()
+
+
+def _CalculateDbLossForEachPointAndGetContour(install_param, antenna_gain, latitude, longitude):
+  db_loss = np.zeros(len(latitude), dtype=np.float64)
+  for index, lat_lon in enumerate(zip(latitude, longitude)):
+    lat, lon = lat_lon
+    code = nlcd_driver.GetLandCoverCodes(lat, lon)
+    region_type = nlcd.GetRegionType(code)
+    db_loss[index] = wf_hybrid.CalcHybridPropagationLoss(install_param['latitude'],
```
sbdt-google: You do not specify the reliability here.
See my previous comment on last round:
 - The calculation shall use the median path loss, so we should specify 'reliability=0.5'
 - the clarification of this is currently being done in WG1 main spec, to remove any doubts, and it should pass soon.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,145 @@
...
+from shapely import geometry, ops
+
+THRESHOLD_PER_10MHZ = -96
+RX_HEIGHT = 1.5
+
+census_tract_driver = census_tract.CensusTractDriver()
+nlcd_driver = nlcd.NlcdDriver()
+
+
+def _CalculateDbLossForEachPointAndGetContour(install_param, antenna_gain, latitude, longitude):
```
sbdt-google: Note that in general we use plural form for "a bunch of thing" and use the singular for a single objects. This makes it more easy for readability as we directly identify sequence type.

In that case, and elsewhere in the code, it would be: latitudes, longitudes

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,145 @@
...
+    code = nlcd_driver.GetLandCoverCodes(lat, lon)
+    region_type = nlcd.GetRegionType(code)
+    db_loss[index] = wf_hybrid.CalcHybridPropagationLoss(install_param['latitude'],
+                                                         install_param['longitude'],
+                                                         install_param['height'],
+                                                         lat, lon, RX_HEIGHT,
+                                                         install_param['indoorDeployment'],
+                                                         region=region_type).db_loss
+
+  db_loss = 10.0 ** (db_loss / 10.0)
```
sbdt-google: I think there is a mistake here:
You are computing a linear value here, then adding the linear value to the eirp below.
No need of this conversion to linear here.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,145 @@
...
+  index_cond, _ = np.where(install_param['eirpCapability'] - db_loss + antenna_gain > THRESHOLD_PER_10MHZ)
+  return index_cond.shape[0] / 5
```
sbdt-google: I think one my previous comment was not taken into account:

>>> Note that there is an issue here: you divide an integer number by another integer number, so you are make a C-style integer division : 7/5 = 1 (this is the case because we use Python 2.7, and we dont specify from future import division on top of the file).
You should replace with a floating point division:
  index_cond.shape[0]/5.

src/harness/reference_models/ppa_model/ppa_ref_model.py
sushantparkhi: I updated the code wherever possible and made in plural form. Let me know if I missed anywhere.

src/harness/reference_models/ppa_model/ppa_ref_model.py
sushantparkhi: Thanks for catching that! I updated it with your recommendation.

src/harness/reference_models/ppa_model/ppa_ref_model.py
```diff
@@ -0,0 +1,145 @@
...
+    region_type = nlcd.GetRegionType(code)
+    db_loss[index] = wf_hybrid.CalcHybridPropagationLoss(install_param['latitude'],
+                                                         install_param['longitude'],
+                                                         install_param['height'],
+                                                         lat, lon, RX_HEIGHT,
+                                                         install_param['indoorDeployment'],
+                                                         region=region_type).db_loss
+
+  db_loss = 10.0 ** (db_loss / 10.0)
+  index_cond, _ = np.where(install_param['eirpCapability'] - db_loss + antenna_gain > THRESHOLD_PER_10MHZ)
```
arhannan: Sushant,
install_param['eirpCapability'] should be adjusted by install_param['antennaGain'] before use in computing the interference power. otherwise, we'll be adding the antennaGain twice

src/harness/reference_models/ppa/ppa.py
```diff
@@ -0,0 +1,150 @@
...
+  """
+  # TODO: Add Validation for Inputs
+  pool = futures.ProcessPoolExecutor(multiprocessing.cpu_count())
+  # Create Contour for each CBSD
+  device_polygon = list(pool.map(_GetPolygon, devices))
+  # Create Union of all the CBSD Contours and Check for hole
+  # after Census Tract Clipping
+  contour_union = ops.cascaded_union(device_polygon)
+  ppa_polygon = _ClipPpaByCensusTract(contour_union, pal_records)
+  return _ConvertToGeoJson(utils.PolyWithoutSmallHoles(ppa_polygon))
```
kate-harrison: Isn't there also a smoothing function that is applied (before? after?) the holes are removed?

src/harness/reference_models/ppa/ppa.py
```diff
@@ -0,0 +1,150 @@
...
+    census_tract_driver.SetCensusTractDirectory(census_tract_dir)
+
+
+def ConfigureNlcdDriver(nlcd_dir=None, cache_size=None):
+  """Configure the NLCD driver.
+
+  Inputs:
+    nlcd_dir: if specified, changes the NLCD Data default directory.
+    cache_size:  if specified, change the NLCD tile cache size.
+  Note: The memory usage is about cache_size * 50MB.
```
sbdt-google: In fact the memory usage is about cache_size * 12MB

src/harness/reference_models/ppa/ppa.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+def _HammingFilter(x, window_len=15):
+  s = np.r_[x[-(window_len / 2):], x, x[0:window_len / 2]]
+  w = np.hamming(window_len)
+  y = np.convolve(w / w.sum(), s, mode='valid')
+  return y
+
+
+def _GetPolygon(device):
```
sbdt-google: Small docstring would be useful, for example:
"""Returns the PPA contour for a single CBSD device, as a shapely polygon."""

src/harness/reference_models/ppa/ppa.py
```diff
@@ -0,0 +1,150 @@
...
+def _ClipPpaByCensusTract(contour_union, pal_records):
+  # Get the Census Tract for Each Pal Record and Convert it to Shapely Geometry
+  census_tracts_for_pal = [geometry.shape(census_tract_driver.GetCensusTract(pal['license']
+                                                                             ['licenseAreaIdentifier'])
+                                          ['features'][0]['geometry']).buffer(0) for pal in pal_records]
+  census_tracts_union = ops.cascaded_union(census_tracts_for_pal)
+  return contour_union.intersection(census_tracts_union)
+
+
+def _ConvertToGeoJson(ppa_polygon):
```
sbdt-google: This and previous function could have a onliner docstring description.
I know they are intended as private function, but a one liner is still useful, for example:
"""Converts a shapely Polygon or MultiPolygon into a GeoJSON feature collection"""

and for previous
""" Clip a PPA 'contour_union' zone (shapely.MultiPolygon) with the census tracts defined by a sequence of 'pal_records'."""

src/harness/reference_models/ppa/ppa.py
sbdt-google: I think this is taken care of:  smoothing by the Hamming filter applied on each individual device, and removal of the small holes of the final contour.

src/harness/reference_models/ppa/ppa_example.py
```diff
@@ -0,0 +1,35 @@
+import json
```
sbdt-google: Please add the Winnforum copyright section

sushantparkhi:
Hello Serge, 

Thanks for the comments! I have updated the PR, let me know if I have missed something.

jni2000:
Serge, thanks for your thorough review. We have made necessary changes to address your concern, Please take a look. If this is all right, please approve and allow us moving forward to the PCR cases.

src/harness/reference_models/ppa/ppa.py
```diff
@@ -0,0 +1,164 @@
...
+  latitudes, longitudes, _ = zip(*[vincenty.GeodesicPoints(install_param['latitude'],
+                                                           install_param['longitude'],
+                                                           distances, azimuth)
+                                   for azimuth in azimuths])
+  # Compute the Gain for all Direction
+  antenna_gains = antenna.GetStandardAntennaGains(azimuths,
+                                                  install_param['antennaAzimuth']
+                                                  if 'antennaAzimuth' in install_param.keys() else None,
+                                                  install_param['antennaBeamwidth'],
+                                                  install_param['antennaGain'])
```
sbdt-google: The comment from Ariful Hannan has not been taken into account. Because we compute the signal level as:
 ``` eirp + antenna_gain - pathloss ```
so we are counting twice the global antenna gain.
Simple solution is to change this line as:
```
norm_antenna_gains = antenna_gains = antenna.GetStandardAntennaGains(azimuths,
                                                  install_param['antennaAzimuth']
                                                  if 'antennaAzimuth' in install_param.keys() else None,                                                  
                                                 install_param['antennaBeamwidth'],
                                                 0)

jni2000:
arhannan and sbdt-google

I believe Sushant has addresses all your concern and made necessary revision in this new submission, Please continue reviewing and let me know if we missed anything, Thanks.

sbdt-google:
I am ok for submitting the code. Let's get Ariful on board, I will send hima quick message.

Serge

sbdt-google:
To me, looks ok, but I will defer this check to Kate, as it relates to compliance with the harness test framework.

jni2000:
Thanks.

PR #261, opened 2017-08-04
Add InjectFccId to GRA_20

harinineelu:


sushantparkhi:
Thanks for correcting that! Maybe use of assertRegisteredAndGranted instead https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/sas_testcase.py#L77, then the code between L362 and L382 can be removed?

harinineelu:
Thanks for the quick review, Sushant. 
Regarding assertRegisteredAndGranted - It probably makes sense to leave it alone for now so that it matches the rest in the file; if we want to change one, we should probably change the others at the same time.

sushantparkhi:
Sounds Good!

PR #264, opened 2017-08-10
Daily Activities Trigger and Status Interface

sushantparkhi:


src/harness/sas_interface.py
```diff
@@ -374,6 +374,21 @@ def TriggerPpaCreation(self, request):
     """
     pass
 
+  def TriggerDailyActivitiesImmediately(self):
+    """SAS admin interface to trigger daily activities immediately which will
+    execute the activities mentioned in the SAS admin agreement
```
kate-harrison: It's probably best to spell them out explicitly so that (1) the information is available to everyone and (2) it's very clear what's going on here (since we actually differ from the doc that Masoud et al. are working on with SAS admins).

src/harness/sas_interface.py
```diff
@@ -374,6 +374,21 @@ def TriggerPpaCreation(self, request):
...
+    """
+    pass
+
+  def GetDailyActivitiesStatus(self):
+    """SAS admin interface to get the daily activities status
+    Returns:
+      A dictionary with single key-value pair where the key is "status" and the
+      value is a string of permitted enumerations specified in SAS Under Testing
```
kate-harrison: See my comment on the testing API doc re: enum vs. boolean. I think we can simplify our lives :)

src/harness/sas_interface.py
sushantparkhi: Updated the Comments for Trigger and added all the activities to be executed.

src/harness/sas_interface.py
sushantparkhi: I have posted a comment regarding this on Testing API Doc!

src/harness/fake_sas.py
```diff
@@ -247,6 +247,12 @@ def TriggerPpaCreation(self, request, ssl_cert=None, ssl_key=None):
     return 'zone/ppa/fake_sas/%s/%s' % (request['palIds'][0]['palId'],
                                         uuid.uuid4().hex)
 
+  def TriggerDailyActivitiesImmediately(self):
+    pass
+
+  def GetDailyActivitiesStatus(self):
+    return {'status': 'completed'}
```
kate-harrison: Update?

sushantparkhi: Done!

PR #265, opened 2017-08-10
Modify GRA_28 for "allowed channel widths for Grant" rule

harinineelu:
Allowed channel widths for a Grant are 10, 15, 20, ... MHz


kate-harrison:
FYI SAS admins agreed that the allowed channel bandwidths for grant requests are 10, 15, 20, 25, … MHz. Masoud at Federated is the owner of the doc which documents this decision (although of course the doc will eventually be made public within WinnForum); please email him if you want to check the source.

I believe that PR #265 is the only change needed to already-checked-in code.

PR #268, opened 2017-08-15
Update HBT testcase names to match spec

harinineelu:


harinineelu:
Removed spectrum_inquiry_testcase.py changes that were added here by mistake. (Thats a different PR #267)

PR #269, opened 2017-08-15
Certificate generation script (for Linux) enhanced

kharekuldeep:
Corrected OID values for SAS extension
Added support for creating EE  certificates based on pre-defined Private Key/CSR


kharekuldeep:
@gregbillock I have done changes in the certificate generation script and openssl.cnf file to align it with CBRS PKI policy spec.Request you to please review the changes.

cert/CACreationScript.sh
```diff
@@ -44,6 +48,22 @@ ROOT_RSA_KEY_SIZE=4096
 INT_RSA_KEY_SIZE=4096
 EE_RSA_KEY_SIZE=2048
 
+if [ $# -eq 2 ]
+then
+  if [ "$1" == "pkey" ]
+  then
+    cp $2 cbsd_req.pkey  
+    #cp $2 cbsd_oem_req.pkey
```
gregbillock: please delete commented lines

cert/CACreationScript.sh
```diff
@@ -18,6 +18,10 @@
 
 echo Run this script from the directory containing the CBRS openssl.cnf file present in it. The openssl command must be present on the path.
 
+echo Usage: ./CACreationScript.sh
+echo Usage: ./CACreationScript.sh pkey cbsd_ee.pkey
```
gregbillock: I'm unclear what these variants do. It looks like it's saving a particular key file to a target before running the rest of the script? Couldn't this just be done in the command line before launching the creation script?

cert/CACreationScript.sh
```diff
@@ -119,7 +139,13 @@ openssl ca -create_serial -cert $ROOT.crt -keyfile $ROOT.pkey -outdir $CSR -batc
 CBSD_EE=cbsd_req
 CBSD_EE_SIGN=$CBSD_EE$SIG
 
-openssl req -new -newkey rsa:$EE_RSA_KEY_SIZE -nodes -reqexts $CBSD_EE -out $CBSD_EE.csr -keyout $CBSD_EE.pkey -subj "/C=US/ST=District of Columbia/L=Washington/O=Wireless Innovation Forum/OU=www.wirelessinnovation.org/CN=CBSD End-Entity Example" -config openssl.cnf
+if [ -f $CBSD_EE.pkey ]
+then
+  openssl req -new -nodes -reqexts $CBSD_EE -out $CBSD_EE.csr -key $CBSD_EE.pkey -subj "/C=US/ST=District of Columbia/L=Washington/O=Wireless Innovation Forum/OU=www.wirelessinnovation.org/CN=CBSD End-Entity Example" -config openssl.cnf
```
gregbillock: If you're generating the CSR as an example, why not just always generate it? If it needs the cbsd_req.pkey just do so after that gets created.

cert/openssl.cnf
```diff
@@ -25,12 +25,12 @@
 oid_section		= new_oids
 
 [ new_oids ]
-ROLE_CA=1.3.6.1.4.1.46609.1.1.1
-ROLE_SAS=1.3.6.1.4.1.46609.1.1.2
-ROLE_OPERATOR=1.3.6.1.4.1.46609.1.1.3
-ROLE_CBSD=1.3.6.1.4.1.46609.1.1.4
-ROLE_PAL=1.3.6.1.4.1.46609.1.1.5
-ROLE_INSTALLER=1.3.6.1.4.1.46609.1.1.6
+ROLE_SAS=1.3.6.1.4.1.46609.1.1.1
```
gregbillock: I checked all these and they look good, thanks.

gregbillock:
Also can you run the script and include the files in the PR? I think the existing examples have the old OID values in them.

cert/CACreationScript.sh
kharekuldeep: These variants are added to support End Entity Certificate generation for CBSD device using pre-defined RSA key or CSR. If RSA key is preconfigured in CBSD device and device can not use any new key,hence this variant shall support those cases.
While running the script user can provide RSA key file or CSR for CBSD End Entity certificate generation.

cert/CACreationScript.sh
kharekuldeep: Removed in latest commit.

cert/CACreationScript.sh
kharekuldeep: If user has provide CBSD device end entity key already while running bash script, this conditional statement is added to avoid RSA key generation for CBSD device.
So script shall only generate CSR for the case.

kharekuldeep:
@gregbillock I have committed the changes as per your comments.Also tried to answered your queries.Let me know if more information required.

cert/CACreationScript.sh
gregbillock: Ah, I see. Maybe add a note about that use case.

kharekuldeep:
@gregbillock Updated the script with the usage comment for key/CSR.

gregbillock:
Thanks! Feel free to merge.

gregbillock:
You can hit the rebase-and-merge button (or squash-and-merge; not sure what the practice is). That should give you options to close, delete the branch, and such.

PR #273, opened 2017-08-18
assaneRed_GRA_4

assaneRed:
adding testcases WINFF_FT_S_GRA_4


kate-harrison:
I believe this has been deprecated by WG4, but maybe it's worth keeping the PR up because the test scenario may have been merged into another test case.

PR #276, opened 2017-08-24
Added Heartbeat Assertion Helper Function

sushantparkhi:


src/harness/sas_interface.py
```diff
@@ -476,3 +476,34 @@ def assertRegisteredAndGranted(self, registration_request, grant_request,
...
+      cbsd_ids list.
+      operation_states: A list containing operation states (AUTHORIZED or 
+      GRANTED) for each devices in the cbsd_ids list.
+    Returns:
+      A list of transmit expire time in the format YYYY-MM-DDThh:mm:ssZ.    
+    """
+    pass
+
+  @abc.abstractmethod
+  def TriggerDailyActivitiesImmediatelyAndWaitUntilComplete(self):
```
kate-harrison: Any idea why this is showing up in the diff?

src/harness/sas_testcase.py
```diff
@@ -96,13 +96,42 @@ def assertRegisteredAndGranted(self, registration_request, grant_request,
...
-          'cbsdId': grant_resp['cbsdId'],
-          'grantId': grant_resp['grantId']
+        'cbsdId': grant_resp['cbsdId'],
+        'grantId': grant_resp['grantId']
       })
 
     # Return cbsd_ids and grant_ids
     return list_of_cbsd_ids_and_grant_ids
 
+  def assertHeartbeatsSuccessful(self, cbsd_ids, grant_ids, operation_states):
```
kate-harrison: It probably makes sense to have the output format of assertRegisteredAndGranted() match the input format of assertHeartbeatsSuccessful(). I'm not sure which should change though. Thoughts? (My gut reaction is that assertRegisteredAndGranted() should change, since it's the oddball right now.)

src/harness/sas_interface.py
sushantparkhi: I added this method, TriggerDailyActivitiesImmediatelyAndWaitUntilComplete in the same PR. May be I will create another PR for this because this was an issue where I forgot to add this method in the sas_interface.py as we agreed to shift this helper method to sas_testcase.py.

src/harness/sas_testcase.py
sushantparkhi: Yes, I think we can change the output format of the assertRegisteredAndGranted() to a tuple containing cbsd_ids and grant_ids?

kate-harrison: Sounds good to me. Then we can use the output of that directly with your new function.

src/harness/sas_interface.py
kate-harrison: Oh, got it -- it's essentially an unrelated bugfix that crept into this PR :) I simply hadn't noticed the bug (oops).

src/harness/sas_testcase.py
sushantparkhi: If I am correct then I have to change the input of this function as well? Instead of taking cbsd_ids and grant_ids separately, do you want me to take this as a tuple as well?

kate-harrison: Sorry, which of the two options are you suggesting?
Option 1: return ( [cbsd1, cbsd2], [grant1, grant2] )
Option 2: return ( (cbsd1, grant1), (cbsd2, grant2) )

I think both are fine, I just prefer consistency across all functions so that they can be easily strung together. I think both can easily be used by the code that calls these functions, although option 1 is slightly closer to what we've been doing (but not necessarily better otherwise).

If you prefer option 2, consider namedtuple.

sushantparkhi: I think option 1 is fine. But with option 1 the input will be `assertHeartbeatsSuccessful(self, tuple_of_cbsd_ids_grant_ids, operation_states):
  cbsd_ids, grant_ids = tuple_of_cbsd_ids_grant_ids` ?

kate-harrison: I think the tuple will be unpacked in the test case where the helper is called.

sushantparkhi: Then I think the current implementation should work!

kate-harrison: The implementation of assertHeartbeatsSuccessful() matches this decision but the implementation of assertRegisteredAndGranted() does not. It returns a list of dicts, each with two keys: 'cbsdId' and 'grantId'.

sushantparkhi: If you want I can make that change in the same PR? :)

kate-harrison: I won't object :)

src/harness/sas_testcase.py
```diff
@@ -94,14 +94,10 @@ def assertRegisteredAndGranted(self, registration_request, grant_request,
...
       self.assertTrue(grant_resp['grantId'])
       self.assertEqual(grant_resp['response']['responseCode'], 0)
-
-      list_of_cbsd_ids_and_grant_ids.append({
-        'cbsdId': grant_resp['cbsdId'],
-        'grantId': grant_resp['grantId']
-      })
+      grant_ids.append(grant_resp['grantId'])
 
     # Return cbsd_ids and grant_ids
```
kate-harrison: Would you mind adding a quick "assertEqual(len(cbsd_ids), len(grant_ids))" just to be absolutely sure there's no discrepancy?

src/harness/sas_interface.py
kate-harrison: I just noticed that this function is defined under the class SasTestcaseInterface rather than the class SasAdminInterface. Can you please move it?

kate-harrison: Nevermind, I'm talking nonsense. Please ignore my previous comment :)

PR #277, opened 2017-08-25
Reference implementation for the Winnforum propagation models (ITM and Hybrid)

sbdt-google:
This relies on backend propagation ITS reference implementation (in
C++) and some python extension module wrappers. Minor modifications to
the ITS code have been done for supporting WinnForum extensions and for
performance.
Top level access are the 2 files propagation wf_itm.py and wf_hybrid.py.

Optimized Geo modules are provided for accessing geo data and computing
great circles. In particular common Vincenty operation has been vectorized, as well
 as terrain profile extraction.

See README files for details.


redtechnologies:
Review performed (RED Technologies)

(Comment 1)
ITM (R2-SGN-03)

"For Initial Certification, SAS shall use the NTIA ITS Irregular Terrain Model (ITM) , in point-to-point mode, for propagation determination for use in FSS earth station, DPA protection, and ESC sensor protection. ..." 

There is no Python code to cover distance below 1km where ITM is not valid (itm.cpp) ... recommendation is to apply the HYBRID model for distance below 1km.

(Comment 2)
HYBRID (R2-SGN-04)

For pathloss calculation with distance > 80km the proposed reference model implementation is the following:

# TODO: validate the following approach with WinnForum participants:
# Note: For best performance, and avoiding another profile extraction,
#       we get the 80km from the original profile instead of extraction
#       from scratch which would require the following slow operations:
#
# lat_80km, lon_80km, _ = vincenty.GeodesicPoint(lat_cbsd, lon_cbsd,
#                                                80., bearing)
# its_elev_80km = terrainDriver.TerrainProfile(lat_cbsd, lon_cbsd, lat_80km, 
#							  lon_80km, 30, 1501)

RED Technologies implementation differs from the proposed reference model as we recompute terrain profile for distance = 80km; 

Recommendation: only one method should remain

(Comment 3)
HYBRID (R2-SGN-04)

For distance 0.1 < d < 1.0, the MedianBasicPropLos() should be computed using a fixed distance of 1km as per R2-SGN-04; proposed implementation makes distance (dist_km) varying between  0.1 < d < 1.0.

Recommendation: set dist_km to 1km for istance 0.1 < d < 1.0

(Comment 4)
R2-SGN-17 

The surface refractivity value varies by location and shall be derived by the methods and associated data files in ITU-R Recommendation P.452. The refractivity values shall be evaluated at the mid-point of the great circle path between transmitter and receiver. The climate value varies by location and shall be derived by the methods and associated data files in ITU-R Recommendation P.617.

The proposed implementation of CalcItmPropagationLoss() gives an extra possibility to overwrite the surface refractivity value as per the requirement R2-SGN-17 by passing directly another value through the API 

Recommendation: remove this possibility to overwrite surface refractivity value

(comment 5)
R2-SGN-22

The extended Hata mean path loss shall be computed starting with the extended Hata median path loss in dB (L=LEH or L=LEHB as defined in R2-SGN-04).  Then defining ξ=10/ln10, the mean extended Hata path loss (dB) is L+σ^2/2ξ   where σ is the land use category (urban or suburban) dependent log-normal distribution using a standard deviation given by equations A-18(a,b,c) in [TR 15-517 ref] and the land category is chosen using R2-SGN-04.

The proposed implemention of GetEHataStandardDeviation() do not implement the standard deviation as per R2-SGN-22, instead a correction factor is applied; this is a source of possible divergence with compliant implementation as RED's implementation.

Recommendation: implement CDF as per R2-SGN-22

(Comment 6)
EHATA cpp implementation

The actaul size of pfl_segment[] should not be set to 400 (hard coded) but be set to "10000/resolution" or as RED's implementation use dynamic allocation. The current proposed code crashes if resolution below 30 meter.

(Comment 7)
EHATA cpp implementation, compilation warning as follows

EXTENDED_HATA.cpp(462): warning C4305: '=' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(463): warning C4305: '=' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(808): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(810): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(811): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(841): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(842): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(843): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(887): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(889): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(891): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(893): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(967): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(968): warning C4305: 'initializing' : truncation from 'double' to 'float'
EXTENDED_HATA.cpp(969): warning C4305: 'initializing' : truncation from 'double' to 'float'

(Comment 8)
EHATA cpp implementation, it is recomended to initialize interValues data struct within the cpp; don't rely on Python data initialisation.

(Conclusion)
Beside the above comments, Google's and RED's implementation gave same path loss results.

kate-harrison:
(Re-opening, as I suspect the closure was unintentional.)

sbdt-google:
You wrote:
>> ITM (R2-SGN-03)
>> "For Initial Certification, SAS shall use the NTIA ITS Irregular Terrain Model (ITM) , in point-to-point 
>> mode, for propagation determination for use in FSS earth station, DPA protection, and ESC sensor 
>> protection. ..."
>> There is no Python code to cover distance below 1km where ITM is not valid (itm.cpp) ... 
>> recommendation is to apply the HYBRID model for distance below 1km.

Can you please give more information about this requirement of not using ITM below 1km ? I have not seen this in the spec, but I will check with Andy Clegg who has been following the spec discussion on these topics from our side

sbdt-google:
Regarding Comment 2:
We feel that extracting again the terrain with exact distance 80km is time consuming and does not bring much benefit compared to the proposal of:
 1) using the original terrain profile up to the receive point
 2) selecting the closest point to 80km, and using that sub-profile for the calculation involving the delta between ehata and itm.
In fact this last approach should better remove discontinuity at 80km distance, which is the purpose of this correction.

sbdt-google:
Regarding Comment 3:
  Thanks for finding this - the issue has been fixed and submitted

Regarding Comment 4:
  This was for flexibility and unit testing. But I agree, and changed the functions appropriately (and the test routines to fit the new function signatures.

sbdt-google:
Regarding comment 5:

Please have a look below. The median to mean has been modified slightly and an experimental test has been added to validate the formula vs a random lognormal sample.

sbdt-google:
Regarding comment 6:
This is a known issue of the ITS EHata implementation: it does not support resolution below 25m (and not 30m), because of the fixed array. Winnforum use of this code is always for resolution 30m or higher, so it is not affected by the issue. 
The idea was to keep the ITS implementation mostly unchanged, except for the things which affect Winnforum (see the README). And to fix other issues such as this one, only when original ITS implementation has been modified in the ITS Github repository.

Regarding comment 7:
Please explain what you mean. There is no file EXTENDED_HATA.cpp in this PR.
Perhaps you mean that your internal implementation returns these warnings, and not the PR implementation? Please note of one modification in ehata.h to use double instead of float (see also the README), otherwise:
 - the profile is inconsistent with the ITM model which uses double
 - it creates unstable code which returns different values depending on compiler used

Regarding comment 8:
Please explain what you mean. The python does not initialize the interValues, which is an internal structure of the cpp code.

src/harness/reference_models/geo/refractivity.py
```diff
@@ -0,0 +1,105 @@
...
+  Typical usage:
+    refractor = RefractivityIndexer()
+    r = refractor.Refractivity(19.66, -155.55)
+  """
+  def __init__(self, datafile_or_dir=None):
+    self.ConfigureDataFile(datafile_or_dir, do_load=False)
+    self._num_lat_rows = 121  # Number of latitude rows in the data file
+    self._num_lon_cols = 241  # Number of longitude columns in the data file
+    self._lat_start = 90.0  # Latitude corresponding to first row of file (deg)
+    self._lon_start = 0.0  # Longitude corresponding to first column of file (deg)
```
jcereced: comment from Commscope:

This should be self._lon_start = -180.0 # Longitude corresponding to first column of file (deg). This longitude in the file starts from -180

src/harness/reference_models/geo/refractivity.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Returns:
+      the sea level refractivity on that point.
+    """
+    if self._data is None:
+      self._data = np.loadtxt(self._datafile)
+      logging.info('Loaded refractivity data from %s' % self._datafile)
+
+    if lon < 0:
+      lon = lon + 360.0
```
jcereced: comment from Commscope

The lines (79-80) are not needed. Longitude falls between -180 and 180 for the data file (Spectrum-Access-System/src/harness/reference_models/geo/testdata/itu/n050.txt ).

src/harness/reference_models/geo/refractivity.py
sbdt-google: I am sorry, but the spec P452.16 found at https://www.itu.int/rec/R-REC-P.452-16-201507-I/en, clearly specifies that the longitude starts at 0 degree:
"
The maps are contained in the files DN50.TXT and N050.TXT, respectively. The data are from 0d to 360d in longitude and from +90d to –90d in latitude, with a resolution of 1.5d in both latitude and longitude. The data are used in conjunction with the companion data files LAT.TXT and LON.TXT containing respectively the latitudes and longitudes of the corresponding entries (gridpoints) in the files DN50.TXT and N050.TXT
"

Looking at the LON.TXT file, we also see that it starts at 0 degrees.
Can you confirm this? It is the climate file which starts near -180degree.

src/harness/reference_models/geo/refractivity.py
sbdt-google: see previous comment

jcereced: comment also withdrawn, please see above

src/harness/reference_models/propagation/ehata/ehata.py
```diff
@@ -0,0 +1,110 @@
...
+  dist_km = np * xi          # path distance, in km
+  elev_cbsd = its_elev[2]
+
+  if height_cbsd < 20:
+    height_cbsd = 20.
+
+  if dist_km < 3.0:
+    eff_height = height_cbsd
+
+  else: # dist_km >= 3 km
```
jcereced: else: # dist_km >= 3 km
    i_start = 2 + int(3.0 / xi)
    i_end = np + 2
    if dist_km > 15:
      i_end = 2 + int(15.0 / xi)
      dist_km = 15.0
  
  avg_height = sum(its_elev[i_start:i_end+1]) / float(i_end - i_start + 1)
 
This will always pick the elevation profile at a point before the 3.0 km mark and use it in the averaging process.
It should be i_start = 3 + int(3.0 / xi)
 

WINNF-17-I-0095r9 Prop Model Edits said
 
For path lengths greater than 3 km but less than 15 km, hb is the structure height, hcbsd, plus (d-3)/12 times the difference between the terrain elevation at the CBSD antenna and the average terrain elevation over a distance of 3 km to the location of the receiver.

src/harness/reference_models/geo/refractivity.py
sbdt-google: No problem.

src/harness/reference_models/propagation/ehata/ehata.py
sbdt-google: Right: initially I actually had the following which meet your comment as well:
```
  i_start = 2 + ceil(3.0 / xi)
```
However the internal C++ eHata implementation computes the effective height with the 'int' method - See PreprocessTerrainPath.cpp:
```
    else if (3.0 <= d__km && d__km <= 15.0)
    {
        i_start = 2 + int(3.0 / xi);
        i_end = np + 2;
         for (int i = i_start; i <= i_end; i++)
            sum = sum + pfl[i];
```
So in order to be consistent with eHata ITS implementation, I decided to use the same logic using the point exactly a 3km or the one just before. I dont think this will make a big numerical difference at the end, but we should definitely choose one or the other:
  - use same method as in ehata
  - follow exactly the winnforum spec (which itself tried to describe the ehata implementation, I guess, but with slightly wrong description).
What do you think ?

jcereced: It does make a small difference. We have seen some corner cases. So as you rightly pointed out to be consistent we should follow the winnforum requirement.
We should use i_start = 2 + ceil(3.0 / xi) and also modify PreprocessTerrainPath.cpp accordingly.

sbdt-google: Please check the latest changes made: 
modified to use path profile strictly includuded in the 3km-15km range,
both in Python and eHata implementation

jcereced: changes look good.  thanks!

redtechnologies:
Regarding Comment 2

Here is below an exemple where both implementations provide significant different results; again we should all agree on one single method either method 1 or 2 

Method 1: Google (To compute "J" reuse profil up to 80 km)
Method 2: recommended by RED (To compute "J" recompute profil for 80km)

lat1, lng1, height1 = 37.751985, -122.443890, 20.0
lat2, lng2, height2 = 37.094745, -122.040671, 10.0  # 81km away

res = wf_hybrid.CalcHybridPropagationLoss(lat1, lng1, height1, lat2, lng2, height2,
                                reliability=0.5, freq_mhz=3625.,
                                region='SUBURBAN', climate=5, refractivity=314)

Result method 1

Loss ITM median :  269.067
Ehata_loss_80km : 242.864608853
Itm_loss_80km : 244.058768926
J = max(ehata_loss_80km - itm_loss_80km, 0) : 0
Loss final = Loss ITM median + J : 269.067

Result method 2

Loss ITM median :  269.067
Ehata_loss_80km : 242.365
Itm_loss_80km : 229.078
J = max(ehata_loss_80km - itm_loss_80km, 0) : 13,287
Loss final = Loss ITM median + J : 282.354

redtechnologies:
Regarding comment 5:

Fine but this implementation still diverges from R2-SGN-22; All SAS providers should agree on this implementation; furthermore the formula will not avoid SAS providers to use random lognormal sampling for the interference aggregation part as per R2-SGN-12.

redtechnologies:
Comment 9 

Retrieves the elevation of a given lat/lon.

_GetTerrainElevationPoint    (terrain.py)

Replace 

  xm = math.floor(float_x) and ym = math.floor(float_y)

by

  xm = np.floor(float_x).astype(int) and ym = np.floor(float_y).astype(int)

if not tile_cache[] is worngly indexed with float not integer and crashed.

redtechnologies:
Comment 10

In point_to_point (itm.cpp) a new parameter has been added : bool eno_is_final with the following purpose

   eno = eno_ns_surfref;

  //*** WinnForum modification
  //original ITM test code is using a final value of 314 without altitude compensation
  //enso = 0.0;
  //q = enso;

  q = eno;

  if(!eno_is_final)
  {
    if (q <=0) q = 310;
    ja = 3.0 + 0.1 * elev[0];
    jb = np - ja + 6;
    for(i=ja-1;i<jb;++i)
      zsys+=elev[i];
    zsys/=(jb-ja+1);
  }

We understand in the context of the SAS/WINNF the boolean must never be set to "true".

To avoid a misuse of the API we recommend to remove the bool eno_is_final and apply altitude compensation in all cases

sbdt-google:
Regarding Comment 2

After more thinking, I agree with your method. Please see my changes accordingly. 

My change description is the following:

Initial implementation proposed to reuse a subset of the full
profile to avoid a new profile extraction exactly at 80km.
However doing this may create discontinuous issues, as the J
compensation factor (which is supposed to insure continuity from
below 80km to above 80km) may change drastically due to:
 - the profile sampling points will change from one  point to an other
 given that we only keep 1500pts in the profile and make the resolution
 vary.
 - so the J factor may be varying value for any points above 80km, especially
 given the high non linearity of the ITM models. The test_over_80km()
 for example demonstrate such behavior. The J calculated at 80km is 0.
 The J calculate using profile at 81km cut to just below 80km (ie we are looking
 at the profile up to point 79km 974m instead of exactly 80km) shows J=13dB.

sbdt-google:
Regarding Comment 5:

I am sorry but can you specify exactly how it diverges from R2-SGN-22, as I do not understand what you mean? Please make your review directly on the code with the GitHub reviewing tool, and proposing a new code implementation. Thanks

PR #279, opened 2017-09-08
assaneRed_GRA_13

assaneRed:
adding testcases WINNF_FT_S_GRA_13 associate with certificates


src/harness/testcases/grant_testcase.py
```diff
@@ -253,6 +253,56 @@ def test_WINNF_FT_S_GRA_12(self):
...
+    The responseCode should be (103, 105), with NO channels in result.
+    """
+    device_a_cert = os.path.join('certs', 'device_a.cert')
+    device_a_key = os.path.join('certs', 'device_a.key')
+
+    device_c_cert = os.path.join('certs', 'device_c.cert')
+    device_c_key = os.path.join('certs', 'device_c.key')
+
+    # Checks that devices certificate are valid (checking all chain).
+    self.assertValidClientCertificate(device_a_cert, sas.CA_CERT)
```
kate-harrison: I think we can skip this check, as it's not asked for in the test spec. That will also let you revert the changes which redefine assertValidClientCertificate().

PR #283, opened 2017-09-14
Antenna reference code.

sbdt-google:
Allows to compute antenna gain for CBSD, DPA (radar), ESC and FSS.
Also provides a routine to find the FSS possible pointing angles according to the licensed information (range of east/west elevation, azimuth and GSO arc longitude).

This is an extension of RedTechnologies  PullRequests of July with better integration in the reference_models/ framework developed since then.

### DIFF
--- a/src/harness/reference_models/antenna/antenna.py
+++ b/src/harness/reference_models/antenna/antenna.py
@@ -75,7 +75,7 @@ def GetAntennaPatternGains(hor_dirs, ant_azimuth,
   return gains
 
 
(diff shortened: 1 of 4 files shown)


redtechnologies:
Unable to review this code since it is unclear whether this code is an extension, an alternative, a replacement or a reuse of previous proposed CBSD and FSS antenna gain reference models RED Technologies PullRequests from July 2017.Note those models was reviewed and commented already. We need ONE single reference in Git before to undergo final  review.

sbdt-google:
Thanks to Red and Javier from Commscope who approved it.

jcereced:
Commscope approves as well

PR #285, opened 2017-09-15
DPA move list reference code

Thao-Nguyen:


src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,435 @@
...
+
+    # Get indices, bearings, interference contributions of the sorted grants
+    sorted_grant_index = [i.get('grantIndex') for i in interf_list]
+    sorted_bearing = [i.get('bearing_c_cbsd') for i in interf_list]
+    I = np.array([i.get('randomInterference') for i in interf_list]).transpose()
+
+    return I, sorted_grant_index, sorted_bearing
+
+
+def grantsInMainBeam(ant_az, bearing):
```
sbdt-google: Function useless if using the `antenna.GetRadarNormalizedAntenna()` routine.
See a comment below.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,435 @@
...
+
+    # Loop through every azimuth angle.
+    for a in azimuths:
+        # Calculate interference contributions at output of receiver antenna.
+        G = np.full((1,Nc),G_discrim)
+
+        idx = grantsInMainBeam(a,bearing)
+        if idx.size:
+            G[:,idx] = 1.0
+        IG = I_mW*G
```
sbdt-google: All the previous can be replaced using function `antenna.GetRadarNormalizedAntenna()` which returns directly the gains in dB:
```
for azi in azimuths:
  dpa_gains = antenna.GetRadarNormalizedAntennaGains(bearings, azi)
  IG = I_mW  *  10**(dpa_gains/10.)
```

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,435 @@
...
+        # Compute the 95th percentile of the aggregate interference, and remove
+        # grants until the protection threshold is met or all grants are moved.
+        A95 = np.percentile(IG.sum(1), PROTECTION_PERCENTILE, interpolation='lower')
+        while (A95 > t_mW):
+            I_mW[:, nc - 1] = 0.0  # zero-out the nc-th column of I_mW
+            IG[:, nc - 1] = 0.0    # zero-out the nc-th column of IG
+            nc = nc - 1
+            if nc == 0:
+                return 0
+            A95 = np.percentile(IG.sum(1), PROTECTION_PERCENTILE, interpolation='lower')
```
sbdt-google: no need to zero out if you use np.sum, and take a view:
```
A95 = np.percentile(np.sum(IG[:,0:nc+1], axis=1
```

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+def findMoveList(protection_points, low_freq, high_freq, threshold, num_iter,
+                 registration_request, grant_request, exclusion_zone):
+
+    """
+    Main routine to find CBSD indices on the move-list.
+
+    Inputs:
+        protection_points:  a list of protection points
```
sbdt-google: For the top level function, can we describe better the format of inputs and outputs.
For example the protection points currently assume being a list of dict of 'latitude' and 'longitude', etc..

sbdt-google:
Hi Thao,

I created a PR for the HAAT calculation, based on the code I sent you some time ago. 
See #318.

Thao-Nguyen:
Thanks so much, Serge! I will review and incorporate it into the move_list.py.

Thao-Nguyen:
Hi Serge, we have incorporated the wf_itm.ComputeHaat() into our move_list.py. Thanks so much!

sbdt-google:
Your welcome.. 
By the way, I am continuing some tests on your DPA move list implementation. I hope I will have finished sometime soon, so we can then proceed with last minor modification (if needed) before submit.

Thao-Nguyen:
FYI, we are making changes to the DPA move list implementation as well. But let's know if you find any problems with the current code. Thanks!

src/harness/reference_models/move_list_calc/move_list_example.py
```diff
@@ -0,0 +1,109 @@
...
+from pykml import parser
+import move_list
+import time
+from collections import namedtuple
+
+if __name__ == '__main__':
+
+    # Inputs
+    low_freq = 3600000000 	# low frequency of protection constraint (Hz)
+    high_freq = 3610000000  # high frequency of protection constraint (Hz)
```
sbdt-google: See some other comment. I dont think this will work with the rest of the code, as it supports only 1MHz bandwdith otherwise it introduces an offset.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+                                            reliability=reliabilities,
+                                            freq_mhz=FREQ_PROP_MODEL)
+    path_loss = np.array(results.db_loss)
+
+    # Compute CBSD EIRP inside the frequency range of protection constraint
+    eirpPerMHz_cbsd = antenna.GetStandardAntennaGains(results.incidence_angles.hor_cbsd,
+                                                      ant_az_cbsd, ant_beamwidth_cbsd,
+                                                      maxEirpPerMHz_cbsd)
+    eff_bandwidth = min(high_freq_cbsd, high_freq_c) - max(low_freq_cbsd, low_freq_c)
+    eirp_cbsd = eirpPerMHz_cbsd + 10 * np.log10(eff_bandwidth / 1000000.0)
```
sbdt-google: I am not sure we actually capture the EIRP per MHz here.

The constraint bandwidth is typically 10MHz (see your script for example). So eff_bandwidth is 10MHz. So we are taking eirpPerMHz_cbsd and adding 10dB. So we are effectively computing the EIRP per 10MHz here.
The threshold of -144dBm for DPA is to be compared to EIRP per Mhz, so in your script we have a 10dB off.
All this works only if the caller is passing a constraint freq range of exactly 1Mhz.

All the EIRP thing is a bit confusing... I think that at the end we need to compute a separate move list per SAS channel (ie 5MHz or 10MHz depending how latest discusionn goes in Winnforum), and that we simply need to check:
 1) do we intersect with the radar detected frequency (what you call the constraint frequency) - you do it properly in the findGrantsInsideNeighborhood() routine
 2) in that case, simply take the EIRP per MHz , since there is no fractional intersection of the radar bandwdidth

In conclusion the two linesshould be removed

Please think about this and let me know if I get this wrong !

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+
+    # Calculate the interference contributions
+    interf = eirp_cbsd - path_loss
+    median_interf = interf[-1]      # last element is the median interference
+    K_interf = interf[:-1]          # first 'K' interference
+
+    # Store interference contributions
+    interference = {
```
sbdt-google: Again I recommend namedtuple rather than dict for simple records

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+    """
+
+    # Compute interference contributions of each grant to the protection constraint
+    interf_list = []
+    for cbsd_grant in cbsd_grant_list:
+        interf = computeInterference(cbsd_grant, constraint, num_iter)
+        interf_list.append(interf)
+
+    # Sort grants by their median interference contribution, smallest to largest
+    interf_list.sort(key=itemgetter('medianInterference'))
```
sbdt-google: In case you decide to chang interf to namedtuple (see previous comment), then you can get easily the sorting with:
```
 from operator import attrgetter
 interf_list.sort(key=attrgetter('medianInterference'))
 sorted_grant_index = [interf.grantIndex for interf in interf_list]
...
```

src/harness/reference_models/move_list_calc/move_list.py
sbdt-google: Update: I was wrong : the -144 is to be compared with the EIRP/10Mhz, I got confused by internal mismatched documentation. 
So I guess this change my previous remark, although I still think it is still valid in the principle of simply computing the power as eirpPerMhz +10dB, rather than eff_bandwidth ?? 
Please comment

Thao-Nguyen: The constraint bandwidth is typically 10 MHz (e.g., 3600-3610 MHz in the move_list_example.py). However, the overlapping/intersecting bandwidth (denoted as 'eff_bandwidth') between the CBSD grant and the protection constraint frequency range <= 10 MHz. For example, if the CBSD grant has a frequency range of 3605-3610 MHz, eff_bandwidth = 5 MHz and 10*np.log10(eff_bandwidth / 1000000.0) = ~7 dB (not 10 dB). That's why, we need to compute 'eff_bandwidth'. By the way, 'eff_bandwidth' might not be a good name, I will change it to 'overlapping_bw' as in the findGrantsInsideNeighborhood(), what do you think? 

Yes, the protection threshold for co-channel offshore DPAs is -144 dBm/10MHz. More confusing is that the threshold for out-of-band inland DPAs is -125 dBm/MHz (not dBm/10MHz) (in R2-IPM-04). 

Please let me know if it's still not clear to you.

Thao-Nguyen:
Thanks so much, Serge, for reviewing our code carefully and for providing helpful comments. To your 2 most important comments:
- EIRP related issue: no change for now.
- Reliability values: agree with you to limit the range to 0.001-0.999.

We will update the code and commit the latest version as soon as we can. Thanks, again!

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+    """
+    Identify the Nc CBSD grants in the neighborhood of protection constraint.
+
+    Inputs:
+        reg_request:    a list of CBSD registration requests
+        grant_request:  a list of grant requests
+        grant_index:    an array of grant indices
+        constraint:     protection constraint, a tuple with named fields of
+                        'latitude', 'longitude', 'low_frequency', 'high_frequency'
+        exclusion_zone: a list of locations used to form the exclusion zone
```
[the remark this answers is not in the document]
Thao-Nguyen: Good point about passing either an object shaply.geometry.Polygon or lat_lng as an input. I will modify the code and the description accordingly.

Passing a MultiPolygon including only East-Gulf and West coastal exclusion zones to the move_list.py would be ok for co-channel offshore DPAs protection. But I'm afraid that it will get complicated when we incorporate co-channel inland DPAs protection. I would like to use a Polygon for now, but I will keep MultiPolygon option in mind.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+# This version of the code only addresses 'co-channel' frequency range protection
+# in R2-SGN-24.
+
+# The main routine is 'move_list.findMoveList()'
+#==================================================================================
+
+import numpy as np
+from operator import itemgetter
+from shapely.geometry import Point as SPoint
+from shapely.geometry import Polygon as SPolygon
```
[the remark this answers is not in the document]
Thao-Nguyen: Please see below.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+
+    Returns:
+        grants_inside:  a list of grants, each one being a dictionary containing
+                        installation and grant information of all
+                        CBSDs inside the neighborhood of the protection constraint.
+    """
+
+    grants_inside = []
+    polygon_ex_zone = SPolygon(exclusion_zone)  # create polygon of the exclusion zone
+
```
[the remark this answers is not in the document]
Thao-Nguyen: Please see the response above.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+                                                               lat_c, lon_c)
+        flag_geo = False
+        if cbsd_cat == 'A':
+            point_cbsd = SPoint(lat_cbsd, lon_cbsd)  # create point
+            if dist_km <= CAT_A_NBRHD_DIST and polygon_ex_zone.contains(point_cbsd):
+                flag_geo = True
+        elif cbsd_cat == 'B':
+            if dist_km <= CAT_B_NBRHD_DIST:
+                flag_geo = True
+            elif (dist_km <= CAT_B_NBRHD_DIST_FOR_HT) and \
```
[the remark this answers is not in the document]
Thao-Nguyen: Nice catch! The new version of the code will be slightly different with the above, but I will ensure there is no useless parenthesis or \ line continuation.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+            get('operationFrequencyRange', {}).get('lowFrequency')
+        high_freq_cbsd = grant_request[i].get('operationParam', {}).\
+            get('operationFrequencyRange', {}).get('highFrequency')
+        low_freq_c = constraint.low_frequency
+        high_freq_c = constraint.high_frequency
+        overlapping_bw = min(high_freq_cbsd, high_freq_c) - max(low_freq_cbsd, low_freq_c)
+        flag_freq = (overlapping_bw > 0)
+
+        # Return CBSD information if it is inside the neighborhood of protection constraint
+        if (flag_geo == True) and (flag_freq == True):
```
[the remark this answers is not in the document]
Thao-Nguyen: You are right. But I might need to change 'freq_intersect' to 'freq_check' (or something like that). The reason is that for out-of-band inland DPAs (R2-SGN-24.b.), we need to consider all Cat B CBSDs within 25 km of the DPA protection point p, regardless of whether their frequency ranges overlap/intersect with the protection frequency range or not.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+            cbsd_grant = {
```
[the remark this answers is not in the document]
Thao-Nguyen: Yep, we have used a namedtuple 'CBSDGrant' in the latest version of the code. Thanks!

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+    # Get protection constraint information (protection point and frequency range)
+    lat_c = constraint.latitude
+    lon_c = constraint.longitude
+    low_freq_c = constraint.low_frequency
+    high_freq_c = constraint.high_frequency
+    h_c = REF_INC_ANTENNA_HT
+
+    # Compute median and K random realizations of path loss/interference contribution
+    # based on ITM model as defined in [R2-SGN-03] (in dB)
+    reliabilities = np.random.rand(num_iteration)    # get K random reliability values from an
```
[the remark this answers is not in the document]
Thao-Nguyen: Totally agree.

src/harness/reference_models/move_list_calc/move_list.py
Thao-Nguyen: We will have a namedtuple 'InterferenceContribution' in the updated script.

src/harness/reference_models/move_list_calc/move_list.py
Thao-Nguyen: Got it. Thanks!

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,392 @@
...
+    # Index the grant requests in the same order in the registration_request and
+    # grant_request lists [1, 2,..., N].
+    grant_index = range(1, len(registration_request) + 1)
+
+    # Loop over each protection point and find the move list
+    M_temp = []
+
+    # Define protection constraint, i.e., a tuple with named fields of
+    # 'latitude', 'longitude', 'low_frequency', 'high_frequency'
+    ProtectionConstraint = namedtuple('ProtectionConstraint', ['latitude', 'longitude',
```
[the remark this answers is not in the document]
Thao-Nguyen: Oh, good to know. Thanks!

src/harness/reference_models/move_list_calc/move_list_example.py
Thao-Nguyen: I think we are clear on this.

src/harness/reference_models/move_list_calc/move_list_example.py
```diff
@@ -0,0 +1,109 @@
...
+    # Get east-gulf coastal exclusion zones (enclosed coastal exclusion zones with U.S. border)
+    filename = os.path.join(_BASE_DATA_DIR, 'protection_zones.kml')
+    with open(filename, 'r') as kml_file:
+        coastalZoneDoc = parser.parse(kml_file).getroot()
+    placemarks = list(coastalZoneDoc.Document.Placemark)
+    exclusion_zone = []
+    for pm in placemarks:
+        name = pm.name.text
+        if (name == 'East-Gulf Combined Contour'):
+            line = pm.MultiGeometry.Polygon.outerBoundaryIs.LinearRing.coordinates.text
```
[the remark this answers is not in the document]
Thao-Nguyen: Agree.

src/harness/reference_models/move_list_calc/move_list_example.py
```diff
@@ -0,0 +1,109 @@
...
+            coords = line.split(' ')
+            lat_long = []
+            for c in coords:
+                if c.strip():
+                    xy = c.strip().split(',')
+                    lat_long.append([float(xy[1]), float(xy[0])])
+                    exclusion_zone = lat_long
+					
+    # Populate protection points (a list of namedtuples with fields 'latitude' and 'longitude')
+    ProtectionPoints = namedtuple('ProtectionPoints', ['latitude', 'longitude'])
```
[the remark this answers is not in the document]
Thao-Nguyen: I will try this. Thanks!

redtechnologies:
Regarding the  range of relialabiility values, we propose to use a step of 0.01, similarly to the following comment we made  for "ITM Power aggregation" :

"Computing 1000 samples for each CBSD is costly from a performance point of view, since it requires to call "avar" procedure 1000 times.

Using a step of 0.01 (i.e., 100 samples to compute) as defined in R2-SGN-22 would seem preferable.
For the Monte Carlo computation, 1000 samples would then be randomly chosen from those 100 samples."

Thao-Nguyen:
Hi redtechnologies,

Using only 100 reliability values (actually 99 values of 0.01:0.01:0.99) for 1000 Monte Carlo iterations (R2-SGN-12) and for 2000 iterations (R2-IPM-03) does not seem to provide enough granularity of the path loss CDF. Are you going to ignore all the reliability values in [0.001, 0.01) and (0.99, 0.999)? We noticed that the difference between path losses (at reliability=0.001 and reliability=0.01), is quite large. The same behavior was observed for reliability=0.99 and reliability=0.999.  

Serge proposed a slightly different approach from yours, i.e., using 101 reliability values of 0:0.01:1 and then do interpolation to get the path losses for other reliability values. Please see below Michael's reply to his approach. 

##############################################################################
Hi Serge,

If you’re referring to interpolation of the path loss CDF, as opposed to calling the prop model with all reliability values, that saved only 6% of computation time in our tests.  More importantly, R2-SGN-12 in WINNF-TS-0112 is silent on interpolation; it’s only explicitly stated in R2-SGN-22 in relation to computing the mean path loss for IAP.  Given, the marginal savings and the current requirements, we’ve decided to draw from the full CDF, and the benchmark numbers I quoted below reflect that.

Michael
##############################################################################

sbdt-google:
To add more insights on this issue:
First I notice that results is very sensitive to low values. Thao is right that clipping at 0.01 would remove some of the extrema values (where you see the biggest change in values), and this makes quite a big difference when running the move list algo, in the end result that you get.
On the other hand the ITM model is clearly stated to be valid only only for values between 0.01 and 0.99 (see for example comments in the ITM source code), so I am not sure how valid it is to look at values below 0.01.
I recommended that at least we do not go below 0.001 (ie 0.1%) in the values that we use.

Regarding the implementation, my take is that the reference implementation can use the non-interpolation approach taken by NIST, and the actual SAS can implement this the way they want. 
Doing a simple linear interpolation of values 0 to 1.0 with 101 values does not work well due to the strong non linearity. I had good results by doing linear interpolation using 100 values between 0.01 and 0.99, plus adding 10 values on each end between 0.001 and 0.01. But best is to do first a non linear transform (of the log type) before doing the linear interpolation. In that case you can reduce further the number of avar() function that you need to call, by only computing values for equidistant values in this transformed space.
tail_itm

redtechnologies:
Thanks a lot for those insights.

In our opinion, it does not make sense to take into account values beyond those that are valid for the ITM model.

So if the model is only valid between 0.01 and 0.99, we should not compute values using reliability below 0.01 or above 0.99 (note that below, 0.01 the pathloss decreases a lot and therefore diminishes spectrum availability for CBSDs, while not necessarily reflecting a realistic physical propagation, since it seems outside of ITM allowed range).

msouryal:
The use of ITM reliabilities in the range 0.001 to 0.999 was based on a recommendation from NTIA/ITS (conference call with Paul McKenna, Nov. 3, 2017).

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,582 @@
...
+    # Compute EIRP of CBSD grant inside the frequency range of protection constraint
+    if dpa_type != 'out-of-band inland': # For co-channel offshore/inland DPAs
+
+        # CBSD co-channel bandwidth overlapping with
+        # frequency range of protection constraint
+        co_channel_bw = min(high_freq_cbsd, high_freq_c) - max(low_freq_cbsd, low_freq_c)
+
+        # Overlapping bandwidth between the first 10 MHz of CBSD grant adjacent-channel and
+        # the frequency range of protection constraint
+        adj_channel_bw = min(10000000.0, high_freq_c - low_freq_c - co_channel_bw)
```
sbdt-google: Can you please share with me the spec describing the adjacent channel inclusion? I coulkd not find it among the various documentation.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,582 @@
...
+            eirp_cbsd = 10 * np.log10(power_cbsd_mW) + ant_gain
+
+    else:   # For out-of-band inland DPAs
+        # Compute CBSD antenna gain in the direction of protection point
+        ant_gain = antenna.GetStandardAntennaGains(results.incidence_angles.hor_cbsd,
+                                                   ant_az_cbsd, ant_beamwidth_cbsd,
+                                                   ant_gain_cbsd)
+
+        # Compute out-of-band conducted power
+        oob_power = ComputeOOBConductedPower(low_freq_cbsd, low_freq_c, high_freq_c)
```
sbdt-google: Here also, the oob_power is missing the overall power scalor of the cbsd, I believe. The ComputeOOBConductedPower actually computes a relative scalor, not a final power.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,582 @@
...
+                grantIndex=grant_index[i],
+                maxEirp=grant_request[i].get('operationParam',{}).get('maxEirp'),
+                lowFrequency=low_freq_cbsd,
+                highFrequency=high_freq_cbsd)
+
+            grants_inside.append(cbsd_grant)
+
+    return grants_inside
+
+def ComputeOOBConductedPower(low_freq_cbsd, low_freq_c, high_freq_c):
```
sbdt-google: Can you share the spec document describing the detail of this calculation ?

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,582 @@
...
+                              + 10 * np.log10(overlap_bw_outside_10MHz / 1000000.0)
+        power_outside_10MHz_mW = np.power(10.0, power_outside_10MHz/10.0)
+    if overlap_bw_below_3530MHz > 0:
+        power_below_3530MHz = OOB_POWER_BELOW_3530MHZ \
+                              + 10 * np.log10(overlap_bw_below_3530MHz / 1000000.0)
+        power_below_3530MHz_mW = np.power(10.0, power_below_3530MHz/10.0)
+
+    power_mW = power_within_10MHz_mW + power_outside_10MHz_mW + power_below_3530MHz_mW
+
+    return 10 * np.log10(power_mW)
```
sbdt-google: Missing the overall scalor from the cbsd power (ie something like (max_cbsd_eirp - max_ant_gain)).

src/harness/reference_models/move_list_calc/move_list_example.py
sbdt-google: Ok forget this comment.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,582 @@
...
+        if len(Nc_grants):  # Found CBSDs in the neighborhood
+
+            # Form the matrix of interference contributions
+            I, sorted_grant_index, bearing = formInterferenceMatrix(Nc_grants, constraint,
+                                                                    h_inc_ant, num_iter, dpa_type)
+
+            # Find the index (nc) of the grant in the ordered list of grants such that
+            # the protection percentile of the interference from the first nc grants is below
+            # the threshold for all azimuths of the receiver antenna.
+            nc = find_nc(I, bearing, threshold)
```
sbdt-google: I am not sure the current find_nc is appropriate for out of band inland DPA.
According to the spec R2-IPM-4: 
> Unless otherwise noted by NTIA, inland out-of-band DPAs (below 3500 MHz) are defined as the 3 km exclusion zone around the sites listed in TR 15-517 (check if an existing Reference in document). For inland out-of-band DPAs, the protection level applied at DPA points shall be -125 dBm/MHz, and the reference antenna characteristic is an isotropic antenna 5 m above ground-level.  

So we shall not simulate the rotating antenna with 1.5degree, but simply add all signals of all neighboring CBSD with a single 0dBi antenna gain.

src/harness/reference_models/move_list_calc/move_list.py
sbdt-google: Ok I see that you use something similar to the FSS out of band purge list, where the power is the absolute power in dBm/MHz that the CBSD cannot exceed. (as defined in R0-DEV-4 and R0-DEV-5).
Where in a spec or doc, do we have this way of doing defined ?

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,582 @@
...
+            eirp_cbsd = eirpPerMHz_cbsd + 10 * np.log10(co_channel_bw / 1000000.0)
+
+        else:   # Consider both co-channel bandwidth and first 10 MHz adjacent-bandwidth
+            # Compute CBSD antenna gain in the direction of protection point
+            ant_gain = antenna.GetStandardAntennaGains(results.incidence_angles.hor_cbsd,
+                                                       ant_az_cbsd, ant_beamwidth_cbsd,
+                                                       ant_gain_cbsd)
+
+            # Compute conducted CBSD power within co-channel bandwidth
+            power_co_channel_bw = (maxEirpPerMHz_cbsd - ant_gain_cbsd) \
```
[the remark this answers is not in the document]
sbdt-google: For the adjacentpower you use the max power of conducted power from the mask (same logic as the out of band). Any document describing this ?

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,594 @@
...
+    Main routine to find CBSD indices on the move list.
+
+    Inputs:
+        protection_specs:   a list of protection specifications, a named tuple with
+                            fields of 'lowFreq' (in Hz), 'highFreq' (in Hz),
+                            'antHeight' (in meter), 'beamwidth' (in degree),
+                            'threshold' (in dBm/10MHz)
+        protection_points:  a list of protection points, each one being a tuple with
+                            named fields of 'latitude' and 'longitude'
+        registration_requests: a list of CBSD registration requests, each one being
```
kate-harrison: With the way that the test cases are written in the test spec, we will always be using information in the FAD (Full Activity Dump) format as input to the DPA move list reference model. A useful interface to this reference model would be something like:

Inputs:

- protection_specs
- protection_points
- SAS UUT FAD information
- List of FADs from SAS Test Harnesses

Output:

- Move list for SAS UUT, format:  [ (cbsd_id1, [grant_id1, grant_id4]), (cbsd_id3, [grant_id2]) ]
- List of move lists for SAS Test Harnesses (i.e. a list of lists, where each list is in the format above)

One option is to write a wrapper around this existing function, perhaps in a followup PR.

src/harness/reference_models/move_list_calc/test_data/GrantRequest_3.json
```diff
@@ -0,0 +1,7 @@
+{
+  "cbsdId": "sas1/cbsd1",
```
kate-harrison: Do you mean "sas1/cbsd3"? (Same below)

src/harness/reference_models/move_list_calc/test_data/RegistrationRequest_1.json
```diff
@@ -0,0 +1,25 @@
...
+  "userId": "John Doe",
+  "airInterface": {
+    "radioTechnology": "E_UTRA",
+    "supportedSpec": "LTE-Rel10"
+  },
+  "cbsdInfo": {
+    "vendor": "Equipment Maker C"
+  },
+  "cbsdSerialNumber": "4321dcba_1",
+  "measCapability": [ "EUTRA_CARRIER_RSSI_ALWAYS" ],
```
kate-harrison: FYI this meas. capability value is no longer supported. To make things easier, you could simply use [ ], which is what we use in many of the other test cases.

axsahoo:
When NIST contributed to the SAS UUT test code, we put language as “.. part of this code is developed by NIST….   US Govt. work is not subject to copyright ..” etc. You can look at the exact language in WINNF_FT_S_GRA_testcase.py. But in that file, other companies did not put any claim/copyright of their own. However , that file has Apache license agreement at the top.

So, I am not sure how it should be handled in this case (I am not a lawyer either).

thanks and regards

-Anirud

Anirudha (Anirud)  Sahoo
National Institute of Standards and Technology
Wireless Networks Division
Communications Technology Lab
100 Bureau Drive, Stop 6730
Gaithersburg, MD  20899
Ph: 301-975-4439

src/harness/reference_models/move_list_calc/README.md
```diff
@@ -0,0 +1,62 @@
...
+## Test Data
+
+To test the `move_list.findMoveList()` routine for a co-channel offshore DPA protection, 
+one can use the example given in the `move_list_example.py` file along with the test 
+data stored in the `\test_data` sub-folder. To run the example, use the command 
+`python move_list_example.py`.
+
+## Copyrights and Disclaimers
+
+This software was developed by employees of the National Institute of Standards and
```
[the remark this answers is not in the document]
kate-harrison: Anirud's reply:

"""
When NIST contributed to the SAS UUT test code, we put language as “.. part of this code is developed by NIST….   US Govt. work is not subject to copyright ..” etc. You can look at the exact language in WINNF_FT_S_GRA_testcase.py. But in that file, other companies did not put any claim/copyright of their own. However , that file has Apache license agreement at the top.

So, I am not sure how it should be handled in this case (I am not a lawyer either).

thanks and regards
-Anirud
"""


If you're OK to use "Some parts of this..." rather than "This..." (as you did for the grant testcase), that addresses all of my not-a-lawyer concerns :)

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,594 @@
...
+        # Compute EIRP within co-channel bandwidth
+        eirp_cbsd = (maxEirpPerMHz_cbsd - ant_gain_cbsd) + ant_gain \
+                    + 10 * np.log10(co_channel_bw / 1000000.0)
+
+        # Overlapping bandwidth between the first 10 MHz adjacent-channel of CBSD grant and
+        # the frequency range of protection constraint
+        adj_channel_bw = min(10000000.0, high_freq_c - low_freq_c - co_channel_bw)
+
+        # Incorporate EIRP within the adjacent-bandwidth into the total EIRP
+        if adj_channel_bw != 0:
```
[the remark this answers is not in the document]
msouryal: Thank you for the very helpful comments.  We've accepted most of them in the latest revision.

The calculation of interference will likely be clarified in a new point release of WINNF-TS-0112, at which time we'll update that part of the code accordingly.  The same is true for neighborhood definitions.

src/harness/reference_models/move_list_calc/README.md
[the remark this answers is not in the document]
msouryal: NIST requires that we use the current text as a disclaimer.  It does state that others may “use, copy, modify … provided that this notice and disclaimer of warranty appears in all copies.”

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -0,0 +1,594 @@
...
+    Inputs:
+        protection_specs:   a list of protection specifications, a named tuple with
+                            fields of 'lowFreq' (in Hz), 'highFreq' (in Hz),
+                            'antHeight' (in meter), 'beamwidth' (in degree),
+                            'threshold' (in dBm/10MHz)
+        protection_points:  a list of protection points, each one being a tuple with
+                            named fields of 'latitude' and 'longitude'
+        registration_requests: a list of CBSD registration requests, each one being
+                            a dictionary containing CBSD registration information
+        grant_requests:     a list of grant requests, each one being a dictionary
```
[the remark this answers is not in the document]
msouryal: If a single CBSD has more than one grant, there should be corresponding duplicate entries in the registration list.

src/harness/reference_models/move_list_calc/test_data/GrantRequest_3.json
msouryal: The cbsdId is not used by the move-list code, so its value is irrelevant.  The same is true of measCapability in the grant request.

src/harness/reference_models/move_list_calc/move_list.py
msouryal: Agree with the suggestion to write a wrapper.

src/harness/reference_models/move_list_calc/test_data/GrantRequest_3.json
kate-harrison: The value may not be relevant to the test, but presumably it would be better to make it at least "look right" anyway? Otherwise someone may think that the tests are outdated/flawed, e.g. if they saw an old measCapability being used.

src/harness/reference_models/move_list_calc/move_list.py
kate-harrison: Great! Is there any chance I can assume that NIST would do this? :)

src/harness/reference_models/move_list_calc/move_list.py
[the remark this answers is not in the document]
kate-harrison: Hmm, a bit strange (though I see why you did it), but I guess it can be handled in the wrapper without too much difficulty. It might be worth adding a comment here though.

jni2000:
look good to us as well. Please go ahead to approve if others agree.

arhannan:
We have looked at the code. It look okay by us. As Serge mentioned,  it can be updated when there are new specifications.

src/harness/reference_models/move_list_calc/move_list.py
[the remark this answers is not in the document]
msouryal: This approach was actually based on a suggestion you made to us back in August :).  But I agree with adding a clarifying comment.  We'll do that.

src/harness/reference_models/move_list_calc/test_data/GrantRequest_3.json
msouryal: Understood.  We have fixed the cbsdID values in the grant request messages and set measCapability to [] in the registration request messages.

src/harness/reference_models/move_list_calc/move_list.py
msouryal: Sure, we can write a wrapper.  Can you point us to agreed upon formats for the inputs?

Just to clarify, there is only one output move list, not a move list for the SAS UUT and a separate move list for each SAS Test Harness.  Are you asking for this single output list of grants to be segmented by their managing SAS?

src/harness/reference_models/move_list_calc/move_list.py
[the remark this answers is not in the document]
kate-harrison: I've already forgotten nearly everything that happened in August so I didn't even realize I was being self-contradictory. Sorry about that!

src/harness/reference_models/move_list_calc/move_list.py
kate-harrison: Thank you! I will send you the relevant information via email and I'll be happy to answer any clarifying questions (or redirect you to the right party).

I think that in the end we'll want to segment the move lists based on originating SAS so that we can use them directly in e.g. MCP.1. At the very least we'll need to distinguish between "SAS UUT ML" and "union of MLs for all SAS Test Harnesses", and I think we'll find it most convenient to split them fully.

PR #288, opened 2017-09-20
Registration test cases using new IDs: 1, 5, 9, 10.

harinineelu:
registration_testcase.py → WINNF_FT_S_REG_testcase.py
Old REG_3 → REG_1
Old REG_11 → REG_5
Old REG_18 → REG_9
Old REG_20 → REG_10

Note: Intentionally made minimal changes to the code.


assaneRed:
can you, please, separate your PR by using one or two testcases by PR

assaneRed:
can you, please, separate your PR  using one or two testcases by PR?

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,266 @@
...
+    device_c = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_c.json')))
+
+    # Pre-load conditionals
+    conditionals_a = {
+        'cbsdCategory': device_a['cbsdCategory'],
+        'fccId': device_a['fccId'],
+        'userId': device_a['userId'],
+        'cbsdSerialNumber': device_a['cbsdSerialNumber'],
+        'airInterface': device_b['airInterface'],
```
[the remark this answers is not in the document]
harinineelu: Done.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,266 @@
...
+        'installationParam': device_b['installationParam']
```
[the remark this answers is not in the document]
harinineelu: Done.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,266 @@
...
+        'fccId': device_c['fccId'],
+        'userId': device_c['userId'],
+        'cbsdSerialNumber': device_c['cbsdSerialNumber'],
+        'airInterface': device_c['airInterface'],
+        'installationParam': device_c['installationParam']
+    }
+    conditionals = {
+        'registrationData': [conditionals_a, conditionals_b, conditionals_c]
+    }
+    self._sas_admin.PreloadRegistrationData(conditionals)
```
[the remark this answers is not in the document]
harinineelu: Done.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,266 @@
...
+
+    # Inject FCC IDs
+    self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
+    self._sas_admin.InjectFccId({'fccId': device_b['fccId']})
+    self._sas_admin.InjectFccId({'fccId': device_c['fccId']})
+
```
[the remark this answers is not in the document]
harinineelu: Done.

harinineelu:
Thank you for reviewing it this time with 4 test cases. I think the review should be simple enough (since it's based heavily on existing code) but I'll make smaller PRs in the future. 

I have made the changes you had suggested. Could you please take a look. Thanks!

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,266 @@
...
+        'fccId': device_a['fccId'],
+        'userId': device_a['userId'],
+        'cbsdSerialNumber': device_a['cbsdSerialNumber'],
+        'airInterface': device_b['airInterface'],
+        'installationParam': device_b['installationParam']
+    }
+    conditionals_b = {
+        'cbsdCategory': device_b['cbsdCategory'],
+        'fccId': device_b['fccId'],
+        'userId': device_b['userId'],
```
jcereced: userID is a "Required" parameter and not "Reg-Conditional" do we need to include it as part of the conditional parameters that get loaded into the SAS (as in the CPI would be loading them)?

harinineelu: Yes, The SAS UUT needs the user ID, FCC ID, and serial # so that it can link the conditional reg. info with the device itself. I agree the code could be a little clearer, but we intentionally made only minimal changes to the existing code.

jcereced: It needs the FCC ID and serial number (that makes up the CBSD ID) but it should not need the user ID.  The test is trying to emulate the scenario where a CPI inserts data into the SAS and then a registration occurs.  When a CPI enters data into the SAS they will enter REG-Conditional data and the user ID is not part of that.  It's quite possible the the CPI entering the data doesn't even know what the user ID that will register the CBSD later would be

kate-harrison: User ID is a required field in RegistrationRequest according to the SAS-CBSD TS, so it needs to be present here. (The only other required fields are serial number and FCC ID.)

jcereced: the User ID will be included in the RegistrationRequest (as is required) but I can't find anywhere where it says that it must be provided by the CPI when the other Reg-Conditional parameters are provided.  

The TC (using Reg1 as an example) says 

"All REGConditional
parameters for the three CBSDs (a combination of both Cat A and
Cat B) should already be pre-loaded into the SAS."

User Id is required (but not Reg-Conditional) therefore it should be provided in the registrationirequest message but it is not needed as part of the "load conditional parameters into the SAS"

When you are marrying the incomplete registration-request with the data preloaded into the SAS you can do so using just FCC-ID and serial number.

It's possible that there is a CPI document somewhere that says that user ID must be loaded with the other Reg-Conditionals and I just haven't seen it yet; if so can someone point me to it?

kate-harrison: Oops, sorry, I was too eager to reply to your comment and missed which line this was referencing :)

It's also not clear to me whether user ID is currently needed, but it feels likely that it would be. I suspect WG5 is still working out those details. I propose we pick one path for now and add an item to the issue tracker (https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues) to look into this and fix any code, if needed. Would you be OK to leave the proposed code as-is for now, pending resolution of this issue?

jcereced: sounds great to me!

The only reason I feel the user Id doesn't need to be in there is because we are mimicking how things would work when a CPI inputs data into the SAS and i don't think CPIs are going to have access to user ID but they will know the FCC ID and the serial number of the CBSD they are entering data for.

kate-harrison: Thanks! Issue created: https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues/290

harinineelu:
Removed the userId from the conditional registration data.

kate-harrison:
Javi and/or Assane, can you please complete your review of this PR? Thanks!

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,262 @@
...
+    device_c = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_c.json')))
+
+    # Pre-load conditionals
+    conditionals_a = {
+        'cbsdCategory': device_a['cbsdCategory'],
+        'fccId': device_a['fccId'],
+        'cbsdSerialNumber': device_a['cbsdSerialNumber'],
+        'airInterface': device_a['airInterface'],
+        'installationParam': device_a['installationParam']
```
jcereced: Do we need to load measCapability into the SAS-under-test when we are loading the conditional parameters?  A value of "measCapability": [] is an accepted value indicating that the CBSD has meascapability="none".  I'm reading the latest version of part 1 (released 10/13/2017) and it says "All REGConditional
parameters for the three CBSDs (a combination of both Cat A and
Cat B) shall already be pre-loaded into the SAS" so it seems like we should load meascapability into the SAS-under test even if that value is a blank value which means none.......the end result should be the same but if we want to fully follow the instructions we may want to load a 'measCapability': [ ] along wiht the other ocnditionals...thoughts?

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,262 @@
...
+    del device_a['cbsdCategory']
+    del device_a['airInterface']
+    del device_a['installationParam']
+    del device_b['cbsdCategory']
+    del device_b['airInterface']
+    del device_b['installationParam']
+    del device_c['cbsdCategory']
+    del device_c['airInterface']
+    del device_c['installationParam']
+
```
jcereced: if we decide to inject measCapability with the other conditionals we would have to remove it from the registration message here as well

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,262 @@
...
+    self._sas_admin.InjectFccId({'fccId': device_d['fccId']})
+
+    # Inject User IDs
+    self._sas_admin.InjectUserId({'userId': device_a['userId']})
+    self._sas_admin.InjectUserId({'userId': device_b['userId']})
+    self._sas_admin.InjectUserId({'userId': device_c['userId']})
+    self._sas_admin.InjectUserId({'userId': device_d['userId']})
+
+    self._sas_admin.PreloadRegistrationData(conditionals)
+
```
jcereced: maybe i'm missing it ...but shouldn't we remove the conditionals from the registration message like we did above for test Id #1?

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,262 @@
...
+    """Blacklisted CBSD in Array Registration request (responseCode 101).
+
+    The response should be BLACKLISTED 101.
+    """
+
+    # Load Devices
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
```
jcereced: device B is a Cat B cbsd and it should not be able to Single-step register unles the CPI data is encoded and signed by a valid CPI.  For the purposes of this test we should only use Cat A CBSDs (easier to single-step- register in current code

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -0,0 +1,262 @@
...
+    The response should be FAILURE 100.
+    """
+
+    # Use higher than supported version
+    self._sas._sas_version = 'v2.0'
+
+    # Load Devices
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_b = json.load(
```
jcereced: same comment as in test Id #9, we should use all cat A CBSDs for this test

src/harness/testcases/WINNF_FT_S_REG_testcase.py
harinineelu: Agreed, Done.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
harinineelu: Done.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
harinineelu: Yes, Done.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
harinineelu: Done.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
harinineelu: Done.

harinineelu:
Thanks for the review. Fixed all. Could you please take a look.

harinineelu:
Hi Javi,
Apologies for having 4 test cases here, which is the primary cause of this confusion. :) 

To make things clear:
REG_1 uses Cat A and Cat B devices.
REG_5, 9 and 10 have only Cat A devices. 
Could you please take a look. Thanks!

kate-harrison:
I'm taking the liberty of merging now even though Red marked themselves as a reviewer since Javi, the most active reviewer, has approved these changes. Assane, if you have any comments please let us know and we'll be happy to quickly fix any errors.

PR #293, opened 2017-09-26
Add geodata submodule retrieved from WinnForum/SAS-Data repo.

sbdt-google:
Introduce the SAS-data repository as a submodule under data/geo.

Discussions within WinnForum led to the continuing use of Git LFS to hold the geo reference data (for NED and NLCD) but in a separate repository at: https://github.com/Wireless-Innovation-Forum/SAS-Data

This PR simply introduce the ability to see this repository as a GIT submodule under data/geo

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "data/geo"]
+	path = data/geo
+	url = https://github.com/Wireless-Innovation-Forum/SAS-Data


sbdt-google:
Thanks for the review. Could you also approve the PR #281 ?  (Remove all obsolete NED and NLCD tiles)
so that I submit both of them at same time, and we end up with a clean database of geo tiles (not 2).

janetliuboston:
Approved PR #281

sbdt-google:
FYI Janet: I submitted but then reverted as the submodule ended up not appearing as it should.
I will fix this early next week.
Also note that there is a related PR #280 on the scripts for manipulating the tiles (cleanup of obsolete scripts, add unzip convenience script and add the scripts used to produced the NLCD tiles). If you have the possibility you can put yourself as reviewer on that one as well.

PR #295, opened 2017-09-27
NEW Spectrum Inquiry test cases using new IDs: 11.

harinineelu:
Old SIQ_20 → SIQ_11

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
@@ -1,0 +1,90 @@
+#    Copyright 2016 SAS Project Authors. All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -0,0 +1,90 @@
...
+  def test_WINNF_FT_S_SIQ_11(self):
+    """Unsupported frequency range inquiry array.
+
+    The response for Inquiry #2 should be UNSUPPORTED_SPECTRUM, code 300
+    """
+    # Register the devices
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
```
jcereced: device B is Category B and cannot single-step register unless CPIData is embedded in registration request so we need to do that or do multi-step reg and load conditional parameters.  I think the easier fix is to use device_c which is Category A and can single-step register

harinineelu: Done.

harinineelu:
Thanks for reviewing. 
 - Changed device_b to device_c
Could you please take a look.

PR #297, opened 2017-10-06
NEW HBT_9 Heartbeat Test Case

RedaHossino:


src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+    that impacts incumbent activity.
+
+    successful heartbeat for the grants of two cbsds that don't impact the incumbent
+    and failed with responseCode 500 for the third one.
+    """
+    # STEP 1
+    # register three devices 
+    device_c = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_c.json')))
+    device_d = json.load(
```
kate-harrison: Device D is  Cat B and therefore we can't use single-step registration for it. We usually use A, C, E when we need three CBSDs and the category is unspecified.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+        open(os.path.join('testcases', 'testdata', 'device_d.json')))
+    device_e = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_e.json')))
+    devices = [device_d, device_e, device_c]
```
kate-harrison: Why not use the same order as above?

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+    for device in devices:
+        # inject FCC IDs
+        self._sas_admin.InjectFccId({'fccId': device['fccId']})
```
kate-harrison: Please add a comment to later inject user ID. The code to add the function to the admin interface is in a pending PR from Google; hopefully we can submit that soon so that you can update the code instead of leaving a "TODO".

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+    cbsd_ids.sort(key=lambda cbsdId: cbsdId)
+    del request, response    
+    # create and send grant requests
+    grant_c = json.load(
+        open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_c['cbsdId'] = cbsd_ids[0]
+    grant_d = json.load(
+        open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_d['cbsdId'] = cbsd_ids[1]
+    grant_d['operationParam']['operationFrequencyRange']['lowFrequency'] = 3560
```
kate-harrison: The units here are Hz, not MHz.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+    grant_d['operationParam']['operationFrequencyRange']['highFrequency'] = 3570
+    grant_e = json.load(
+        open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_e['cbsdId'] = cbsd_ids[2]
+    grant_e['operationParam']['operationFrequencyRange']['lowFrequency'] = 3560
```
kate-harrison: This range isn't consistent with the latest test spec.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
kate-harrison: Why change this value for the second device? (Maybe it was in the old test spec but not the new one?)

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+            'grantId': grant_id,
+            'operationState': status
+        })
+    return {'heartbeatRequest': heartbeat_request}
+
+  def sendAndAssertSuccessfulHeartBeat(self, cbsd_ids, grant_ids, previousTransmitExpireTime):    
+    request =  self.getHeartBeatRequest(cbsd_ids, grant_ids, previousTransmitExpireTime)
+    response = self._sas.Heartbeat(request)['heartbeatResponse']
+    # check Heartbeat response 
+    self.assertEqual(len(response), 3)
```
kate-harrison: You really can't hard-code this value in a general function.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+    for response_num, resp in enumerate(response):
+        self.assertEqual(resp['cbsdId'], cbsd_ids[response_num])
+        self.assertEqual(resp['grantId'], grant_ids[response_num])
+        self.assertEqual(resp['response']['responseCode'], 0)
+        grant_expire_time_1 = datetime.strptime(resp['grantExpireTime'],'%Y-%m-%dT%H:%M:%SZ')
+        transmit_expire_time_1 = datetime.strptime(resp['transmitExpireTime'],'%Y-%m-%dT%H:%M:%SZ')
+        self.assertLessEqual((transmit_expire_time_1 - datetime.utcnow()).total_seconds(), 240)
+        self.assertLessEqual(transmit_expire_time_1, grant_expire_time_1)
+    del request
```
kate-harrison: There's no value to deleting the variable just before it goes out of scope anyway.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+
+class HeartbeatTestcase(sas_testcase.SasTestCase):
+  def setUp(self):
+    self._sas, self._sas_admin = sas.GetTestingSas()
+    self._sas_admin.Reset()
+
+  def tearDown(self):
+    pass
+
+  def getHeartBeatRequest(self,cbsd_ids, grant_ids, previousTransmitExpireTime):
```
kate-harrison: Do we really need all of this machinery? Since it's a very simple message, we usually just create the heartbeat request in the test case itself so that it's easier to read (and verify that it's doing the right thing). This is especially true for the HBT test cases where we often want to create the request in a very specific way. Is there a reason to do something else?

I'm also worried about the correctness of this function, but I'll hold off on those comments until we establish that it's a useful function to have.

(Same comment for the following function.)

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+    del request, response
+    # First Heartbeat
+    first_HB_response = self.sendAndAssertSuccessfulHeartBeat(cbsd_ids, grant_ids, [None, None, None])
+    transmitExpireTime = (first_HB_response[0]['transmitExpireTime'], first_HB_response[1]['transmitExpireTime'], first_HB_response[2]['transmitExpireTime'])
+    # STEP 2
+    # load and inject FSS data
+    fss_c = json.load(
+        open(os.path.join('testcases', 'testdata', 'fss_record_c.json')))
+    # Inject Incumbent Activity with Overlapping Frequency of CBSD
+    fss_c['deploymentParam']['operationParam']['operationFrequencyRange']['lowFrequency'] = \
```
kate-harrison: Isn't this supposed to overlap with the third CBSD's grant, not the second?

Also it probably makes more sense to set this to 3650-4200 which is more consistent with the FSS data we expect (and you can do this in the file below).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+                             grant_c['operationParam']['operationFrequencyRange']['lowFrequency']
+    fss_c['deploymentParam']['operationParam']['operationFrequencyRange']['highFrequency'] = \
+                             grant_c['operationParam']['operationFrequencyRange']['highFrequency']
+    
+    self._sas_admin.InjectFss({'record': fss_c})   
+    # load and inject GWPZ data
+    gwpz_c = json.load(
+        open(os.path.join('testcases', 'testdata', 'gwpz_record_c.json')))
+    # Inject Incumbent Activity with Overlapping Frequency of CBSD
+    gwpz_c['deploymentParam']['operationParam']['operationFrequencyRange']['lowFrequency'] = \
```
kate-harrison: Let's choose a range that's more in line with the real data we expect for GWPZs (and you can do it in the file below; there's no need to override the value here).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+                             grant_c['operationParam']['operationFrequencyRange']['lowFrequency']
+    gwpz_c['deploymentParam']['operationParam']['operationFrequencyRange']['highFrequency'] = \
+                             grant_c['operationParam']['operationFrequencyRange']['highFrequency']
+    
+    self._sas_admin.InjectWisp({'record': gwpz_c})
```
kate-harrison: What about the 'zone' key? See https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/sas_interface.py#L288

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+    # Trigger daily activities and wait for it to get it complete
+    self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
+    # Send Heartbeat after triggering the IAP
+    request =  self.getHeartBeatRequest(cbsd_ids, grant_ids, transmitExpireTime)
+    response = self._sas.Heartbeat(request)['heartbeatResponse']
+    # CHECK
+    for response_num, resp in enumerate(response):
+        self.assertEqual(resp['cbsdId'], cbsd_ids[response_num])
+        self.assertEqual(resp['grantId'], grant_ids[response_num])
+        grant_expire_time_2 = datetime.strptime(resp['grantExpireTime'],'%Y-%m-%dT%H:%M:%SZ')
```
kate-harrison: This variable name feels out of place. Why do you have "_2" at the end? (Same for other variable names.)

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+        transmit_expire_time_2 = datetime.strptime(resp['transmitExpireTime'],'%Y-%m-%dT%H:%M:%SZ')
+        self.assertLessEqual((transmit_expire_time_2 - datetime.utcnow()).total_seconds(), 240)
+        self.assertLessEqual(transmit_expire_time_2, grant_expire_time_2)
+        if cbsd_ids[response_num] == cbsd_ids[2]:
```
kate-harrison: There's an easier way to write this condition, although normally we would split this block of code into two pieces to make it clearer:

// Checks for CBSDs 1, 2
// Check for CBSD 3 (since it's different)

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,155 @@
...
+            self.assertEqual(resp['response']['responseCode'], 500)
+            if request['heartbeatRequest'][response_num]['operationState'] == 'AUTHORIZED':
```
kate-harrison: This condition has changed in the latest test spec.

src/harness/testcases/testdata/fss_record_c.json
```diff
@@ -0,0 +1,25 @@
...
+  "id": "incumbent/ibfs/0",
+  "type": "FSS",
+  "deploymentParam": 
+    {
+      "installationParam": {
+        "latitude": 38.815291,
+        "longitude": -97.098805,
+        "height": 7.6,
+        "heightType": "AGL",
+        "indoorDeployment": false,
```
kate-harrison: Is this field applicable for an FSS record?

src/harness/testcases/testdata/fss_record_c.json
```diff
@@ -0,0 +1,25 @@
...
+        "antennaGain": 16,
+        "antennaBeamwidth": 30,
+        "antennaAzimuth": 270,
+        "antennaDowntilt": -90
+      },
+      "operationParam": {
+        "maxEirp": 10.1,
```
kate-harrison: Does this make sense for an FSS record?

src/harness/testcases/testdata/fss_record_c.json
```diff
@@ -0,0 +1,25 @@
+{
```
kate-harrison: I think you can just name this "fss_record.json" -- no need for the "_c" (same for the GWPZ file below).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: you are right, for the next function, because it is used only once in the test, but this function is used two times in the test, and i think it will be useful for other HB test cases so don't you think it is better to reuse it?
I think for this function it is also related to the new added machineryfor choosing the right status of the grant inside the HB request.
we need to verify that it is synchronized with the status in SAS" especially after the IAP which can take some time", other wise the SAS will send response 502, right?
so if you don't agree on this status machinery, we don't need to have this function.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: thank you this point is fixed, please verify this in the commits I added.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: you are right, thank you, it is done.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: I changes the used devices, to be a,c and e, thank you

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: I modified the registration and grant requests, so I use assertRegisteredAndGranted

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: you are right, thank you, it is done.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: I modified it, thank you.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: thank you, I modifies it to be on 3650-4200 MHz.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: I reset the range of the record to 3650-3700 MHz, which seems ok according to FCC data, right?

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: fixed, thank you

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: yes, you are right, I updated it.
thank you

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: the issues of the comments related to the record files are done, please verify them in the latest change.
for the zone, I added it a polygon(rectangle) around the center, but we should modify it later according to realistic GWPZ shape data, is that ok ?
thank you

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,135 @@
...
+    # First Heartbeat
+    request =  self.getHeartBeatRequest(cbsd_ids, grant_ids, [None, None, None])
+    response = self._sas.Heartbeat(request)['heartbeatResponse']
+    # check Heartbeat response 
+    self.assertEqual(len(first_hb_response), len(grant_ids))
+    for response_num, resp in enumerate(first_hb_response):
+        self.assertEqual(resp['cbsdId'], cbsd_ids[response_num])
+        self.assertEqual(resp['grantId'], grant_ids[response_num])
+        self.assertEqual(resp['response']['responseCode'], 0)
+        grant_expire_time = datetime.strptime(resp['grantExpireTime'],\
```
kate-harrison: I think we can avoid doing this and the following checks since they are not requested by the spec.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,135 @@
...
+    # Send Heartbeat after triggering the IAP
+    request =  self.getHeartBeatRequest(cbsd_ids, grant_ids,\
+                                        first_hb_transmit_expire_time)
+    response = self._sas.Heartbeat(request)['heartbeatResponse']
+    # CHECK
+    # Check first two CBSDs with a successful HB response
+    for response_num, resp in enumerate(response[:2]):
+        self.assertEqual(resp['cbsdId'], cbsd_ids[response_num])
+        self.assertEqual(resp['grantId'], grant_ids[response_num])
+        grant_expire_time = datetime.strptime(resp['grantExpireTime'],\
```
kate-harrison: We don't expect grantExpireTime to be set in the HeartbeatResponse (except if grantRenew==True, which it does not in this case). This value should be obtained from the GrantResponse.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,135 @@
...
+                                                 '%Y-%m-%dT%H:%M:%SZ')
+        self.assertLessEqual((transmit_expire_time - \
+                              datetime.utcnow()).total_seconds(), 240)
+        self.assertLessEqual(transmit_expire_time, grant_expire_time)
+        self.assertEqual(resp['response']['responseCode'], 0)
+    # Check the third  CBSD with unsuccessful HB response of code 500
+    self.assertEqual(response[2]['cbsdId'], cbsd_ids[2])
+    self.assertEqual(response[2]['grantId'], grant_ids[2])
+    self.assertEqual(response[2]['response']['responseCode'], 500)
+    self.assertLessEqual((datetime.utcnow() - \
```
kate-harrison: You can compare the times directly without having to subtract.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,135 @@
...
+                                   first_hb_response[2]['transmitExpireTime'])
+    # STEP 2
+    # load and inject FSS data with Overlapping Frequency of CBSD
+    fss_e = json.load(
+        open(os.path.join('testcases', 'testdata', 'fss_record.json')))
+    self._sas_admin.InjectFss({'record': fss_e})   
+    # load and inject GWPZ data with Overlapping Frequency of CBSD 
+    gwpz_e = json.load(
+        open(os.path.join('testcases', 'testdata', 'gwpz_record.json')))  
+    self._sas_admin.InjectWisp({'record': gwpz_c})
```
kate-harrison: Given the way you defined gwpz_record.json, you don't need 'record' here.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
kate-harrison: Aha, now I understand your concern. I think the machinery is unnecessary during the first heartbeat request (we know that it should always be GRANTED). For the second heartbeat request, I think we can safely use GRANTED as well -- even if the SAS thinks the Grant is AUTHORIZED, it should not give an error. What do you think?

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
kate-harrison: Agreed -- thanks!

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
kate-harrison: This sounds good!

src/harness/testcases/testdata/fss_record_c.json
kate-harrison: Sorry, my original suggestion was slightly flawed -- I should have suggested "fss_record_0.json" (so that we can create more in the future, if needed; this will match what we've done for similar data files).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: yeah, you are right, I thought the CBSD should always send in HB the status which exists in the SAS otherwise the SAS will send error(502), but I think you are right if the CBSD sends granted status in the HB request and its status on the SAS is Authorized it is safe, and there should be no error.
We don't need this machinery then, thank you.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: thank you for the review, I updated the test according to the requested changes.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,131 @@
...
+    grant_c = json.load(
+        open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_c['cbsdId'] = cbsd_ids[1]
+    grant_e = json.load(
+        open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_e['cbsdId'] = cbsd_ids[2]
+    grant_e['operationParam']['operationFrequencyRange']\
+                             ['lowFrequency'] = 3650000000
+    grant_e['operationParam']['operationFrequencyRange']\
+                             ['highFrequency'] = 3700000000
```
kate-harrison: Why did you change this to 3660000000? What you had before matches the spec (although this version does as well).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,131 @@
...
+    request = {'grantRequest': [grant_a, grant_c, grant_e]}
+    # send grant request
+    response = self._sas.Grant(request)['grantResponse']
+    grant_ids = []
+    grant_expire_times = []
+    for response_num, resp in enumerate(response):
```
kate-harrison: Please assert that the responseCode for each is 0 (it's not requested by the spec but it could help a lot during debugging and doesn't cause any problems otherwise).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,131 @@
...
+    response = self._sas.Heartbeat({'heartbeatRequest': request})['heartbeatResponse']
+    del request
+    # CHECK
+    # Check first two CBSDs with a successful HB response
+    for response_num, resp in enumerate(response[:2]):
+        self.assertEqual(resp['cbsdId'], cbsd_ids[response_num])
+        self.assertEqual(resp['grantId'], grant_ids[response_num])
+        self.assertEqual(resp['response']['responseCode'], 0)
+        transmit_expire_time = datetime.strptime(resp['transmitExpireTime'],\
+                                                 '%Y-%m-%dT%H:%M:%SZ')
```
kate-harrison: Please also check that transmit_expire_time >= Now().

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,133 @@
...
+    # First Heartbeat
+    request = []
+    for cbsd_id, grant_id in zip(cbsd_ids, grant_ids):
+        request.append({
+            'cbsdId': cbsd_id,
+            'grantId': grant_id,
+            'operationState': 'GRANTED'
+        })
+    
+    response = self._sas.Heartbeat({'heartbeatRequest': request})['heartbeatResponse']
```
kate-harrison: Please assert that the HB response code for each is 0.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,133 @@
...
+    
+    del request, response
+    # STEP 2
+    # load and inject FSS data with Overlapping Frequency of CBSD
+    fss_e = json.load(
+        open(os.path.join('testcases', 'testdata', 'fss_record_0.json')))
+    self._sas_admin.InjectFss({'record': fss_e})   
+    # load and inject GWPZ data with Overlapping Frequency of CBSD
```
kate-harrison: Technically this should be GWBL data. I think you can just change "gwpz" to "gwbl" here and in the name of the file to avoid confusion.

src/harness/testcases/testdata/gwpz_record_0.json
```diff
@@ -0,0 +1,58 @@
...
+      },
+      "operationParam": {
+        "operationFrequencyRange": {
+          "lowFrequency": 3650000000.0,
+          "highFrequency": 3700000000.0
+        }
+      }
+    }
+},
+"zone":{
```
kate-harrison: I think you can actually leave out the zone here since the intent is to inject a GWBL, not a GWPZ (see the documentation [here](https://docs.google.com/document/d/19YbeQsSzm5kCIOeVzsF8piqI0Lizy3p345aiVGf-Hlg/edit?pli=1#heading=h.rz2qbup2aut9) which someone else helpfully reminded me of on Tuesday; sorry for the back-and-forth on this, it's completely my fault).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
redtechnologies: Thank you for the very good review, I addressed them, please feel free to look at them.

RedaHossino: Thank you for the very good review, I addressed them, please feel free to look at them.

src/harness/testcases/testdata/fss_record_0.json
```diff
@@ -0,0 +1,23 @@
+{
+  "id": "incumbent/ibfs/0",
+  "type": "FSS",
+  "deploymentParam":
```
kate-harrison: I think Takashi's comment from SIQ.2 applies here as well: deploymentParam should be an array of DeploymentParam objects rather than a single object (same for the GWBL below). I'll try to make this clearer in the testing API doc -- I think it's my fault that it's not very clear right now.

PR #299, opened 2017-10-06
Revert "Add geodata submodule retrieved from WinnForum/SAS-Data repo."

sbdt-google:
Reverts Wireless-Innovation-Forum/Spectrum-Access-System#293


sbdt-google:
Reverting this changes as it does not fully perform as expected. Will fix it on further CL

PR #302, opened 2017-10-09
Fix bugs in REG_5, REG_22, REG_25

norbertbarbosa:


norbertbarbosa:
Thanks for this very quick review

PR #303, opened 2017-10-16
Cleanup of the geo data retrieval/creation scripts.

sbdt-google:
Identical to previous #280, but new PR cleanup to be on the latest master  with previous cleanup done (for ease of review mainly).

The NED and NLCD tiles snapshot are in repository Wireless-Innovation-forum/SAS-data under zip format for each tile.

Scripts provided:
- extract_geo.py: extract the zip tiled from the data/geo/ submodule
(requires that the SAS-data repository included as submodule under
/data/geo/ - see other PR)
- retrieve_orig_nlcd.py, retile_nlcd.py and nlcd_origin.py: creation of
the reference NLCD tiles from original data. Provided as the reference
code which has been used to create the reference NLCD data (for replay).
- retrieve_ned_files.py: obtain all NED data from USGS. Not to be used to
get reference NED data, as the data are changing regularly. The
reference data is a snapshot made in July using this script.

Obsolete scripts are removed as well: extract_ned.py, retile_nlcd.sh,
retrieve_ned_nlcd_files.py


sbdt-google:
I will submit this PR since:
 - it is identical to #280 provided more than 6 weeks ago, and no one stepped up for review
 - just a reference PR on data retrieval scripts and NLCD database creation. So not directly related or used by harness tests

If any modification needs to be done, it can be provided as part of further PR.

PR #306, opened 2017-10-17
NEW Heartbeat test case using new IDs: 1.

harinineelu:
heartbeat_testcase.py → WINNF_FT_S_HBT_testcase.py
Old HBT_2 → HBT_1


jcereced:
sorry i was off by one version...latest doc is 5.0 but HBT1 is the same as I described above

harinineelu:
Thanks for catching it, Javi. I've changed the code to match the latest spec, now all 3 devices enter the Authorized state. Could you please take a look.

PR #308, opened 2017-10-17
NEW Grant test case using new IDs: 16.

harinineelu:
grant_testcase.py → WINNF_FT_S_GRA_testcase.py
Old GRA_28 → GRA_16


src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -0,0 +1,112 @@
...
+    request = {'grantRequest': [grant_0, grant_1]}
+    # Send grant request and get response
+    response = self._sas.Grant(request)['grantResponse']
+    # Check grant response
+    self.assertEqual(len(response), 2)
+    self.assertEqual(response[0]['cbsdId'], cbsd_id)
+    self.assertEqual(response[1]['cbsdId'], cbsd_id)
+    self.assertTrue(response[0]['response']['responseCode'] == 401
+                    or response[1]['response']['responseCode'] == 401)
+    for resp in response:
```
jcereced: so we are checking that either one of the response codes is 401, but what happens if both response codes are 401?  Looks like the test will still pass since the following check we do for the grantexpirytime and hb interval only occurs if the response code if 0.  If both response codes are 401 the next check is skipped and the test cases passes.

Should we also add a check that either one of the response codes is 0, so we would be checking that one is 0 and the other is 401?

harinineelu: Thanks for reviewing, Javi! 

This implementation seems to be on the understanding that the spec specifies: "at least one element has responseCode == 401" and so clearly "both == 401" is allowed by the spec? 

Please let us know if this is not how it should be interpreted or if it needs further discussion.

jcereced: So right after it tells you to check that at least one of the responses is code 401 it says "The grantExpireTime with a correct format and range is included for the
element with responseCode = 0".  The way I read that it seems to imply that one of the responses is 401 and the other one needs to be 0 (even though it doesn't spell it out).  Since both grant request are getting there at essentially the same time the one that gets processed first should yield a response "0" and the one that gets process after should yield a "401".  The reason it checks that at least one is 401 is because we don't know which one will get granted first so your response should look like {0, 401} or {401,0}.

it would be a lot easier if they had said "One of the responses should be 401 and the other one should be 0"


I'm fine with leaving it the way it is but we may be asked to modify later on when ITS reviews the final code

harinineelu: I agree, Javi.

jcereced: did we decide what to do on this one (make changes or mark approved as is)?  

thanks

harinineelu: I'm fine with the code as it is and modify later when asked. Thanks for your quick reviews, Javi

PR #311, opened 2017-10-18
Make fccMaxEirp an optional param in InjectFccId()

harinineelu:
Add fccMaxEirp as an optional parameter to the InjectFccId() function with a default value of 47 dBm/10 MHz.


kate-harrison:
Thank you!!

PR #314, opened 2017-10-20
NEW Deregistration test cases using new IDs: 1, 2

jcereced:
NEW Deregistration test cases using new IDs: 1, 2

DRG1
DRG2


src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -0,0 +1,102 @@
...
+      self.assertEqual(resp['response']['responseCode'], 0)
+      cbsd_ids.append(resp['cbsdId'])
+    del request, response
+
+    # Deregister the device
+    request = {'deregistrationRequest': [
+        {'cbsdId': cbsd_ids[0]},
+        {}]}
+    response = self._sas.Deregistration(request)['deregistrationResponse']
+    # Check the deregistration response
```
kate-harrison: Let's check that the length of `response` is equal to 2 since the CHECK step says it explicitly.

jcereced:
Change made, please check, thanks!

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -0,0 +1,127 @@
...
+
+    # Send the grant requests
+    request = {'grantRequest': [grant_1, grant_2]}
+    response = self._sas.Grant(request)['grantResponse']
+
+    # Check the grant responses
+    # valid cbsdIds, responseCode should be 103 for both cbsdIds
+    self.assertEqual(len(response), 2)
+    for x, resp in enumerate(response):
+      self.assertEqual(resp['cbsdId'], cbsd_ids[x])
```
kate-harrison: I just noticed that the spec here doesn't seem to make much sense. If the CBSD has been deregistered, the CBSD ID is no longer valid and therefore should not be returned. Do you agree? (Your code matches the spec 100%, I'm just not sure the spec is right.)

jcereced: That's what happens when you follow directions!.....You are right , the CBSDs were deregistered so they should no longer exist in the SAS.  When the SAS responds to a non-existent CBSD ID request there should be no CBSD ID in the response (GRA3, p39)

kate-harrison: It's good that you did, otherwise I wouldn't have caught the error in the spec :) I will bring it up at today's WG4 meeting and one way or another hopefully we can get it changed. I think you have two options: (1) submit now and make a small PR later to remove this line, assuming I'm successful or (2) wait until this evening when I can give a status report from WG4 and you might be able to save yourself a small amount of work. Your choice :)

jcereced: I choose #2..thanks!!

kate-harrison: Good news: they approved this change (to remove the CBSD ID check) this afternoon!

jcereced:
i removed the cbsd Id check.  thanks!

PR #318, opened 2017-10-26
Provide  HAAT calculation routine.

sbdt-google:
New function wf_itm.ComputeHaat().

Note: this is provided as part of  wf_itm which manages the terrain driver.
A future PR will have some cleanup so that the wf_itm module is the unique source of data drivers (currently duplicated in wf_hybrid).


sbdt-google:
You are right, but since every radial has exactly the same number of points by definition, the 2 approaches are equivalent:
  1/8 . sum_k(average_k) = 1/8. sum_k ( 1/N . sum_j(height_kj) = 1/(8.N) . sum_kj (height_kj)

PR #325, opened 2017-11-03
NEW Grant test case using new ID: 10

harinineelu:
Old GRA_20 → GRA_10


src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -133,4 +134,87 @@ def test_WINNF_FT_S_GRA_7(self):
...
+    response = self._sas.Grant(request)['grantResponse']
+    # Check grant response, must be response code 0.
+    self.assertEqual(len(response), 2)
+    for response_num in [0, 1]:
+      self.assertEqual(response[response_num]['cbsdId'], cbsd_ids[response_num])
+      self.assertTrue('grantId' in response[response_num])
+      self.assertTrue('grantExpireTime' in response[response_num])
+      datetime.strptime(response[response_num]['grantExpireTime'],
+                        '%Y-%m-%dT%H:%M:%SZ')
+      self.assertTrue('heartbeatInterval' in response[response_num])
```
jcereced: The TC says "The heartbeatInterval parameter set to valid value with correct format. "   but it doesn't specify what a valid value, do they mean >0?  should we just check 'heartbeatInterval'>0?  since we are not transmitting yet I don't believe a maximum value is enforced at this point.

harinineelu: Yes, Thanks for spotting it. Added 'heartbeatInterval' format checks. Could you please take a look.

PR #326, opened 2017-11-03
NEW SIQ test case using new ID: 9

harinineelu:
Old SIQ_18 → SIQ_9


src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -206,6 +206,154 @@ def test_WINNF_FT_S_SIQ_8(self):
...
+          all(channel['frequencyRange']['highFrequency'] <= 3700000000
+              for channel in response[resp_num]['availableChannel']))
+    # Check the frequency range of assigned PAL and GAA channels.
+    for channel in response[0]['availableChannel']:
+      if channel['channelType'] == 'PAL':
+        self.assertTrue(
+            channel['frequencyRange']['lowFrequency'] == 3550000000)
+        self.assertTrue(
+            channel['frequencyRange']['highFrequency'] == 3560000000)
+      if channel['channelType'] == 'GAA':
```
RedaHossino: if we deleted the first loop, this can be "else" with assert that ['channelType'] == 'GAA', right?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -206,6 +206,154 @@ def test_WINNF_FT_S_SIQ_8(self):
...
+            spectrum_inquiry_3
+        ]
+    }
+
+    # Check Spectrum Inquiry response
+    response = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse']
+    for resp_num, resp in enumerate(response):
+      self.assertEqual(resp['cbsdId'], cbsd_ids[resp_num])
+      self.assertEqual(resp['response']['responseCode'], 0)
+
```
RedaHossino: I am not sure that this additional passing by all the responses for checking those two conditions is better than adding them in the later loops.
do you agree?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -206,6 +206,154 @@ def test_WINNF_FT_S_SIQ_8(self):
...
+    self._sas_admin.InjectPalDatabaseRecord(pal_record1[0])
+    zone_id = self._sas_admin.InjectZoneData({'record': ppa_record1})
+    self.assertTrue(zone_id)
+
+    # Update PPA record with device_c's CBSD ID and Inject data
+    ppa_record2['ppaInfo']['cbsdReferenceId'] = [cbsd_ids[1]]
+    self._sas_admin.InjectPalDatabaseRecord(pal_record2[0])
+    zone_id = self._sas_admin.InjectZoneData({'record': ppa_record2})
+    self.assertTrue(zone_id)
+
```
RedaHossino: I think we need to trigger the daily activity to activate the PPA, right?

harinineelu: Reda, 
In my understanding we trigger daily activities only if the PPA is being protected. Here, we're just looking to mark the channel as "PAL", so no need to trigger.

RedaHossino: thank you Harini for your response, 
In my understanding the SAS may not consider the impact of the PAL on the spectrum channels unless it runs the IAP.
and also in the SIQ.5 we can see that it triggers the daily activities, right?

harinineelu: Reda, it was the decision of WinnForum that you need IAP ("daily activities"/"CPAS") to trigger *protection* of a PPA but that you can get a PAL channel earlier than that.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
harinineelu: Agreed. Removed this and added them to the next loop.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -206,6 +206,154 @@ def test_WINNF_FT_S_SIQ_8(self):
...
+          any(channel['channelType'] == 'PAL'
+              for channel in response[resp_num]['availableChannel']))
+      self.assertTrue(
+          all(channel['ruleApplied'] == 'FCC_PART_96'
+              for channel in response[resp_num]['availableChannel']))
+      # Verify that the upper frequency is valid (lower frequency will be
+      # checked below).
+      self.assertTrue(
+          all(channel['frequencyRange']['highFrequency'] <= 3700000000
+              for channel in response[resp_num]['availableChannel']))
```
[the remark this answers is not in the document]
harinineelu: Removed this and added them to the loop you suggested.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
harinineelu: Done.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -206,6 +206,154 @@ def test_WINNF_FT_S_SIQ_8(self):
...
+            channel['frequencyRange']['lowFrequency'] >= 3560000000)
+    for channel in response[1]['availableChannel']:
+      if channel['channelType'] == 'PAL':
+        self.assertTrue(
+            channel['frequencyRange']['lowFrequency'] == 3600000000)
+        self.assertTrue(
+            channel['frequencyRange']['highFrequency'] == 3610000000)
+      if channel['channelType'] == 'GAA':
+        self.assertTrue(
+            channel['frequencyRange']['lowFrequency'] >= 3610000000)
```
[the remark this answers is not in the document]
harinineelu: Done.

harinineelu:
Hello Reda, 

Have made all the changes that you had suggested. Could you please take a look. Thanks!

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
RedaHossino: thank you Harini , you are right so we need to update SIQ.5.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
RedaHossino: thank you very much for your modifications, they seem ok, I have one additional suggestion ,if you think it is ok, instead of the values of frequencies(3550000000, 3560000000, 3600000000, 3610000000 ) of PALs, we could use the variables that we have already pal_low_frequency1, pal_high_frequency1, pal_low_frequency2, frequency2,  right?

harinineelu: Sure, that makes sense. I have changed the way you suggested. Could you please take a look, thanks.

PR #328, opened 2017-11-03
NEW Heartbeat test cases using new IDs: 6, 8

harinineelu:
heartbeat_testcase.py → WINNF_FT_S_HBT_testcase.py
Old HBT_17 → HBT_6
Old HBT_20 → HBT_8


src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,191 @@
...
+    self.assertEqual(len(response), len(cbsd_ids))
+    grant_expire_time = []
+    for resp_number, resp in enumerate(response):
+      self.assertEqual(resp['cbsdId'], cbsd_ids[resp_number])
+      self.assertTrue(resp['grantId'])
+      self.assertEqual(resp['response']['responseCode'], 0)
+      grant_expire_time.append(
+          datetime.strptime(resp['grantExpireTime'], '%Y-%m-%dT%H:%M:%SZ'))
+
+    grant_ids = [resp['grantId'] for resp in response]
```
RedaHossino: I think we don't to do another loop here, we could use the previous one, right?

harinineelu: Changed it to use the previous loop.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,191 @@
...
+    response = self._sas.Grant(request)['grantResponse'][0]
+    self.assertEqual(response['cbsdId'], cbsd_ids[0])
+    self.assertEqual(response['response']['responseCode'], 0)
+    grant_id = response['grantId']
+    grant_expire_time = datetime.strptime(response['grantExpireTime'],
+                                          '%Y-%m-%dT%H:%M:%SZ')
+    #  Test will automatically fail if the SAS UUT is not configured to use a
+    #  short Grant duration to prevent this test from taking unnecessarily
+    #  long to execute.
+    self.assertTrue(grant_expire_time < datetime.utcnow()+timedelta(hours=24))
```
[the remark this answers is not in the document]
harinineelu: Hello Reda,

Thanks for taking the time to review. The decision to configure with shorter grant duration is documented in the testing API doc (Under Section: 'Manual Configuration of SAS under test).

Please feel free to follow-up with the test-dev group, I'll be happy to change this code to reflect any changes. For now, this should be ok to keep and approve?

harinineelu:
Hello Reda, Uploaded corrected version, could you please take a look. Thanks!

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -0,0 +1,191 @@
...
+      transmit_expire_time = datetime.strptime(resp['transmitExpireTime'],
+                                               '%Y-%m-%dT%H:%M:%SZ')
+      self.assertLess(datetime.utcnow(), transmit_expire_time)
+      self.assertLessEqual(
+          (transmit_expire_time - datetime.utcnow()).total_seconds(), 240)
+      self.assertLessEqual(transmit_expire_time,
+                           grant_expire_time[index])
+
+    # Third device is in Blacklist, must have Response Code 101.
+    self.assertEqual(response[2]['response']['responseCode'], 101)
```
RedaHossino: this line is ok according to the tests spec, but I don't know why here we don't allow response code 103, that we allow in the next step. this comment is not concerning the code of the test, but more the specification of the test.
what do you think?

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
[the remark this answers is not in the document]
RedaHossino: Hello, thank you for your response and for your changes which are ok.
in this section of admin API doc, it is written : Set grant duration to 5 minutes (exact duration TBD)
but I couldn't understand the link with the condition in the code for grant_expire_time to be less than (now + 24 hours).

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
harinineelu: Reda, 
I understand this is regarding the specification. If you have any concerns with the test spec, you should file comments by tomorrow on the test spec.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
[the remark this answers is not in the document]
harinineelu: Reda, 
Since a SAS in normal operation would set grant duration to much more than 24 hours,  here we just do a simple check to make sure that it's been reconfigured to "something reasonable" (Could be 5 minutes or 3 hours. Its still not clear and yet to be decided.)
Also, we are OK to let this assertion be lenient -- if we specify 3 hours and someone gives 3 hours + 5 minutes, we don't want to fail them for that. So (now + 24 hours) seems reasonably accommodative.

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
RedaHossino: thank you

src/harness/testcases/WINNF_FT_S_HBT_testcase.py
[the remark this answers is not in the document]
RedaHossino: thank you for your explanation, I think we can discuss this later with dev group.

PR #331, opened 2017-11-09
add support for inject CPI user

jcereced:
adding API support for inject CPI user

### DIFF
--- a/src/harness/fake_sas.py
+++ b/src/harness/fake_sas.py
@@ -260,6 +260,9 @@ class FakeSasAdmin(sas_interface.SasAdminInterface):
   def GetDailyActivitiesStatus(self):
     return {'completed': True}
 
+  def InjectCpiUser(self, request):
+    pass
+
 
 class FakeSasHandler(BaseHTTPRequestHandler):
   def _parseUrl(self, url):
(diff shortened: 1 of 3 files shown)


src/harness/sas_interface.py
```diff
@@ -513,4 +513,16 @@ def TriggerDailyActivitiesImmediatelyAndWaitUntilComplete(self):
...
+  
+  @abc.abstractmethod
+  def InjectCpiUser(self, request):
+    """SAS admin interface to add a CPI User as if it came directly from the CPI database.
+
+    Args:
+      request: A dictionary with the following key-value pairs:
+        "cpiId": (string) valid cpiId to be injected into SAS under test
+        "cpiName": (string) valid name for cpi user to be injected into SAS under test
+        "cpiPublicKey": public key value for cpi user to be injected into SAS under test
```
kate-harrison: This is also a string.

src/harness/fake_sas.py
```diff
@@ -260,6 +260,9 @@ def TriggerDailyActivitiesImmediately(self):
   def GetDailyActivitiesStatus(self):
     return {'completed': True}
 
+  def InjectCpiUser(self, request):
```
kate-harrison: Actually you just need to add something near (what is currently) line 310. We only define the functions if we need to return a value to the test harness; in the cases where we don't need to return anything, we just add to the list below.

jcereced:
ok ...hopefully this will do it...sorry we don't really use fake_sas.

src/harness/fake_sas.py
kate-harrison: You can safely remove this function since it's never called. (We only define these functions if we need to return an "interesting" response.)

jcereced:
ok let's try again....please check.  thanks

PR #332, opened 2017-11-10
Delete SIR and ESM tests files

RedaHossino:


RedaHossino:
for now we deleted the test files without deleting neither the corresponding  methods in other files nor their test data.

PR #340, opened 2017-11-21
Derive from sas_testcase instead of unittest.

harinineelu:
Derive from sas_testcase.SasTestCase instead of unittest.testcase


harinineelu:
Merged #340

PR #349, opened 2017-12-08
Fix average calculation in ITM model.

sbdt-google:
According to latest revision on R2-SGN-22.


msouryal:
I've reviewed the changes, and they look good.

msouryal:
I've reviewed the changes, and they look good.

sbdt-google:
Thanks for the quick review!

PR #350, opened 2017-12-09
Few fixes to the reference models.

sbdt-google:
Following discussion with FW, a few fixes are proposed:
 - vincenty: when both points are identical position
 - eHata: issue with FindQuantile: data access over range
 - ITM: case of reliability/confidence 0.5 does not lead to
   exact 0 value as expected.


sbdt-google:
Thanks for the review.

sbdt-google:
Restored the branch. Something happened in GitHub, and apparently a bunch of submitted PR around the 20 dec are lost. The changes although committed do not appear in the master. Strange.

Will resubmit this PR then

PR #351, opened 2017-12-11
NEW REG test cases using new IDs: 8

jcereced:
Reg test with new id #8

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_REG_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_REG_testcase.py
@@ -423,7 +423,49 @@ class RegistrationTestcase(sas_testcase.SasTestCase):
     for x in range(1, 4):
       self.assertEqual(
           response['registrationResponse'][x]['response']['responseCode'], 102)
+  
+  @winnforum_testcase
+  def test_WINNF_FT_S_REG_8(self):
+    """Invalid REG-Conditional parameters in Array Registration Request (responseCode 103)
 
+    The response should be SUCCESS for the first CBSD,
+    FAILURE 103 for the second and third CBSDs.
+    """
+
+    # Load devices
+    device_1 = json.load(open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_2 = json.load(open(os.path.join('testcases', 'testdata', 'device_c.json')))
+    device_3 = json.load(open(os.path.join('testcases', 'testdata', 'device_e.json')))
+
+    # Inject FCC IDs
+    self._sas_admin.InjectFccId({'fccId': device_1['fccId']})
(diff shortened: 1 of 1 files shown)


src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -423,7 +423,49 @@ def test_WINNF_FT_S_REG_5(self):
...
+
+    # Inject FCC IDs
+    self._sas_admin.InjectFccId({'fccId': device_1['fccId']})
+    self._sas_admin.InjectFccId({'fccId': device_2['fccId']})
+    self._sas_admin.InjectFccId({'fccId': device_3['fccId']})
 
+    # Inject User IDs
+    self._sas_admin.InjectUserId({'userId': device_1['userId']})
+    self._sas_admin.InjectUserId({'userId': device_2['userId']})
+    self._sas_admin.InjectUserId({'userId': device_3['userId']})
```
RedaHossino: it looks good, only one optional tiny comment, isn't it shorter to inject User Id and Fcc Id by a loop through the devices?

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -423,7 +423,49 @@ def test_WINNF_FT_S_REG_5(self):
...
+    # Device 3 out-of-range, or the wrong Type value for latitude.
+    device_3['installationParam']['latitude'] = 91.0
+
+    # Register the devices
+    request = {'registrationRequest': [device_1, device_2, device_3]}
+    response = self._sas.Registration(request)['registrationResponse']
+
+    # Check registration response
+    # valid cbsdId and responseCode 0 for 1st cbsd
+    self.assertTrue('cbsdId' in response[0])
```
RedaHossino: following a discussion with Kate in an email I sent to dev test list(Question about "Validation" of CbsdId), we may need to check the length in each test that asks us to verify that a valid CBSD ID is returned.
do you agree with that?

src/harness/testcases/WINNF_FT_S_REG_testcase.py
jcereced: probably....I think we were just modeling things after the way earlier TCs were impemented

RedaHossino: yes, no problem then :)

src/harness/testcases/WINNF_FT_S_REG_testcase.py
RedaHossino: I agree with Kate, maybe we cann't do more than checking the length of CBSD ID.

jcereced:
this PR is obsolete,,,created a new one to merge and avoid the conflicts...closing

PR #352, opened 2017-12-12
NEW REG test case using new ID: 13.

harinineelu:
Also includes the generated config file.

Note:
- Dependent on the configurability PR [# 348]
- util.py and sas.py are duplicates of whats in PR # 348


harinineelu:
Merged #352

PR #354, opened 2017-12-13
NEW SIQ test case using new ID: 13

gilschieber:


src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -554,3 +554,47 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    response = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse']
+
+    # Check Spectrum Inquiry Response
+    self.assertEqual(len(response), 3)
+    # First and second cbsd
+    for resp_num in range(2):
+      self.assertEqual(response[resp_num]['cbsdId'], cbsd_ids[resp_num])
+      self.assertEqual(response[resp_num]['response']['responseCode'], 0)
+    # Third cbsd (blacklisted)
+    self.assertEqual(response[2]['cbsdId'], cbsd_ids[2])
```
assaneRed: SAS doesn't send cbsdId if CBSD is blacklisted
The rest look good -- thank you!

gilschieber: You're right. This assert was included because the spec for this test stated: 
"SAS response includes the cbsdId that match the request from **each** CBSD in the array". 
But this was a spec error, it has been updated in the latest version.
I've deleted the corresponding line.

gilschieber:
Deleted the cbsdId assert for blacklisted cbsd.

PR #357, opened 2017-12-19
NEW REG test case using new ID: 7

RedaHossino:


src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -424,6 +424,185 @@ def test_WINNF_FT_S_REG_5(self):
...
+        'registrationData': [
+            conditionals_12, conditionals_13, conditionals_14
+        ]
+    }
+    self._sas_admin.PreloadRegistrationData(conditionals)
+
+    # Remove conditionals from registration
+    del device_12['cbsdCategory']
+    del device_12['airInterface']
+    del device_12['measCapability']
```
kate-harrison: I think you also need to delete installationParam here and for `device_13`.

RedaHossino: it is deleted when we convert the request to RequestWithCpiSignature, right?

kate-harrison: Oops, right you are! Sorry :)

jni2000:
### 1. the spec mentions these REG-Conditionals to be included in the registration.

    del device_14['cbsdCategory']
    del device_14['airInterface']
    del device_14['installationParam']
    del device_14['measCapability']

### 2. the spec requires the only these two parameters(indoor flag and eirp value) to be included and rest of the installation parameters to be pre-loaded into SAS. The rest of the parameters should not be included within the cpisigneddata.

    device_10['installationParam']['indoorDeployment'] = True
    device_10['installationParam']['eirpCapability'] = 31
    # Convert device_10's registration request to embed cpiSignatureData
    convertRequestToRequestWithCpiSignature(cpi_private_key, cpi_id,
                                            cpi_name, device_10)

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -424,6 +424,196 @@ def test_WINNF_FT_S_REG_5(self):
...
+    del device_10['cbsdCategory']
+    del device_10['airInterface']
+    del device_10['installationParam']
+    del device_10['measCapability']
+    del device_12['cbsdCategory']
+    del device_12['airInterface']
+    del device_12['measCapability']
+    del device_13['cbsdCategory']
+    del device_13['airInterface']
+    del device_13['measCapability']
```
harinineelu: The airInterface and cbsdCategory need to be provided outside of the CPI signed data if CPI signed data is included. So, these fields are not to be deleted here.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -424,6 +424,196 @@ def test_WINNF_FT_S_REG_5(self):
...
+                                            cpi_name, device_7)
+    device_8['installationParam']['latitude'] = 38.882162
+    device_8['installationParam']['longitude'] = -77.113755
+    device_8['installationParam']['height'] = 4.0
+    device_8['installationParam']['heightType'] = 'AGL'
+    device_8['installationParam']['indoorDeployment'] = False  
+    device_9['installationParam']['eirpCapability'] = 31
+    device_10['installationParam'] ={}
+    device_10['installationParam']['indoorDeployment'] = True
+    device_10['installationParam']['eirpCapability'] = 31
```
harinineelu: This does not include all installationParam fields for the cpiSignedData. To ensure that we meet the criteria of the test case, we could modify the original installationParam, *not* delete them on lines 542-544, and use the same installationParam when creating CPI signed data.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
RedaHossino: Thank you for this comment, you are right.
Actually we should preload only the isntallation parameters for the devices 12, 13 ,14, not all REG Conditional data.
I updated the code.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
RedaHossino: thanks, could you please read the discussion exchanged by email raised by Takashi and titled "Quick Review of PR357(REG.7)"  with test dev group.
and then we can do the modifications if needed.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
```diff
@@ -424,6 +424,165 @@ def test_WINNF_FT_S_REG_5(self):
...
+                                cpi_id, cpi_name, device_16)  
+    # Register devices
+    devices = [device_1, device_2, device_3, device_4, device_5, device_6,\
+               device_7, device_8, device_9, device_10, device_11, device_12,\
+               device_13, device_14, device_15, device_16 ]
+    request = {'registrationRequest': devices}
+    response = self._sas.Registration(request)
+    # Check registration response
+   
+    self.assertEqual(
```
jni2000: We also need to assert that a valid cbsdId is included for the first CBSD as per the test spec.

I think it would also be better (though not explicitly stated by the test spec ) to ensure that a cbsdId is included for thirteenth and not included for the rest of the failed registrations.

RedaHossino: thank you for the comment, the test spec doesn't mention doing any verification regarding the cbsd Id  for this test, and I think the SAS will fail in other REG tests if it behaves differently.
I find your proposed checks logical but actually I removed them after I had added them,
because I think it is better to be aligned with the test spec otherwise we need to add a comment in the code to tag it as unapproved part, right?

jni2000: we agree with your reasoning and we don’t need to add a check for the thirteenth CBSD. However the spec mentions that the response for the first CBSD has to be checked if it has a valid cbsdId.

jni2000: Please hold on the merge, we just noticed that the following comment has not yet be addressed yet. 

We need either do
1.	Suggest to change the spec. to remove SAS response includes a valid cbsdId for the first CBSD. 
Or do
2.	Add the check for validating the cbsdId in the response for the first CBSD

jni2000:
Please hold on the merge, we just noticed that the following comment has not yet be addressed yet. 

We need either do
1.	Suggest to change the spec. to remove SAS response includes a valid cbsdId for the first CBSD. 
Or do
2.	Add the check for validating the cbsdId in the response for the first CBSD

jni2000:
I would suggest to take option 2 to make the test codes inline with the spec. and then request a change to the test spec.

src/harness/testcases/WINNF_FT_S_REG_testcase.py
RedaHossino: thanks, yes I agree with your proposal, I added verification for "valid" Cbsd Id of device 1.
could you check it?

jni2000:
Thanks for adding the changes. The changes look good.

PR #360, opened 2017-12-21
NEW SIQ test case using new ID: 4

RedaHossino:
- test SIQ.4 which uses DPA methods in testing API.
- necessary DPA methods in sas_interface, sas and fake_sas

* the wait time for DPA activation to be discussed


src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -29,6 +29,52 @@ def setUp(self):
...
+    spectrum_inquiry_0['inquiredSpectrum'][0]['highFrequency'] = \
+                frequency_range['highFrequency']
+    request = {'spectrumInquiryRequest': [spectrum_inquiry_0]}
+    response = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse'][0] 
+    # Check spectrum inquiry response
+    self.assertEqual(response['cbsdId'], cbsd_ids[0])
+    self.assertTrue('availableChannel' in response)
+    self.assertEqual(response['response']['responseCode'], 0)
+    # Verify the low & high frequency are same as requested
+    self.assertEqual(response['availableChannel'][0]['frequencyRange']\
```
kate-harrison: Just a thought... if a SAS returned the channel availability in terms of 1 MHz or 5 MHz chunks, it would fail this test case despite having valid output.

RedaHossino: it is a good point! so we should treat the response with any number of channels.
i.e we should verify that those channels can form a frequency range(the unions of the channels) which contains the requested range and that those channels don't have any conflict, missing or repeated channels

kate-harrison: Yeah, although it's a pain to do the check :/ A helper function might be the best since we need to do something similar for SIQ.3.

Technically the text only asks us to make sure that *all* frequencies are marked available, not to check for e.g. overlapping or repeated available channels. Repeated channels might be a bug but I'm not sure overlapping channels are: suppose I know a device wants to use 15 MHz, so maybe I tell it [3550, 3565], [3555, 3570], ... with the maxEirp for each so that it can make the best decision. Another "solution" would be to give [3550, 3555], ... and let the device take the min of each 15 MHz segment. But in this case the typical [3550, 3560], ... might not be as useful for the device.

All of that to say that I don't think we need to verify that they don't overlap. If you want to add that check, let's double-check with WG3 first. You can probably also skip checking for duplicates, but WG4 may have a different opinion :)

RedaHossino: I agree with your opinion, and I updated the test to check that the channels form the requested frequency range and that there is no missing channels.

jni2000:
# 1. The DPA_ID and the frequency range has to be made configurable for
    self._sas_admin.TriggerDpaActivation({'frequencyRange':frequency_range,\
                                           'dpaId':'east_dpa4'})
    time.sleep(300)
    # Load and register CBSD
    device_a = json.load(
# 2. The device information can be placed under the specific config directory for this test case, similar to the REG_11 test case.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -30,6 +30,60 @@ def tearDown(self):
...
+    spectrum_inquiry_0['inquiredSpectrum'][0]['highFrequency'] = \
+                frequency_range['highFrequency']
+    request = {'spectrumInquiryRequest': [spectrum_inquiry_0]}
+    response = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse'][0] 
+    # Check spectrum inquiry response
+    self.assertEqual(response['cbsdId'], cbsd_ids[0])
+    self.assertTrue('availableChannel' in response)
+    self.assertEqual(response['response']['responseCode'], 0)
+    # Verify available channels contains the requested range and don't have any missing or repeated channels
+    response['availableChannel'].sort(key=lambda ch: ch['frequencyRange']\
```
kate-harrison: Cute solution! It's probably worth bundling it into a utility function that (1) can be unit tested and (2) can be used by other test cases (I think SIQ.3 has a similar check).

As I understand it, this will sort available channels by lowFreq in reverse order, so we will see the high freq channels before the low freq channels. Given that, I think you need to swap the bodies of the 'if' statements below since the first element in the sorted list will be the highest freq channel. The body of the 'else' statement also needs a slight adjustment for the same reason, as right now it will trivially pass (instead I think you want to check that highFreq[i] >= lowFreq[i-1] -- but please double-check my logic).

Even with this adjustment, I think there's a corner case where, for example, we return [ [3565, 3575], [3550, 3560], [3550, 3565] ] as available channels (already sorted in this example), the inequality assertion will fail since 3560 is not >= 3565, even though the SAS has indeed provided information for [3550, 3575] (and the SAS would have passed if it had reversed the order of the last two elements). I think this could be fixed by sorting first by high freq then by low freq (assuming that the sort function preserves relative order of "equal" elements) but I'm not sure it's worth the hassle. If you agree with my assessment, consider adding a comment recognizing this limitation (and please pull out into its own function so that our implementations are consistent).

RedaHossino: thank you Kate for this useful comment, you are right I made an error, I wanted to sort them by lowFrequency, so the reverse should be set to false, I think by default it sorts the list then by high frequency(treat frequency range as tuple) but I added high frequency to the sort to be sure.
I fixed that and I added this function to the helper functions with unit tests in a different file.
I think the only assumption now is(which is not wrong assumption) for the response: the channel with min low frequency should have lowFreq exactly equals to requested low frequency and the channel with max high frequency should have high frequency equals to the request high frequency.

kate-harrison: [quick_testcase.py.txt](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/1643870/quick_testcase.py.txt)

I did a quick test (see attached file; please excuse the ugly code and please remove the .txt extension that GitHub forced me to add) and this function seems to work as intended. We may need to add the ability to toggle the strict overlap requirement in order to use this with e.g. CBSD 2 in SIQ.3 (see the GAA channel check) but that could be done in a subsequent PR.

RedaHossino: I executed your tests and they pass, I don't know if you noticed a new file I added with the name : helper_unit_testcase.py, which does different test cases for this assert method. I thought it may be helpful to unit test some helper functions, but I think it shouldn't be executed with harness tests(product).

kate-harrison: Oh, excellent! Indeed I did not notice :) This is more than I was hoping for -- thank you! I added some small style comments but you can feel free to ignore them.

src/harness/testcases/helper_unit_testcase.py
```diff
@@ -0,0 +1,60 @@
+import sas_testcase
+import unittest
+
+class HelperFunctionUnitTest(sas_testcase.SasTestCase):
```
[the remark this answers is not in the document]
RedaHossino: Thank you for the review, all those comments are addressed in the current code.

jni2000:
looks good, please move forward.

PR #361, opened 2017-12-21
Improve Geo robustness.

sbdt-google:
Improve robustness of geo drivers.

Previously the code was not differentiating between a bad configuration with missing tiles, and expected out of bounds tiles (for which the data is set to zero).

The changes are: 
 - introduces the list of expected tiles (ned and nlcd) and will fails with an exception if an expected tile cannot be read (instead of providing wrong propagation results by taking values at 0).
 - make the geo drivers thread safe by encapsulating the change of states (related to cache) in a threading.Lock()
 - simplify the terrain code by removing code duplication (which was used for single point read).
 - share the terrain driver between the ITM and Hybrid models. Before two separate driver instance were used which increased 2x the memory usage. For best performance, one can now increase the number of cached tiles by 2x for the same memory footprint.
 - introduce a tile statistic counter which can be used to analyze the tile loading behavior. Because a good part of the reference_models  speed is related to loading tiles in and out of the cache, it is best to tune the cache size to avoid constant eviction of tiles (and also framing the client codes to work on neighboring areas if possible).
 - add a 'test_config.py' script which can be used to test the proper setup and use of the reference_models.


sbdt-google:
Thanks for the review Mo.

arhannan:
In the function def TerrainProfile(self, lat1, lon1, lat2, lon2,
                     target_res_meter=-1,
                     target_res_arcsec=1,
                     do_interp=True,
                     max_points=-1):

•	We are computing the number of grid points for the Tx and Rx as follows: 
num_points = int(dist/target_res_meter) + 1
•	After that, we are adjusting the grid resolution based on the calculated number of points (num_points)
resolution = dist / float(num_points-1)

Thus the final resolution that will be used for extracting the elevation points will be slightly greater than 30m

R2-SGN-08 states "For Initial Certification, the SAS shall choose a great circle path spacing that is as close to 30 m as possible without exceeding 30 m for R2-SGN-07 calculated distances up to 45 km for the propagation and path loss calculations."

Should we instead do num_points = ceil(dist/target_res_meter) + 1

sbdt-google:
Good catch. 
Please check the fix in the new PR #390.

PR #367, opened 2018-01-05
NEW GRA test case using new ID: 1

RedaHossino:
- Test GRA.1
- necessary code for DPA methods is duplicated with PR #357


src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,80 @@ def setUp(self):
   def tearDown(self):
     pass
 
+  @winnforum_testcase
+  def test_WINNF_FT_S_GRA_1(self):
+    """CFederal Incumbent present in the PAL frequency range requested by the CBSD
```
kate-harrison: Small typo: "Federal"

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,80 @@ def setUp(self):
...
+        'longitude'] = getRandomLatLongInPolygon(ppa_record)
+    # Register device
+    cbsd_ids = self.assertRegistered([device_a])
+    # Update PPA record with device_a's CBSD ID and Inject data
+    ppa_record['ppaInfo']['cbsdReferenceId'] = [cbsd_ids[0]]
+    self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+    zone_id = self._sas_admin.InjectZoneData({'record': ppa_record})
+    self.assertTrue(zone_id)
+    # Trigger SAS to load DPAs
+    self._sas_admin.TriggerLoadDpas()
```
kate-harrison: In most test cases (e.g. IPR.*, MCP.1) these steps come at the very beginning of the test case. If you move them to the beginning of this test case as well, SAS admins may be able to make valuable assumptions when implementing this part of the testing API. The test spec doesn't specify one way or another but I think it's assumed that it happens at the beginning.

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,80 @@ def setUp(self):
...
+    # Trigger SAS to de-active all the DPAs 
+    self._sas_admin.TriggerDpaActivation({'activate':False})
+    # Trigger SAS to active one DPA on channel c
+    self._sas_admin.TriggerDpaActivation(\
+        {'frequencyRange':{'lowFrequency': pal_low_frequency ,\
+                           'highFrequency':pal_high_frequency },'dpaId':'east_dpa4'})
+    time.sleep(300)
```
kate-harrison: Please mark as `UNAPPROVED` since this is not technically in the latest test spec (although I agree it should be).

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,80 @@ def setUp(self):
...
+    self._sas_admin.TriggerDpaActivation({'activate':False})
+    # Trigger SAS to active one DPA on channel c
+    self._sas_admin.TriggerDpaActivation(\
+        {'frequencyRange':{'lowFrequency': pal_low_frequency ,\
+                           'highFrequency':pal_high_frequency },'dpaId':'east_dpa4'})
+    time.sleep(300) 
+    # Send grant request
+    grant_0 = json.load(
+      open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    grant_0['cbsdId'] = cbsd_ids[0]
```
kate-harrison: I think you need to have the freq range match that of the PPA, otherwise it will not be affected by the DPA.

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,80 @@ def setUp(self):
...
+    request = {'grantRequest': [grant_0]}
+    response = self._sas.Grant(request)['grantResponse'][0]
+    grant_id = response[0]['grantId']
```
kate-harrison: If 'grantId' isn't in the response, this line will fail. This needs to come after line 114.

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,80 @@ def setUp(self):
...
+    # Check grant response
```
kate-harrison: An easier way to write this is:

assert(cbsdId matches)
if responseCode == 400:
  return (i.e. pass the test case)

assert(responseCode == 0)
grant_id = ...
... (remaining checks here)

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,80 @@ def setUp(self):
...
+        'cbsdId': cbsd_ids[0],
+        'grantId': grant_id,
+        'operationState': 'GRANTED'
+        }
+        del request, response
+        # Send heartbeat request
+        request = {'heartbeatRequest': heartbeat_request}
+        response = self._sas.Heartbeat(request)['heartbeatResponse'][0]
+        # Check heartbeat response
+        response['response']['responseCode'] == 501
```
kate-harrison: I think you want to assert these are equal.

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
redtechnologies: thank you Kate for your review, I fixed the code according to your very useful comments.
Could you please check it?

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
RedaHossino: thank you Kate for your review, I fixed the code according to your very useful comments.
Could you please check it?

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,85 @@ def setUp(self):
...
+    """Federal Incumbent present in the PAL frequency range requested by the CBSD
+     who is inside the DPA Neighborhood.
+     
+     grant responseCode = 0 and in heartbeat responseCode = 501(SUSPENDED_GRANT)
+     or  grant responseCode = 400
+    """
+    # Trigger SAS to load DPAs
+    self._sas_admin.TriggerLoadDpas()
+    # Trigger SAS to de-active all the DPAs 
+    self._sas_admin.TriggerDpaActivation({'activate':False})
```
kate-harrison: This should be `TriggerBulkDpaActivation`, right?

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
```diff
@@ -58,6 +58,85 @@ def setUp(self):
...
+    ppa_record['ppaInfo']['cbsdReferenceId'] = [cbsd_ids[0]]
+    self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+    zone_id = self._sas_admin.InjectZoneData({'record': ppa_record})
+    self.assertTrue(zone_id)
+    # Trigger SAS to active one DPA on channel c
+    self._sas_admin.TriggerDpaActivation(\
+        {'frequencyRange':{'lowFrequency': pal_low_frequency ,\
+                           'highFrequency':pal_high_frequency },'dpaId':'east_dpa4'})
+    # UNAPPROVED Not in WINNF-TS-0016 Release 1 Spec, but necessary Step for DPA
+    time.sleep(300)
```
kate-harrison: I think this will be 240 when we update the spec, but if you prefer you can leave as 300 since it's unapproved anyway.

src/harness/testcases/WINNF_FT_S_GRA_testcase.py
redtechnologies: yes,  you are right.
done, thank you.

RedaHossino: yes, you are right.
done, thank you.

PR #364, opened 2018-01-03
NEW SCS testcases: SCS 6 to 10

janetliuboston:

PR #371, opened 2018-01-11
Federated Wireless submitted test cases SCS.12, SCS.13, SCS.14,SCS.15

jni2000:
Test cases SAS.11 through SCS.20 completed. Please review.

Please note that test cases SCS.6 through SCS.10 currently under review is also in this pull request. We will consolidate and merge them separately when the review of SCS.6 - 10 are done and approved for checking in. So please skip the review of SCS.6 through SCS.10 in this pull request.


src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,236 @@ def test_WINNF_FT_S_SCS_5(self):
...
+    self.assertTlsHandshakeFailure(device_cert, device_key)
+
+  @winnforum_testcase 
+  def test_WINNF_FT_S_SCS_11(self):
+    """Blacklisted Certificate presented during registration.
+    Checks that SAS UUT response with fatal alert message.
+    """
+    #Blacklist certificate
+    device_a = json.load(open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    self._sas_admin.BlacklistByFccIdAndSerialNumber({'fccId': device_a['fccId'],'CbsdSerialNumber': device_a['cbsdSerialNumber']})
```
kate-harrison: We need to actually blacklist the *certificate*, not the FCC ID + SN.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,236 @@ def test_WINNF_FT_S_SCS_5(self):
...
+    self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
+    self._sas_admin.InjectUserId({'userId': device_a['userId']})
+
+    device_a_cert = os.path.join('certs', 'client_inapplicable.cert')
+    device_a_key = os.path.join('certs', 'client.key')
+    request = {'registrationRequest': [device_a]}
+    response = self._sas.Registration(request,device_a_cert,device_a_key)['registrationResponse'][0]
+
+    # Check registration response
+    self.assertTrue('cbsdId' in response)
```
kate-harrison: There should not be a CBSD ID is the registration request produced an error.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,236 @@ def test_WINNF_FT_S_SCS_5(self):
...
+  def test_WINNF_FT_S_SCS_16(self):
+    """
+    Certificate signed by a revoked CA presented during registration.
+    """
+    device_cert = os.path.join('certs', 'client.cert')
+    device_key = os.path.join('certs', 'client.key')
+    ca_cert = os.path.join('certs', 'WINNF_FT_S_SCS_16_ca.cert')
+    self.assertTlsHandshakeFailure(device_cert, device_key, ca_cert=ca_cert)
+
+  @winnforum_testcase
```
kate-harrison: Please use the pattern established with REG.11 for generating, writing, and loading config files for configurable test cases.

src/harness/util.py
```diff
@@ -277,3 +279,13 @@ def convertRequestToRequestWithCpiSignature(private_key, cpi_id,
...
   request['cpiSignatureData']['encodedCpiSignedData'] = jwt_message[1]
   request['cpiSignatureData']['digitalSignature'] = jwt_message[2]
+
+
+def countdown(t):
+  while t:
+    mins, secs = divmod(t, 60)
+    timeformat = '{:02d}:{:02d}'.format(mins, secs)
+    time.sleep(1)
+    sys.stdout.write("Waiting for client certificate to be invalidated [mm:ss] : %s   \r" % str(timeformat))
```
kate-harrison: Do you really want to print this out every second? Why not just sleep the required number of seconds?

src/harness/security_testcase.py
```diff
@@ -102,24 +103,83 @@ def _InfoCb(conn, where, ok):
...
+          logging.debug('Not After %s ', cnotAfter)
+          self.assertIsNotNone(cnotBefore)
+          logging.debug('Not Before %s ', cnotBefore)
+          logging.debug('Common Name ->subject %s ', crypto.X509Name(cert.get_subject()).commonName)
+          logging.debug('Common Name ->Issuer %s ', crypto.X509Name(cert.get_issuer()).commonName)
+          logging.debug('organizationName %s ', crypto.X509Name(cert.get_issuer()).organizationName )
+          logging.debug('ROLE  %s ', cert.get_extension(4))
+          logging.debug('CA  %s ', cert.get_extension(2))
+          self.assertEqual(cert.get_issuer().commonName, default_values["Issuer"])
+          self.assertEqual(cert.get_subject().commonName, default_values["Subject"])
```
kate-harrison: Won't this check always pass because the if() condition above requires it?

Related: if the if() condition fails, doesn't this skip all of the certificate checks? Is that really the right behavior?

norbertbarbosa:
Perhaps better to split this PR in several small PR:
- one with all the 'trivial' testcases: SCS.12, SCS.13, SCS.14, SCS.15 (which could be quickly reviewed after the pending SCS_6_to_10)
- one for SCS.17 to 19
- one for each other (11, 16,  20)

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,236 @@ def test_WINNF_FT_S_SCS_5(self):
...
+    """Certificate with inapplicable fields presented during registration. The response should be 104."""
+    
+    device_a = json.load(open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
+    self._sas_admin.InjectUserId({'userId': device_a['userId']})
+
+    device_a_cert = os.path.join('certs', 'client_inapplicable.cert')
+    device_a_key = os.path.join('certs', 'client.key')
+    request = {'registrationRequest': [device_a]}
+    response = self._sas.Registration(request,device_a_cert,device_a_key)['registrationResponse'][0]
```
norbertbarbosa: You have to specifically call     
  self.SasReset()
at the begin of this test to ensure that the SAS UUT is clean (as you are changing the SAS state).
See comment line 22

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
norbertbarbosa: In this case not sure that the 'configurable testcase' pattern is pertinent, as this test is mostly static (you don't want to generate variant of this test with different data)

norbertbarbosa: I think SCS_11 and SCS_17 to 19 require some test API extension, like a way to simulate invalidating a certificate by OCSP.
I think it was your intent in PR #373.

jni2000:
We have revised the test codes based on comments, please continue to review and approve for check in if not issues.

kate-harrison:
Just to be clear: we're not sure yet whether we will use CRL, OCSP, or both. It's currently an open question since we did not have time to discuss in the meeting yesterday. Maybe it's best to remove SCS.11 and any others that need blacklisting from this PR so that the other test cases can be submitted sooner?

jni2000:
Our suggestion is to support both in a progressive manner. At this initial stage before OCSP server is available, we use CRL and add OCSP support when the server is ready. This will allow us to have the very basic test cases done and deliverable without losing supporting new on-demand request in future.

James Ni
Director of SAS System
978.996.7480

kate-harrison:
We need to understand what is required for WinnForum Release 1 certification and support that. I am consulting with security experts internally to determine which is the better path from a security point of view. Then we probably need to have a discussion in WG2 or WG5 to make a formal decision.

(It also seems that the current code isn't using CRL anyway so we can simply work on something else until the decision is made.)

jni2000:
we did additional revision in response to the review comments, details as below

Comments | Testcase/File | Status | Remarks | Line
-- | -- | -- | -- | --
need to actually blacklist the   certificate, not the FCC ID + SN | testcases/WINNF_FT_S_SCS_testcase.py | close | (SCS_11) Not addressed | 72
There should not be a CBSD ID is the   registration request produced an error | testcases/WINNF_FT_S_SCS_testcase.py | close | (SCS_15) Removed the statement   "self.assertTrue('cbsdId' in response)" from the function   definition "test_WINNF_FT_S_SCS_15()" | 119
Please   use  the pattern established with   REG.11 for generating, writing, and loading config files for configurable   test cases     In this case not sure that the 'configurable testcase' pattern is   pertinent, as this test is mostly static | testcases/WINNF_FT_S_SCS_testcase.py | close | (SCS_16) Removed the   statement     "ca_cert = elf.getCertFilename('WINNF_FT_S_SCS_16_ca.cert')   "     As "ca_cert" is required only for the server. | 139
Do you   really want to print this out every second? Why not just sleep the required   number of seconds? | src/harness/util.py | close | Deleted the statement     sys.stdout.write("Waiting for client certificate to be invalidated   [mm:ss] : %s   \r" %   str(timeformat)) | 288
Won't   this check always pass because the if() condition above requires it?     Related: if the if() condition fails, doesn't this skip all of the   certificate checks? Is that really the right behavior? | src/harness/security_testcase.py | close | Yes,it will always pass as   "if()" condition above    requires it and will skip all the certificate checks if the   "if()" condition fails. | 142
because   it will be used only for SCS_20 it is probably better to move this specific   logic directly in SCS_20 | src/harness/security_testcase.py | close | Added a new function   "verifyServerCertificate()"      in "security_testcase.py"   file specific to SCS_20 test case.     And called this function in the definition of function "test_WINNF_FT_S_SCS_20()"   defined in "WINNF_FT_S_SCS_testcase.py " file. | 110
else set the cipher list to the default   ciphers? | src/harness/security_testcase.py | close | Default Tls objects already has   the ciphers Set in the constructor. | 253
reuse   same comment than in other testcase (like SCS_12): the spec is less verbose   on the detail (and you currently don't specifically check the alert message   content) | testcases/WINNF_FT_S_SCS_testcase.py | close | (SCS_13) Reduced the comment. | 96
probably   better to keep the default ciphers, as we want to only test a failure due to   unmanaged TLS method | testcases/WINNF_FT_S_SCS_testcase.py | close | (SCS_13)Removed the parameter   while calling the function "assertTlsHandshakeFailure()". | 96
You have   to specifically call    self.SasReset()  at the begin of   this test to ensure that the SAS UUT is clean (as you are changing the SAS   state). See comment line 22 | testcases/WINNF_FT_S_SCS_testcase.py | close | (SCS_15)Added the statement  "self.SasReset()" at the start of   the function definition "test_WINNF_FT_S_SCS_15()" . | 119
Please   use the pattern established with REG.11 for generating, writing, and loading   config files for configurable test cases | testcases/WINNF_FT_S_SCS_testcase.py | close | (SCS_16) Removed the   statement     "ca_cert = elf.getCertFilename('WINNF_FT_S_SCS_16_ca.cert')   "     As "ca_cert" is required only for the server. | 139
I think   SCS_11 and SCS_17 to 19 require some test API extension, like a way to   simulate invalidating a certificate by OCSP. | testcases/WINNF_FT_S_SCS_testcase.py | close | As discussed in the response by   James, OCSP support will be added later |  
Just to   be clear: we're not sure yet whether we will use CRL, OCSP, or both. It's   currently an open question since we did not have time to discuss in the   meeting yesterday. Maybe it's best to remove SCS.11 and any others that need   blacklisting from this PR so that the other test cases can be submitted   sooner? |   | close |   |

kate-harrison:
Can you please respond directly to the GitHub comments so that we can keep the thread of conversation? It is very difficult to read the table and to understand the responses, and it's also difficult to respond to your comments in a way that will make sense to everyone.

Also the first issue is marked closed but the notes indicate that the comment was not addressed. Which is correct?

jni2000:
Thanks, will do in future.

James Ni
Director of SAS System
978.996.7480

jni2000:
Kate, to answer your question "Also the first issue is marked closed but the notes indicate that the comment was not addressed. Which is correct?"

In our previous reply to your comments on OCSP vs CRL, we had indicated that blacklisting by OCSP will be supported in progressive manner and the support can be added into the these test cases when OCSP is ready. Hence the OCSP changes were not added but the comment was marked closed.

kate-harrison:
James, the problem is that the current code is not blacklisting by *either* method, and is thus incorrect regardless of whether we want to support CRL or OCSP.

jni2000:
Ok, my bad. We will work on it.

kate-harrison:
Can you please upload your table of responses (from 21 Jan) in another format, e.g. XLS or PDF? It is very hard to read right now because the page is too narrow to show the whole table at once.

Do you believe that all of the test cases, except SCS.11,17,18,19 (blacklisting), are ready for review at this time? I just want to make sure we're not reviewing code that you're still modifying.

If you want to submit the other test cases soon, it might help to move SCS.11,17,18,19 to a separate PR so that we can quickly approve the others. I think it may take some time to add support for blacklisting and it would be nice to submit the others ASAP.

Thanks,

Kate

jni2000:
I will split them and put them to the corresponding review comments separately to all you to follow.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: will check this with the developer and get back to you soon.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: (SCS_15)   Removed the statement "self.assertTrue('cbsdId' in response)" from   the function definition "test_WINNF_FT_S_SCS_15()"

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: (SCS_16)   Removed the statement     "ca_cert = elf.getCertFilename('WINNF_FT_S_SCS_16_ca.cert')   "     As "ca_cert" is required only for the server.

src/harness/util.py
jni2000: Deleted   the statement     sys.stdout.write("Waiting for client certificate to be invalidated   [mm:ss] : %s   \r" %   str(timeformat))

src/harness/security_testcase.py
jni2000: Yes,it   will always pass as "if()" condition above  requires it and will skip all the   certificate checks if the "if()" condition fails.

src/harness/security_testcase.py
```diff
@@ -102,24 +103,83 @@ def _InfoCb(conn, where, ok):
...
     ctx.set_info_callback(_InfoCb)
 
-    # only for potential debugging info...
-    def _VerifyCb(conn, cert, errnum, depth, ok):
-      certsubject = crypto.X509Name(cert.get_subject())
-      commonname = certsubject.commonName
-      logging.debug('TLS handshake verify: certificate: %s  -> %d', commonname, ok)
-      return ok
-    ctx.set_verify(SSL.VERIFY_PEER, _VerifyCb)
+    if IsVerifyServercert:
```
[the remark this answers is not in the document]
jni2000: Added   a new function "verifyServerCertificate()"      in "security_testcase.py"   file specific to SCS_20 test case.     And called this function in the definition of function "test_WINNF_FT_S_SCS_20()"   defined in "WINNF_FT_S_SCS_testcase.py " file.

src/harness/security_testcase.py
```diff
@@ -164,3 +224,53 @@ def doTestCipher(self, cipher, client_cert=None, client_key=None):
...
+    logging.debug('TLS handshake: connecting to: %s:%d', url.hostname,
+                  url.port or 443)
+    logging.debug('TLS handshake: privatekey_file=%s', client_key)
+    logging.debug('TLS handshake: certificate_file=%s', client_cert)
+    if (ssl_method is not None) :
+      ctx = SSL.Context(ssl_method)
+    else:
+      ctx = SSL.Context(SSL.TLSv1_2_METHOD)
+    if ciphers is not None:
+       ctx.set_cipher_list(ciphers)
```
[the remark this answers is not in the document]
jni2000: Default   Tls objects already has the ciphers Set in the constructor.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,236 @@ def test_WINNF_FT_S_SCS_5(self):
...
+    """
+    device_cert = os.path.join('certs', 'client_expired.cert')
+    device_key = os.path.join('certs', 'client_expired.key')
+    self.assertTlsHandshakeFailure(device_cert, device_key)
+
+
+  @winnforum_testcase
+  def test_WINNF_FT_S_SCS_13(self):
+    """ Disallowed TLS method attempted during registration .
+        Checks that SAS UUT response with fatal alert message below
```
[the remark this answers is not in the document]
jni2000: (SCS_13) Reduced   the comment.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,236 @@ def test_WINNF_FT_S_SCS_5(self):
...
+        The SAS UUT sends a fatal alert message with the following parameters:
+         AlertLevel = 2 (fatal) --> tlsv1 alert protocol version
+         The SAS UUT immediately terminates the TLS session
+    """
+    device_cert = os.path.join('certs', 'client.cert')
+    device_key = os.path.join('certs', 'client.key')
+    self.assertTlsHandshakeFailure(device_cert, device_key,ciphers='CAMELLIA128-SHA',\
```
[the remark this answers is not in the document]
jni2000: (SCS_13)Removed   the parameter while calling the function   "assertTlsHandshakeFailure()".

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: As we discussed, OCSP support will be added later

jni2000:
I have re-posted our responses to the review comments right follow your original review comments. Please take a look and if now more concerns, I will move forward to split the PR as you suggested. Thanks.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: We have removed the test cases SCS 11,17,18 and 19 from this PR. They will be submitted separately once the approach for blacklisting has been finalized.

src/harness/security_testcase.py
jni2000: This function is invoked only for test case SCS.20. This function was separated out based on an earlier comment. Other test cases do not require validation of each field. 

Added additional comment on top of previous resolution for better understanding

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Missed   the code change in last drop, added now

jni2000:
Dear reviewer, we have just submitted a new commit addressing all the remaining review comments, especially requesting splitting the original PR into separated ones for quick review and approval of the ready test cases and leaving the others in another PR. We included in this PR only the SCS.12, SCS.13, SCS.14, SCS.15, SCS.16 and SCS.20 test cases.  Test cases SAS.11, SAS.17-19 will submitted in a separate PR for reviewing. Please help review this PR and approve for merging if no concerns. Thanks.

src/harness/certs/generate_fake_certs.sh
```diff
@@ -150,6 +180,60 @@ openssl ca -cert unknown_ca.cert -keyfile private/unknown_ca.key -in unknown_dev
...
+openssl ca -gencrl -keyfile private/root_ca.key -cert root_ca.cert \
+     -config ../../../cert/openssl.cnf  -crlhours 1\
+     -out root/crl/root_ca.crl
+
+echo "\n\n Generate CRL for cbsd_ca"
+openssl ca -gencrl -keyfile private/cbsd_ca.key -cert cbsd_ca.cert \
+     -config ../../../cert/openssl.cnf -crlhours 1 \
+     -out root/crl/cbsd_ca.crl
+
+# Certificate for WINNF.FT.S.SCS.17,WINNF.FT.S.SCS.18 and WINNF.FT.S.SCS.19
```
kate-harrison: It might be better to create this cert in the test case itself so that it can run automatically. Right now, the tester would need to create the cert and then quickly run the specified test.

src/harness/certs/generate_fake_certs.sh
```diff
@@ -150,6 +180,60 @@ openssl ca -cert unknown_ca.cert -keyfile private/unknown_ca.key -in unknown_dev
...
+    -out short_lived_client.cert -outdir ./root \
+    -policy policy_anything -extensions cbsd_req_sign -config ../../../cert/openssl.cnf \
+    -batch -notext -create_serial -utf8 -days 1185 -md sha384 -enddate $enddate_value
+
+# Generate trusted CA bundle.
+echo "\n\nGenerate 'ca' bundle"
+cat cbsd_ca.cert sas_ca.cert root_ca.cert cbsd-ecc_ca.cert sas-ecc_ca.cert root-ecc_ca.cert > ca.cert
+
+echo "Appended crl and create new trusted chain that contains revoked CA"
+cat ca.cert root/crl/cbsd_ca.crl root/crl/root_ca.crl  > WINNF_FT_S_SCS_16_ca.cert
```
kate-harrison: How is the CA revoked?

src/harness/fake_sas.py
```diff
@@ -357,16 +358,36 @@ def do_GET(self):
...
 
-def RunFakeServer(version, is_ecc):
+def RunFakeServer(version, is_ecc, ca_cert,verify_crl):
   FakeSasHandler.SetVersion(version)
   if is_ecc:
     assert ssl.HAS_ECDH
   server = HTTPServer(('localhost', PORT), FakeSasHandler)
-  server.socket = ssl.wrap_socket(
+
+  if ca_cert is not None:
```
kate-harrison: What purpose does the new code serve? (I'm not a security expert so it's hard for me to understand it quickly.) In general, I'm inclined to keep fake SAS as simple as possible since it is *not* expected to pass the tests.

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
     self.assertTrue(finished_check_idx > 0)
     self.assertTrue(finished_check_idx > cipher_check_idx)
 
+
+
+  def assertVerifyServerCertificateSuceed(self, base_url, ciphers, client_cert, client_key):
+    """ Specific to SCS_20 testcase
```
kate-harrison: If this is really used only in SCS.20, then let's put the code in the same file. The purpose of this file is to contain helper functions that will be reused across two or more test cases.

src/harness/security_testcase.py
kate-harrison: Even if it is invoked only for a single test case, we should still make sure the logic is correct. Right now, an incorrectly-implemented SAS could pass this test case by making sure that the if() condition fails, which we definitely don't want. Please ensure that this code is correctly executing the checks required by SCS.20.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +67,83 @@ def test_WINNF_FT_S_SCS_5(self):
...
     self.doTestCipher('ECDHE-RSA-AES128-GCM-SHA256')
+
+
+
+  @winnforum_testcase
+  def test_WINNF_FT_S_SCS_12(self):
+    """Expired certificate presented during registration.
+    Checks that SAS UUT response with fatal alert message.
+    """
+    device_cert = self.getCertFilename('client_expired.cert')
```
kate-harrison: Should this be configurable?

src/harness/testcases/configs/test_WINNF_FT_S_SCS_17/default.config
```diff
@@ -0,0 +1,4 @@
+{
```
kate-harrison: I don't think this is used anymore.

src/harness/testcases/configs/test_WINNF_FT_S_SCS_20/default.config
```diff
@@ -0,0 +1,11 @@
+{
```
kate-harrison: Is this used anywhere?

src/harness/util.py
```diff
@@ -277,3 +279,14 @@ def convertRequestToRequestWithCpiSignature(private_key, cpi_id,
   request['cpiSignatureData']['protectedHeader'] = jwt_message[0]
   request['cpiSignatureData']['encodedCpiSignedData'] = jwt_message[1]
   request['cpiSignatureData']['digitalSignature'] = jwt_message[2]
+
+
+def countdown(t):
```
kate-harrison: I don't think this is used anywhere in this PR.

norbertbarbosa:
Please resolve the conflicts before, as merging in git can easily introduce new error.

By also removing the SCS_16 from this PR and keeping only SCS12, SCS_13, SCS_14, SCS_15 and SCS_20, you should end up with a more simple PR, which should only contains the following files:
- cert/openssl.cnf  (keep only the 'cbsd_req_inapplicable_sign' section)
- src/harness/certs/README.md
- src/harnesscerts/generate_fake_certs.sh (keep only used certificate)
- src/harnesstestcases/WINNF_FT_S_SCS_testcase.py (if you migrate the specific code to SCS_20 directly under this test)

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
...
+          cnotBefore = cert.get_notBefore()
+          cnotAfter = cert.get_notAfter()
+          pubkey = cert.get_pubkey()
+          serial_number = cert.get_serial_number()
+          certsubject = crypto.X509Name(cert.get_issuer())
+          commonname = certsubject.commonName
+          # only for potential debugging info...
+          logging.debug('serial number %s ', cert.get_serial_number())
+          logging.debug('signature algorithm %s ', cert.get_signature_algorithm())
+          logging.debug('Number of bits in Publick key  %s ', pubkey.bits())
```
norbertbarbosa: no check on pubkey?

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
...
+        certsubject = crypto.X509Name(cert.get_subject())
+        commonname = certsubject.commonName
+        logging.debug('TLS handshake verify: certificate: %s  -> %d', commonname, ok)
+        if(default_values["Subject"] == crypto.X509Name(cert.get_subject()).commonName ):
+          cnotBefore = cert.get_notBefore()
+          cnotAfter = cert.get_notAfter()
+          pubkey = cert.get_pubkey()
+          serial_number = cert.get_serial_number()
+          certsubject = crypto.X509Name(cert.get_issuer())
+          commonname = certsubject.commonName
```
norbertbarbosa: duplicate line 202

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
...
+          # only for potential debugging info...
+          logging.debug('serial number %s ', cert.get_serial_number())
```
norbertbarbosa: no check on serial_number?

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
...
+          self.assertIn("1.3.6.1.4.1.46609.1.6",policy_sas, 0)
+          policy_ca = str(cert.get_extension(2))
+          self.assertIn("CA:FALSE",policy_ca)
+          self.assertNotIn("pathlen",policy_ca, 0)
+          logging.debug('ROLE SAS  %s ',policy_sas)
+          signature_alogo = cert.get_signature_algorithm()
+          signature_alog_list = ["sha256WithRSAEncryption","sha384WithRSAEncryption","sha512WithRSAEncryption","ecdsa-with-Sha256","ecdsa-with-Sha384","ecdsa-with-Sha512"]
+          self.assertIn(signature_alogo,signature_alog_list)
+          logging.info("All fields in certificates are validated successfully")
+        return ok
```
norbertbarbosa: Basically I think that this function need more work to match the specification:
```
The SAS UUT response must satisfy all of the following conditions:
- The SAS UUT provided certificate has the following valid parameters:
  - Issuer matches a WInnForum approved SAS Provider CA
  - Subject is consistent with the SAS UUT
  - Version = 2 (v3)
  - Serial Number is a positive integer no longer than 20 octets
  - Algorithm ID
  - Validity
     - Not Before is encoded as UTCTime if before the year 2050, or GeneralizedTime for dates in 2050 or later
     - Not After is encoded as UTCTime if before the year 2050, or GeneralizedTime for dates in 2050 or later
  - Subject Public Key Info
     - Public Key Algorithm = SHA-256, SHA-384, or SHA-512
     - Subject Public Key ≥ 2048 bits
  - Subject: Common Name (CN)
  - CA: FALSE
  - Path Length Constraint = None
  - ROLE = SAS
  - FRN = SAS UUT administrator’s assigned FRN
  - Signed by a valid CBRS root of trust
- The SAS UUT provided certificate does not contain the following custom SAS OIDs:
  - ZONE
  - FREQUENCY
  - FCCID
  - SERIAL
  - CPIRID
If any of the above conditions are not met, the SAS UUT FAILS this test. Otherwise, it PASSES.
```

jni2000: This comment will be addressed as part of   the SCS20 test case submission in a new PR.

src/harness/security_testcase.py
jni2000: This comment will be addressed as part of   the SCS20 test case submission in a new PR.

src/harness/security_testcase.py
jni2000: This comment will be addressed as part of   the SCS20 test case submission in a new PR.

src/harness/security_testcase.py
jni2000: This comment will be addressed as part of   the SCS20 test case submission in a new PR.

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
...
+          logging.debug('ROLE  %s ', cert.get_extension(4))
+          logging.debug('CA  %s ', cert.get_extension(2))
+          self.assertEqual(cert.get_issuer().commonName, default_values["Issuer"])
+          self.assertEqual(cert.get_subject().commonName, default_values["Subject"])
+          self.assertEqual(cert.get_subject().countryName , default_values["countryName"])
+          self.assertEqual(cert.get_subject().stateOrProvinceName  , default_values["stateOrProvinceName"])
+          self.assertEqual(cert.get_subject().localityName   , default_values["localityName"])
+          self.assertEqual(cert.get_subject().organizationName   , default_values["organizationName"])
+          self.assertEqual(cert.get_subject().organizationalUnitName    , default_values["organizationalUnitName"])
+          self.assertEqual(cert.get_subject().commonName   , default_values["commonName"])
```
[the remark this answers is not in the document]
jni2000: This comment will be addressed as part of   the SCS20 test case submission in a new PR.

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
...
+
+    client_ssl_informations = []
+    def _InfoCb(conn, where, ok):
+      client_ssl_informations.append(conn.get_state_string())
+      logging.debug('TLS handshake info: %d|%d %s', where, ok, conn.get_state_string())
+      return ok
+    ctx.set_info_callback(_InfoCb)
+
+    
+    default_values = json.load(
```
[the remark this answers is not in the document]
jni2000: Yes.   We will make the change as suggested. This comment will be addressed as part   of the SCS20 test case submission in a new PR.

src/harness/security_testcase.py
```diff
@@ -140,6 +154,146 @@ def findIndexMatching(array, regex):
...
+                    open(os.path.join('testcases', 'configs', 'test_WINNF_FT_S_SCS_20', 'default.config')))
+
+    def _ValidateSeverCertificateCb(conn, cert, errnum, depth, ok):
+        certsubject = crypto.X509Name(cert.get_subject())
+        commonname = certsubject.commonName
+        logging.debug('TLS handshake verify: certificate: %s  -> %d', commonname, ok)
+        if(default_values["Subject"] == crypto.X509Name(cert.get_subject()).commonName ):
```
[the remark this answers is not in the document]
jni2000: This comment will be addressed as part of   the SCS20 test case submission in a new PR.

src/harness/util.py
jni2000: This   file has been removed as the associated test case has been removed.

src/harness/testcases/configs/test_WINNF_FT_S_SCS_20/default.config
jni2000: This   file has been removed as the associated test case has been removed.

src/harness/testcases/configs/test_WINNF_FT_S_SCS_17/default.config
jni2000: This   file has been removed as the associated test case has been removed.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Yes.   It is good to have these negative test cases as configurable, especially the   ones that use a certificate that is different from the default 'client.cert'.   This will allow different certificate to be used during ITS testing without   modifying the test case.      The cert path are configurable now and added into      default.config for tests SCS12 and SCS15.          Please note that the only the client.cert and client.key are made   configurable. The CA Cert file will be the default one that is used in the   sas.py file. If a different root of trust has to be used then this file has   to be changed. This is however applicable to the existing positive test cases   as well (SCS1 to 5).

src/harness/fake_sas.py
jni2000: Checks the validity of all certificates in the chain by   attempting to look up valid CRLs.           This change has been removed as the associated test case has been removed.

src/harness/security_testcase.py
jni2000: We   agree and the    "assertVerifyServerCertificateSuceed()" method will be moved   to "test_WINNF_FT_S_SCS_20()" function of   "WINNF_FT_S_SCS_testcase.py". This change will be submitted as part   of the new PR for SCS.20

src/harness/certs/generate_fake_certs.sh
jni2000: Currently   the CBSD intermediate CA is revoked using the root CA and the revoked list is   generated. However, this change has been removed as the associated test case   has been removed.     This approach will be changed when a CRL server is implemented as part of   SCS16, 17, 18 and 19.

src/harness/certs/generate_fake_certs.sh
jni2000: The test   cases related to this comment has been removed. We will make it a   configurable test case when it gets submitted.

jni2000:
Per your suggestion. We have already removed the test cases SCS.11, 17, 18 and 19 from this PR. We also removed SCS.16 and SCS.20 from this PR and keep this PR simple with just SCS.12, SCS.13, SCS.14, SCS.15. The test SCS.16 also depends on CRL and that will we worked after the CRL approach is finalized in WG4. The test case SCS.20 is a bigger and it will be separated into a different PR. Also, we have made all necessary fix and corrections besed on the review comments on teh rest SCS.12,13,14 and 15. Please help review again and move forward if no more comments.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,75 @@ def test_WINNF_FT_S_SCS_5(self):
...
+    writeConfig(filename, config)
+  
+  @configurable_testcase(generate_SCS_12_default_config)
+  def test_WINNF_FT_S_SCS_12(self,config_filename):
+    """Expired certificate presented during registration.
+
+    Checks that SAS UUT response with fatal alert message.
+    """
+    config = loadConfig(config_filename)
+    self.assertTlsHandshakeFailure(client_cert=config['client_cert_path'],
```
norbertbarbosa: The keys don't match those used in generate_SCS_12_default_config(). should be 'clientCert' and 'clientKey'

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,75 @@ def test_WINNF_FT_S_SCS_5(self):
...
+  def test_WINNF_FT_S_SCS_15(self,config_filename):
+    """Certificate with inapplicable fields presented during registration. 
+    
+    Checks that SAS UUT response with tls connection succedded and response should be 104.    
+    """
+    self.SasReset()
+    config = loadConfig(config_filename)
+    
+    # load device certs and device_a from config file
+    device_a_cert = config['client_cert_path']
```
norbertbarbosa: Same: keys don't match those defined in generate_SCS_15_default_config()

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Thanks   for catching this. The keys have been updated correctly.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Thanks   for catching this. The keys have been updated correctly.

jni2000:
norbertbarbosa   on Jan 13      else set the cipher list to the default ciphers?         
jni2000 10 days ago     Default Tls objects already has the ciphers Set in the constructor.

src/harness/security_testcase.py


We have   missed this comment in the last submission. Updated it correctly now.     if ciphers argument is not passed to the method   "assertTlsHandshakeFailure" then     SSLContext.set_cipher_list is set to one of the cipher loaded by default in   sas._tls      object

jni2000:
Norbert

We have addressed your new comments on key sim-matching issue as well as the "default ciphers" issue. Please review and move on,

Thanks.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -64,3 +66,75 @@ def test_WINNF_FT_S_SCS_5(self):
...
+    
+    # load device certs and device_a from config file
+    device_a_cert = config['clientCert']
+    device_a_key = config['clientKey']
+    device_a = json.load(open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    
+    # Inject fccId and userId
+    self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
+    self._sas_admin.InjectUserId({'userId': device_a['userId']})
+    self.assertTlsHandshakeFailure(client_cert=device_a_cert,client_key=device_a_key)
```
norbertbarbosa: from the spec:
```
2.  CBSD Test Harness attempts to establish a TLS handshake with the SAS UUT using a valid ciphersuite
     CHECK: The SAS UUT response must satisfy all of the following conditions:
        - A valid TLS session is established
3.  CBSD Test Harness sends a valid registration request to the SAS UUT
     CHECK: ...response=104 ...
```

you should use self.assertTlsHandshakeSucceed and not check for a failure.
Also to have the handshake check occuring at first, I will be better to do:

```
def test_WINNF_FT_S_SCS_15(self,config_filename):
     """Certificate with inapplicable fields presented during registration. 
     
     Checks that SAS UUT response with TLS connection succeeded and response should be 104.    
     """
     
     # load the keys/certs and check that TLS handshake is valid
     device_a_cert = config['clientCert']
     device_a_key = config['clientKey']
     self.assertTlsHandshakeSucceed(self._sas_admin._base_url, ['AES128-GCM-SHA256'], device_a_cert, device_a_key)
     
     # Load device and inject fccId and userId
     device_a = json.load(open(os.path.join('testcases', 'testdata', 'device_a.json')))
     self.SasReset()
     self._sas_admin.InjectFccId({'fccId': device_a['fccId']})
     self._sas_admin.InjectUserId({'userId': device_a['userId']})
  
     # Send registration Request with certs(inapplicable fields) to SAS UUT 
     request = {'registrationRequest': [device_a]}
     response = self._sas.Registration(request, ssl_cert=device_a_cert, ssl_key=device_a_key)['registrationResponse'][0]
 
     # Check registration response
     self.assertEqual(response['response']['responseCode'],104) 
```

jni2000: Norbert, thanks for the valuable inputs. Fixed now.

jni2000:
Norbert, we have made necessary changes according to your suggestion and correction. Please review. If no more issues, please approve the PR to allow us moving on to other related test cases like SDS and SSS to meet our end of month deliver goal. Thanks.

jni2000:
We understand that the reviewer mentioned not to commit the file src/harness/security_testcase.py
But there are some changes related to SCS12,13,14 and 15 tests need to be included:
1.	Added the method getCertFilename.
2.	Changed the mandatory arguments client_cert and client_key as optional in assertTlsHandshakeFailure Method.
3.	Added new optional arguments ssl_method and ciphers in assertTlsHandshakeFailure Method.
4.	Removed ca_cert optional argument is no longer required/used for SCS tests.

norbertbarbosa:
You are correct for the security_testcase.py: this PR added new change on `assertTlsHandshakeFailure()` that are not in PR #364, so it have to be included.
Thanks to notice it.

jni2000:
Thanks, Norbert. We now have SDS.6-10, 12-15 and SSS.6-10, 12-15 waiting for your review and inputs. This will greatly help us on reaching the end of Feb delivery goal.

PR #373, opened 2018-01-11
Fw sds6to19

jni2000:
SDS.6 -- SDS.19. Federated Wireless


src/harness/sas_interface.py
```diff
@@ -230,6 +230,17 @@ def BlacklistByFccId(self, request):
     """
     pass
 
+  @abc.abstractmethod
+  def BlacklistDPByFRN(self, request):
```
kate-harrison: I don't think this is how cert blacklisting will work in normal operation. If it is, can you please provide some information that I can relay to the rest of test dev to explain this addition to the testing API?

My understanding is that we will use OCSP in normal operation and that the certs need to be blacklisted in that way.

kate-harrison:
Since this PR shares a lot (both in code and in concept) with PR #371, it may make sense to review and submit that PR before reviewing this one.

jni2000:
agreed. Let's finish all SCS and finalize them and then we can consolidate all in here.

jni2000:
The pull request captured duplicated codes for TC SCS.11 -- SCS.20. Close this one and will create a new clean one. All the review comments on this thread were or will be addressed appropriately.

jni2000:
!!! please do not delete the branch (fwSDS6to19) !!!

PR #374, opened 2018-01-11
Few fixes to the reference models.

sbdt-google:
Following discussion with FW, a few fixes are proposed:
 - vincenty: when both points are identical position
 - eHata: issue with FindQuantile: data access over range
 - ITM: case of reliability/confidence 0.5 does not lead to
   exact 0 value as expected.

NOTE: this PR was already entirely processed in #350 , but a GitHub issue occured around Dec 20, and a bunch of commit disappeared.

### DIFF
--- a/src/harness/reference_models/geo/vincenty.py
+++ b/src/harness/reference_models/geo/vincenty.py
@@ -50,6 +50,8 @@ def GeodesicDistanceBearing(lat1, lon1, lat2, lon2, accuracy=1.0E-12):
   Returns:
     a tuple of distance (km), initial bearing (deg), and back bearing (deg).
   """
+  if lat1 == lat2 and lon1 == lon2:
+    return 0., 0., 0.
 
   a = 6378.1370        # semi-major axis (km), WGS84
   f = 1./298.257223563 # flattening of the ellipsoid, WGS84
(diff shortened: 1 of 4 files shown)


arhannan:
Looking at ehata.CbsdEffectiveHeights
should
     avg_height = sum(its_elev[i_start:i_end+1]) / float(i_end - i_start + 1)
be 
     avg_height = sum(its_elev[i_start:i_end]) / float(i_end - i_start + 1)

arhannan:
Serge,

Thanks for the clarification. I deleted the comment once I recalled that python slice indexing doesn't include the last element.

---------------
Hi Ariful,

I recieved that, but I cannot see the comment in the discussion of the related PR (https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/374).
Perhaps you have deleted the comment after doing it ?

It seems the original line would be correct. In python it goes to the right element -1. So in this case,
it goes from i_start to i_end included. So you have to divide by the total number of points, ie (i_end-i_start + 1).

If your comment is about reducing the right hand side to go to (i_end-1) included only:  
please note that this function reproduces the logic in: FindAverageGroundHeight() (CBSD side only) in file PreprocessTerrainPath()

Thanks,
Serge

sbdt-google:
Thanks for you answer, that makes sense.

Issue #333, opened 2017-11-10
Some Suggested Prop Model Tweaks


sjm-federated:
[PropModelComments.docx](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/1462978/PropModelComments.docx)

sbdt-google:
For sake of keeping track on things.

[PropModelComments_google-answer.docx.pdf](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/1583682/PropModelComments_google-answer.docx.pdf)

sbdt-google:
Additional report from FW, after stress testing on branch #376 


[LastRefCodeBranch_Comments.docx](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/1669460/LastRefCodeBranch_Comments.docx)

sbdt-google:
New report from FW about compiler related issues in model
[FineRollingHillyCompilerIssue.docx](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/1865449/FineRollingHillyCompilerIssue.docx)

sbdt-google:
I reopen this issue regarding an issue with the NTIA E-Hata reference implementation.
It has been put to the attention to NTIA/ITS by email and on[ this issue of the NTIA E-Hata repository](https://github.com/NTIA/ehata/issues/7).
My proposal is to put the condition to avoid applying fine correction if Delta_h < 10m.

sbdt-google:
Reopend

PR #376, opened 2018-01-12
E-Hata fixes for robustness

sbdt-google:
Two fixes:
 - Improve the path profile distance to recover floating point accuracy
   loss (caused by the profile format where the distance is obtained
   by a multiplication N.step, instead of being encoded). This may
   cause mismatch on various compilers, because of the calculation
   non-linearities dependent on specific 'plain' distance (10km...).

 - use internal dynamic array instead of hardcoded one, which causes
   crash if step resolution < 25m

NOTE: related to FW report: #333

### DIFF
--- a/src/harness/reference_models/propagation/ehata/its/ExtendedHata.cpp
+++ b/src/harness/reference_models/propagation/ehata/its/ExtendedHata.cpp
@@ -1,4 +1,5 @@
 #include "stdio.h"
+#include "cmath"
 #include "ehata.h"
 
 // ******* WinnForum extension *******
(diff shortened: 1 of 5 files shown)


kate-harrison:
Is this PR being actively reviewed?

sbdt-google:
Yes. Mo is doing some stress test - discussion done by email. Will update shortly with some new stuff.

jni2000:
Yes, I asked Mo and he agreed to review it, Will follow up.

sbdt-google:
I just added the suggested correction by Mo. Well found !!
This is really a small bug of the e-hata model. For reference, I uploaded your analysis document in the ticket #333

sbdt-google:
Thanks for the review and running your extensive stress test.
I will submit knowing that there another bunch of small changes coming to the Hybrid/E-Hata to fix other things discovered recently.

arhannan:
In the percentile computations pfl10__meter, pfl50__meter, pfl90__meter in computing the RollingHill correction, the lower ordinal rank was used. If bilinear interpolation is used, the results are different. FineRollingHillyTerrainCorectionFactor is very sensitive to the values of h_m_gnd__meter, pfl10__meter, pfl50__meter, pfl90__meter. We have seen large dB variations between both methods. pfl10__meter is the 10th percentile is a sorted (descending) order
e.g. 
Tx(40.7305, -82.7039) and Rx(40.6989, -82.3456), the correction is Krh is ~-19.85dB for the nearest ordinal rank and ~-8dB with bilinear interpolation
Tx(33.7395, -99.5371) and Rx(34.4095, -99.2788), the correction is Krh is ~-13.68dB for the nearest ordinal rank and ~-4.47dB with bilinear interpolation

PR #388, opened 2018-01-17
Federated Wireless added test cases FDB

jni2000:
Federated wireless submitted test cases FDB.1 -- FDB.8. please review.


jni2000:
A couple of high-level comments:

    There seem to be some unrelated changes to e.g. fake_sas.py. Is this branched from master?

The changes are   reverted.
--
    Please break this PR into smaller PRs so that it's easier to review. I suggest getting one of the FDB test cases approved and then submitting the rest for review (see comment 3).

Made   changes in code base only for FDB_1 test case to test and review.
--
    These test cases require inputting (via the config) the DB contents as well as spawning a process (a "fake database") that will serve those files when the SAS UUT requests them from the test harness. This is similar to the SAS Test Harness which also provides files to the SAS UUT. Thus the config file should include snapshots of the database (which are then hosted as part of the test case code) and should not include a database URL (it will be determined by the test code and there's no need for the tester to be able to specify). Note that GitHub is an untenable solution for hosting these files as (a) it is public and (b) it prevents parallel execution of test cases (among other things). Also some DBs -- e.g. FSS -- provide an API for SASes and we need to determine how much of that API needs to be supported by the "fake database".

Changes   made are          in "test_WINNF_FT_S_FDB_1()" of    "WINNF_FT_S_FDB_testcase.py"          in "exclusion_zone_db['url'] "    field  of  "default.config" for reading the   database file using REST API.          in "util.py" added new method "loadDB()" to read a fake   database file.          Added new files "fake_db_server.py" and   "exclusion_zone_db.json" for reading the fake database using  REST API.           in "README.md" for fake database server and exclusion zone   data.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py          src/harness/testcases/configs/test_WINNF_FT_S_FDB_1/default.config          src/harness/util.py          src/harness/fake_db_server.py     src/harness/fakedatabase/exclusion_zone_db.json          src/harness/README.md
--

jni2000:
We are doing thorough internal review on this submission right now, will have update soon in one or two days. Please hold the review for now till will have the next updates. Thanks.

jni2000:
we have responded to the review comments 5 days before and make the changes, please move on to review them. Thanks.

src/harness/fake_db_server.py
```diff
@@ -0,0 +1,78 @@
...
+import json
+import threading
+from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+
+class FakeDatabaseTestHarness(threading.Thread):
+
+  def __init__(self, url = None, database_file = None):
+    """Test Harness acting as a Http Server to receive Pull/GET
+    Requests from SAS Under Test
+    parameters: is the dict variable for database.
```
kate-harrison: As commented in other PRs, please pass parameters directly rather than using a dictionary. This will help us avoid errors and it makes the code much more readable.

src/harness/fake_db_server.py
```diff
@@ -0,0 +1,78 @@
...
+    schema of parameters: {
+    "path": path to the database file
+    "url": actual url   
+    "database_file" : actual database file
+    "port" : port number at which fake database server will run
+    "host_name": entity hosting the fake database server
+    """
+    super(FakeDatabaseTestHarness, self).__init__()
+    self.parameters = {
+        "path" : ''.join( ('/', url.split('/',3)[3]) ) ,
```
kate-harrison: What is going on here?

src/harness/fake_db_server.py
```diff
@@ -0,0 +1,78 @@
...
+    parameters: is the dict variable for database.
+    schema of parameters: {
+    "path": path to the database file
+    "url": actual url   
+    "database_file" : actual database file
+    "port" : port number at which fake database server will run
+    "host_name": entity hosting the fake database server
+    """
+    super(FakeDatabaseTestHarness, self).__init__()
+    self.parameters = {
```
kate-harrison: Please save as individual vars rather than a dictionary. There is no value to using a dict.

src/harness/fake_db_server.py
```diff
@@ -0,0 +1,78 @@
...
+from BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
+
+class FakeDatabaseTestHarness(threading.Thread):
+
+  def __init__(self, url = None, database_file = None):
+    """Test Harness acting as a Http Server to receive Pull/GET
+    Requests from SAS Under Test
+    parameters: is the dict variable for database.
+    schema of parameters: {
+    "path": path to the database file
```
kate-harrison: Will we always know the path in advance? (I can't think of any obvious exceptions, but I'm not an expert in this area).

Will there always be only one file? I think for the ULS database, there are multiple files that need to be combined -- but again, this is not my area of expertise.

You may need a different DB class for each type of DB although I'm not sure at this time what the best approach is.

src/harness/fake_db_server.py
```diff
@@ -0,0 +1,78 @@
...
+    "url": actual url   
+    "database_file" : actual database file
+    "port" : port number at which fake database server will run
+    "host_name": entity hosting the fake database server
```
kate-harrison: Why is this needed?

src/harness/fakedatabase/exclusion_zone_db.json
```diff
@@ -0,0 +1,52 @@
+{
+  "terminated": false,
```
kate-harrison: This is not the format of the exclusion zone database.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
-        'fccIds': [(device_c['fccId'], 47)],
-        'userIds': [device_c['userId']],
-        'registrationRequests': [device_c],
-        'grantRequests': [grant_0],
-        'expectedResponseCodes': [0],
-        'expectedGrantResponseCodes': [0],
-        'expectedHBResponseCodes':[500],
-        'fss_db':fss_db
-    }
+        'registrationRequest': [device_a],
```
kate-harrison: Please fix the variable names to match well-established conventions.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
+    device_a['installationParam']['latitude'], device_a['installationParam']['longitude'] = getRandomLatLongInPolygon(data)
+    device_a['installationParam']['latitude'] = round(device_a['installationParam']['latitude'],6)
+    device_a['installationParam']['longitude'] = round(device_a['installationParam']['longitude'],6)
+
+    #Frequency range of the grant is adjusted to partially or fully overlap the Exclusion zone frequency range. 
+    grant_0['operationParam']['operationFrequencyRange']['lowFrequency'] =  data['zone']['features'][0]['properties']['freqrange']['lowFrequency']
+    grant_0['operationParam']['operationFrequencyRange']['highFrequency'] = data['zone']['features'][0]['properties']['freqrange']['highFrequency']
+
+    #Create the actual config.
     config = {
```
kate-harrison: We need to be able to inject Cat B CBSDs as well.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
-
-    dpa_movelist_db = json.load(open(os.path.join('testcases', 'testdata', 'dpa_db.json')))
-
-    # Create the actual config.
+    #Frequency range 'F2' to modify the exclusion zone database
+    frequency_f2_exz = { 'lowFrequency' : 3680000000,
+                        'highFrequency' : 3690000000 
+                       }
+    #Dictionary containing database information  
+    fake_fcc_db = { 'name': 'Fake_FCC_DB',
```
kate-harrison: This is not an FCC-hosted database.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
+                    'url': 'http://localhost:9090/fakedatabase/exclusionZone',
+                    'databaseFile': 'fakedatabase/exclusion_zone_db.json'
+                  }
+    #Reading the Exclusion zone sample record 
+    data = json.load(
+      open(os.path.join('testcases', 'testdata', 'exz_record_0.json')))
+    #Creating Exclusion zone database
+    writeDB(fake_fcc_db['databaseFile'], data)
```
kate-harrison: The DB file should be part of the config, not saved to disk during config generation. (It is OK to save to disk during test case execution, but the files should be cleaned up afterward.)

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
+
+    #Getting a random point within the polygon 
+    device_a['installationParam']['latitude'], device_a['installationParam']['longitude'] = getRandomLatLongInPolygon(data)
+    device_a['installationParam']['latitude'] = round(device_a['installationParam']['latitude'],6)
```
kate-harrison: Why?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
-    # Check grant response
-    self.assertEqual(responses[0]['cbsdId'], cbsd_ids[0])
-    grant_id = responses[0]['grantId']
-    grant_expire_time = datetime.strptime(responses[0]['grantExpireTime'],
-                                          '%Y-%m-%dT%H:%M:%SZ')
-    del request, responses
-    # Load exclusion zone database which includes a zone that contains the CBSD location or is within 50 meters of the CBSD location
-    self._sas_admin.InjectDatabase_url(config['dpa_movelist_db'])
+    #Create fake database server
+    self._fake_database_server = FakeDatabaseTestHarness( \
```
kate-harrison: As noted in other PRs, please do not modify the test class itself.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
+      config['fake_fcc_db']['url'], config['fake_fcc_db']['databaseFile'] )
+    #Start fake database server
+    self._fake_database_server.start()
+
+    #Inject FCC ID and User Id of the device into UUT.    
+    fcc_id = (config['registrationRequest'][0]['fccId'], 47)
```
kate-harrison: You need to do this for all, not just the first.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
kate-harrison: We need to support multiple registrations and grants. Please double-check the test spec.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
+   
+    #Loading the FCC database URL into the UUT, the database has a zone containing the CBSD location or is within 50 meters of the CBSD location
+    self._sas_admin.InjectDatabaseUrl(config['fake_fcc_db']['url'])
 
     # Trigger daily activities
     self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
 
-    # Response code 500(TERMINATED_GRANT)or 501 (SUSPENDED_GRANT) should be sent for Heartbeat Request
+    #Construct heartbeat message
     request = {
```
kate-harrison: You need to handle multiple CBSDs+Grants (here and below).

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
-    # Response code 500(TERMINATED_GRANT) should be sent for Heartbeat Request
-    request = {
-        'heartbeatRequest': [{
-            'cbsdId': cbsd_ids[0], 'grantId': grant_id, 'operationState': 'GRANTED'
-        }]
-    }
-    response = self._sas.Heartbeat(request)['heartbeatResponse'][0]
-    # Check the heartbeat response
+    #Check the heartbeat response code is 500(TERMINATED_GRANT) or 501 (SUSPENDED_GRANT) 
     self.assertEqual(response['cbsdId'], cbsd_ids[0])
```
kate-harrison: This is not requested by the test spec.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
-    del request, response
-
-    # Load FSS database which includes at least one FSS site near the above registered CBSD location
-    self._sas_admin.InjectDatabase_url(config['fss_db'])
+            'grantId': grant_ids[0]
+          }]
+        }
+      response = self._sas.Relinquishment(request)['relinquishmentResponse'][0]
+      # Check the relinquishment response
+      self.assertEqual(response['cbsdId'], cbsd_ids[0])
```
kate-harrison: Not in the spec.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
+      self.assertEqual(response['grantId'], grant_ids[0])
+      self.assertIn(response['response']['responseCode'], [400])
```
kate-harrison: Use assertEqual

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
+
+    #Request grant with frequency range which partially or fully overlaps with \
+    #Exclusion zone protected frequency range 'F1'
+    config['grantRequest'][0]['cbsdId'] = cbsd_ids[0]
+    #Get an overlapping frequency range for the Exclusion zone Frequency range(F1)
+    freq_range = getOverlappingFrequency(\
```
kate-harrison: This should be taken from the config file, not hard-coded into the test case.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
-    config['grantRequests'][0]['operationParam']['operationFrequencyRange']['highFrequency'] = 3680000000
-    request = {'grantRequest': config['grantRequests']}
-
-    # Check grant response should be 400(INTERFERENCE)
+    config['grantRequest'][0]['operationParam']['operationFrequencyRange']['lowFrequency'] = freq_range[0]
+    config['grantRequest'][0]['operationParam']['operationFrequencyRange']['highFrequency'] = freq_range[1]
+    request = {'grantRequest': config['grantRequest']}
+    #Check grant response should be 400 (INTERFERENCE).
     response = self._sas.Grant(request)['grantResponse'][0]
     self.assertEqual(response['cbsdId'], cbsd_ids[0])
```
kate-harrison: Same comments here.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -51,993 +34,148 @@ def tearDown(self):
...
-    self.assertIn(response['response']['responseCode'],config['expectedGrantResponseCodes'])
-    grant_id = response['grantId']
-    grant_expire_time = datetime.strptime(response['grantExpireTime'],
-                                          '%Y-%m-%dT%H:%M:%SZ')
+    self.assertIn(response['response']['responseCode'], [400])
     del request, response
 
+    #Load modified exclusion zone database(change frequency of protected zone 'F2')
+    #Reading the fake database file.
+    with open(config['fake_fcc_db']['databaseFile'], 'r') as f:
```
kate-harrison: The modification should be done in the config file, not hard-coded into the test case.

src/harness/fake_db_server.py
```diff
@@ -0,0 +1,78 @@
+#    Copyright 2016 SAS Project Authors. All Rights Reserved.
```
kate-harrison: 2018

src/harness/fake_db_server.py
```diff
@@ -0,0 +1,78 @@
...
+    logging.info('Started Test Harness Server')
+    
+
+class FakeDatabaseTestHarnessHandler(BaseHTTPRequestHandler, FakeDatabaseTestHarness):
+  
+  _parameters = None
+
+  def do_POST(self):
+    """Handles POST requests"""
+    pass
```
kate-harrison: This should probably throw an error unless we expect to receive POST requests.

src/harness/fake_db_server.py
jni2000: Passing   the parameters( port, host_name, path, databaseFile) as individual variables   to fake database server.

src/harness/fake_db_server.py
jni2000: The   parsing of URL is now in "testcases/WINNF_FT_S_FDB_testcase.py" in   order to send the GET request to fake database server

src/harness/fake_db_server.py
jni2000: Individual   variables are created in "testcases/WINNF_FT_S_FDB_testcase.py" for   starting the fake database server.

src/harness/fake_db_server.py
jni2000: Yes,   the path is known as the testcase creates exclusion zone database          Yes, single file is created for fake DB implementation          Different types of data are dumped into differrent directory in fake   database so no need to create DB class. For example:   fakedatabase/exclusionzone and fakedatabase/ppa

src/harness/fake_db_server.py
jni2000: These   were the parameters passed to handler class in order send a GET request. (Not   required anymore)

src/harness/fakedatabase/exclusion_zone_db.json
jni2000: Modified   the exclusion record to have Geojson

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Modified the key   names

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Added CatB device   and loaded conditionals

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Modified   the name to "fake_db_info"

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Moved creation of Exclusion zone database (exclusion_zone_db.json) from   config generation to test case definition.      Removing the database file ("exclusion_zone_db.json") at the end   of test case execution

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: The   method getRandomLatLongInPolygon returns value that has more than 6 digits   after the decimal point. As per the spec WINNF-TS-0016-V1.2.1 the latitude   and longitude in the registration request use 6 digits after the decimal   point. ("Values are specified using 6 digits to the right of the decimal   point."). Hence the round function is used to round off the values to   six digits after the decimal point.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Modified   the object name

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Supported Multiple   registrations and grants

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Added CatB device   and sending multiple registration and grants

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Added CatB device   and sending multiple registration and grants

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Removed check of   cbsdid in hearbeat response

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Removed check of   cbsdid in relinquishmnet response

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Modified to   assertEqual()

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Defined   the overlapping frequency ranges for 'F1' and 'F2' in generate config method.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: The overlapping   frequency ranges are part of "default.config" file

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Modified   the frequency range of Exclusion zone database.     (Loading Exclusion Zone Database file with Modified record).

src/harness/fake_db_server.py
jni2000: Changed to 2018

src/harness/fake_db_server.py
jni2000: Added   logging statement : 'Unexpected POST request is received.' It raises an error   now.

jni2000:
Kate, we have responded your review comments on this PR, please review.

jni2000:
please can we wrap up this review to unblock the other FDB test cases and WDB test cases? We have responded to the review comments 3 days before. Thanks.

kate-harrison:
I have been concentrating my efforts on other higher-priority PRs. Maybe another company can help with the review of this PR?

jni2000:
Javier

Can commScope help review this PR ? We need to unblock other FDB and WDB test cases. Thanks.

src/harness/fakedatabase/exclusion_zone_db.json
kate-harrison: Please confirm your source for the exclusion zone DB format. I believe this is still incorrect.

src/harness/testcases/configs/test_WINNF_FT_S_FDB_1/default.config
```diff
@@ -0,0 +1,132 @@
+{
```
kate-harrison: Please remove the config files and the unrelated .json files.

jni2000: removed

jni2000:
Kate and Javier

We have updated the PR addressing all your comments. Especially the followings:
1. removed all default config files
2. sync'ed up the exclusion zone data record with what is approved in EXZ.1
3. revised the fake exclusion zone database using the approved exclusion zone data record approved in EXZ.1, the whole database data is composed of two exclusion zone records in a JSON array.

kate-harrison:
The format in EXZ.1 is what we have defined for *injecting* the data into SAS UUT. The format for the test case FDB.1 needs to follow the format of the DB that we expect to use to retrieve this data in production, which is *not* the same as the format used for injecting the data into SAS UUT.

jni2000:
Kate, do you have any FCC/NTIA EXZ database data samples we can use for this? We have a copy of the ground based exclusion zone data may be used. I will send to you via email because of the data size. Please take a look and let us determine if we can start from there.

kate-harrison:
I'm actually uncomfortably unclear on this particular topic. I've just sent an email to WG4 asking for clarification. Hopefully they will be able to answer quickly and definitively, but I'm not sure if WinnForum has all necessary details at this time.

If they are not able to answer the question quickly, I would suggest working instead on the GWBL DB test cases since the format is very well known already. Next in line would be the FSS DB test cases: the format is nearly finalized but SASes haven't had much time to look at it yet.

jni2000:
Thanks, Kate. I just checked with Masoud on this and confirmed that the data we have is same as what NITA published as of today. Here is the link to the NITA site. So if you have no different opinion, we will start from here. Let me know.

https://www.ntia.doc.gov/fcc-filing/2015/ntia-letter-fcc-commercial-operations-3550-3650-mhz-band

jni2000:
As we discussed in the [Sas_test_dev] Re: [Ssc-wg4] Test dev questions about exclusion zones and DPAs thread, looks like the initial exclusion zone data we will need to start with will be the 3 federal radio locations at Indigoes, Pensacola and Pascagoula. The zone is 80 km around these sites.  As FDB.1 only required to create an exclusion zone database and did not specify the data requirement, I would think using geojson formatted schema allows us the flexibility dealing with any types of exclusion zone data, point, polygon and so on. The following is a geojson copy of the database, an array of three geojson points.

[
    {
	"type":"FeatureCollection",
	"features":[
            {
		"type" : "Feature",
		"properties" : {
                    "id" : "FRL_1",
                    "description" : "Federal radiolocation service, St. Inigoes, MD.",
		    "exclusion_distance_from_site_in_km" : 80000
		},
		"geometry" : {
                    "coordinates" : [-76.38333333, 38.16666667],
                    "type" : "Point"
		}
            }
	]
    },
    {
	"type":"FeatureCollection",
	"features":[
            {
		"type" : "Feature",
		"properties" : {
                    "id" : "FRL_2",
                    "description" : "Federal radiolocation service, Pascagoula, MS.",
		    "exclusion_distance_from_site_in_km" : 80000
		},
		"geometry" : {
                    "coordinates" : [-88.48333333, 30.36666667],
                    "type" : "Point"
		}
            }
	]
    },
    {
	"type":"FeatureCollection",
	"features":[
            {
		"type" : "Feature",
		"properties" : {
                    "id" : "FRL_3",
                    "description" : "Federal radiolocation service, Pensacola, FL.",
		    "exclusion_distance_from_site_in_km" : 80000
		},
		"geometry" : {
                    "coordinates" : [-87.27388889, 30.35777778],
                    "type" : "Point"
		}
            }
	]
    }
]

kate-harrison:
I did not see any conclusion about the database location and format on that thread. If Masoud has more information, please ask him to reply to that thread so that others can confirm.

We cannot simply assume or create a format for the database. The point of these test cases is to verify that the SAS UUT can correctly parse and use the information in the final database, and we need to test that by using a database with the same format and access controls as the final database (but not necessarily the same data). So we need to wait until the database format is finalized before we can implement this test case. If Federated can help drive the questions to closure, that would be greatly appreciated. Until then, you may want to work on the other test cases that I mentioned.

jni2000:
Thanks, Kate. The exclusion zone database data information I referred here -- the 3 FRL sites -- is what Yi suggested. We have the communication records in the discussion thread. Please check with him. Of course, this is not final. However, it looks to me that whatever data in the database is not a critical piece to the test case itself. The good part of using geojson is that it is flexible enough to allow us adding heterogeneous data records. Correct me if I am wrong.

kate-harrison:
As mentioned several times, we cannot assume a proprietary data format for these test cases. We need to use the format that will be used in production. I have asked Masoud via email for clarification on the format that will be used in FDB.1.

kate-harrison:
Please, let's keep this PR focused on FDB.1 and not add back all of the other FDB test cases. If you prefer to work on another FDB test case first (which I think is a good idea), just create another PR for that test case and we can come back to this one later when we have all of the required information.

jni2000:
Yes, Kate. We will wait for the final decisions on the access protocols/methods and then move on. These new submissions are just for record keeping for now.

kate-harrison:
Ok, I guess you mean then that this PR should not be reviewed right now?

jni2000:
Yes, you are right. However, if we can review the other part but leave the database format and access protocol portion for future review, that may help also if you agree. I will let you to decide. Thanks.

kate-harrison:
I really like your idea, but would suggest leaving "TODOs" for the database pieces (creating the DB, configuring the DB, changing the contents of the DB) rather than putting in sample code that's not quite right. Then we can quickly review the relatively straightforward parts of the tests -- registration, grant request, etc. -- and tackle the harder pieces in separate followup PRs. I think Google may be able to help in the DB-related parts of some test cases as well, so I'd appreciate your advice on how to best coordinate that.

jni2000:
totally agreed, will do as you suggested. Thanks.

janetliuboston:
Updated to leave TODOs for the database pieces as suggested. Please take a look. Thanks, Kate!

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1014 @@
+#    Copyright 2018 SAS Project Authors. All Rights Reserved.
```
kate-harrison: I think this should be the only file changed in this PR, right? Is it possible to remove the others (I guess for some you may need to rebase?) so that it's clear what the intent of this PR is?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1014 @@
...
+  """SAS federal government database update test cases"""
+
+  def setUp(self):
+    self._sas, self._sas_admin = sas.GetTestingSas()
+    self._sas_admin.Reset()
+
+  def tearDown(self):
+    pass
+
+  def generate_FDB_1_default_config(self, filename):
```
kate-harrison: At the beginning of each test case, there is the following text:

> For clarity, this test is written with a single CBSD and protected entity. The test code shall allow for parallel execution of non-interacting tests in order to reduce total test execution time.

I think that was missed during the implementation. Please let me know if you have any questions.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1014 @@
...
+    del device_b['measCapability']
+
+    conditionals = [conditionals_b]
+
+    # Create the actual config.
+    config = {
+        'registrationRequest': [device_b],
+        'grantRequests': [grant_g1, grant_g2, grant_g3],
+        'conditionalRegistrationData': conditionals
+      }
```
kate-harrison: Please add a TODO to include the database configuration.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1014 @@
...
+
+    # Load device info
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+
+    # Load grant request.
+    grant_g1 = json.load(
+                open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    # Set the grant frequency to overlap with the exclusion zone 'Yuma Proving Ground'
+    #  frequency range 'F1' which is .
```
kate-harrison: Incomplete sentence.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1014 @@
...
+    """Generates the WinnForum configuration for FDB.1"""
+
+    # Load device info
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+
+    # Load grant request.
+    grant_g1 = json.load(
+                open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    # Set the grant frequency to overlap with the exclusion zone 'Yuma Proving Ground'
```
kate-harrison: Note: I'm still trying to figure out which kinds of zones are tested in this test case (I've received some conflicting information). Hopefully the FCC will give us an answer soon.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1014 @@
...
+
+    # Load the configuration file
+    config = loadConfig(config_filename)
+
+    registration_request = {
+        'registrationRequest': config['registrationRequest']
+    }
+
+    # Step 1: Inject FCC ID and User Id of the devices into UUT.
+    self._sas_admin.InjectFccId({'fccId': registration_request['registrationRequest'][0]['fccId']})
```
kate-harrison: You can't inject the FCC IDs: the whole point of the test case is to see that the SAS UUT correctly pulls this information from the DB.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Does this mean that we shall allow multiple instances of e.g. FDB.1 to run concurrently or we shall allow concurrent parallel instances of different FDB test cases, e.g. one FDB.1, one FDB.2 ....?

kate-harrison: It means that you need to take in N CBSDs and expect that there will be at least N exclusion zones in the fake DB. So for example ITS may configure one exclusion zone in NY and another in CA, with one CBSD near each. The zones are far enough away from each other that the zones and CBSDs will not interfere with one another, but the SAS is processing all at the same time and the test harness is checking the results for all at the same time.

The goal was to save some time by allowing ITS to test N scenarios but only calling CPAS once; the alternative (run the test case N times) would require running CPAS N times, which unnecessarily takes more time.

So really the code is not very different at this stage, you just need to take in N registration requests instead of 1, and N*m grant requests instead of only m. The DB part is the same either way (but anyway that's not part of this PR). Does that make sense?

jni2000: Modified   all the test cases, Introduced Multiple CBSDs and Grants (N*m) in required   test cases.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Updated   as per the comment, added TODO for database configurations

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Modified   as per the comment

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Updated,   Removed the injection of FCC ID and User ID.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: we are working on removing other changes. Will update shortly.

jni2000: we have removed unnecessary changes as suggested.

jni2000:
Kate, we have addressed all your latest review comments, please continue. Thanks.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+      }
+    writeConfig(filename, config)
+
+  @configurable_testcase(generate_FDB_1_default_config)
+  def test_WINNF_FT_S_FDB_1(self, config_filename):
+    """Exclusion Zone Database Update"""
+
+    # Load the configuration file
+    config = loadConfig(config_filename)
+
```
kate-harrison: Do some light checking?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+    # Load devices info
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+
+    # Load grant requests
+    grant_g1_a = json.load(
+                open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+    # Set the grant frequency to overlap with the DPA 'puerto_rico_ver2_dpa_4'.
```
kate-harrison: Is this an ESC-monitored or a portal-controlled DPA?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+    }
+    writeConfig(filename, config)
+
+  @configurable_testcase(generate_FDB_2_default_config)
+  def test_WINNF_FT_S_FDB_2(self, config_filename):
+    """DPA Database Update for DPAs which cannot be monitored by an ESC."""
+
+    # Load the configuration file
+    config = loadConfig(config_filename)
+
```
kate-harrison: Add some light checking? E.g. to verify the size of `grantRequests`.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+    conditionals = [conditionals_b]
+
+    # Create the actual config.
+    config = {
+        'registrationRequests': [device_a, device_b],
+        'grantRequests': [
+            [grant_g1_a, grant_g1_b],
+            [grant_g2_a, grant_g2_b]],
+        'conditionalRegistrationData': conditionals
+        # TODO
```
kate-harrison: You need to take the expected response codes as input, too.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+        'cbsdId': cbsd_id,
+        'grantId': grant_id,
+        'operationState': 'GRANTED'
+      })
+    heartbeat_responses = self._sas.Heartbeat(heartbeat_requests)['heartbeatResponse']
+
+    # Check the heartbeat response code is either 0(SUCCESS) or
+    # 501(SUSPENDED_GRANT)
+    for resp in heartbeat_responses:
+      self.assertIn(resp['response']['responseCode'], [0, 501])
```
kate-harrison: This should be configurable: the tester will specify a set of valid response codes for each heartbeat response.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
```
kate-harrison: Same here (configurable).

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+
+    conditionals = [conditionals_b]
+
+    # Create the actual config.
+    config = {
+        'registrationRequest': [device_a, device_b],
+        'grantRequests': [
+            [grant_g1_a, grant_g1_b],
+            [grant_g2_a, grant_g2_b]],
+        'expectedResponseCodes': [(400,), (0,)],
```
kate-harrison: If I read this consistent with other test case implementations, I would take this to mean "device_a must get responseCode 400 during both CHECKs and device_b must get responseCode 0 during both CHECKs". But in fact we need a 2D matrix: device_a could be expected to get 400 during the first check and 0 during the second check, and device_b could have the opposite (or same!) expectation.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+    # Forming of 'G1' grants for all the configured grants
+    addCbsdIdsToRequests(cbsd_ids, config['grantRequests'][0])
+    grant_request_g1 = {'grantRequest': config['grantRequests'][0]}
+
+    grant_response_g1 = self._sas.Grant(grant_request_g1)['grantResponse']
+
+    # Check the grant response code is as expected (SUCCESS or INTERFERENCE)
+    grant_ids = []
+    for resp in grant_response_g1:
+      self.assertIn(resp['response']['responseCode'], config['expectedResponseCodes'][0])
```
kate-harrison: Why `[0]`?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+      grant_ids.append(resp['grantId'])
```
kate-harrison: If the responseCode is 400, you will not have a grantId (and the Grant should not be relinquished in step 6).

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+      self.assertIn(resp['response']['responseCode'], config['expectedResponseCodes'][1])
+
+    del grant_request_g2, grant_response_g2
+
+  def generate_FDB_7_default_config(self, filename):
+    """Generates the WinnForum configuration for FDB.7"""
+
+    # Load devices info
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
```
kate-harrison: Typo?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+
+    # Update the FCC ID of CBSDs with the FCC ID 'F_ID' from FCC ID database.
+    device_a['fccId'] = "F_ID"
```
kate-harrison: `F_ID` in the spec was meant as a var name, not a specific value. Please use different values for these two.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+        # Need to add data base configurations
+    }
+    writeConfig(filename, config)
+
+  @configurable_testcase(generate_FDB_7_default_config)
+  def test_WINNF_FT_S_FDB_7(self, config_filename):
+    """FCC ID Database Update."""
+
+    # Load the configuration file
+    config = loadConfig(config_filename)
```
kate-harrison: Light checking?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+    del device_b['installationParam']
+    del device_b['measCapability']
+
+    conditionals = [conditionals_b]
+
+    # Create the actual config.
+    config = {
+        'registrationRequests': [device_a, device_b],
+        'expectedResponseCodes': [(103,), (0,)],
+        'conditionalRegistrationData': conditionals
```
kate-harrison: To keep things simple, maybe we want to use only Cat A devices in this test case. Otherwise we may have trouble injecting the conditional reg data (we can only do it once the FCC ID has been whitelisted by the SAS, but we don't know exactly when that will be). The logic for Cat A vs. Cat B FCC ID validation should be the same so I don't think it poses a problem for the test case integrity. What do you think? Is there an easier solution?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+    config = loadConfig(config_filename)
+
+    registration_requests = {
+        'registrationRequest': config['registrationRequests']
+    }
+
+    # Step 1: Pre-load conditional registration data for CBSD.
+    if ('conditionalRegistrationData' in config)\
+      and (config['conditionalRegistrationData']):
+      self._sas_admin.PreloadRegistrationData(
```
kate-harrison: To go with my earlier comment: we can't pre-load here because the SAS UUT doesn't know yet that the FCC ID is valid.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+
+    # Step 6: Trigger daily activities
+    self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
+
+    # Step 7: Send again the same registration request for CBSD 'C'
+    # with FCC ID 'F_ID' to SAS UUT.
+    registration_responses = self._sas.Registration(registration_requests)['registrationResponse']
+
+    # Check the registration response code is as expected (SUCCESS or INVALID_VALUE).
+    for resp in registration_responses:
```
kate-harrison: Instead of `[1]`, we should use the index corresponding to this reg request.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+
+  @configurable_testcase(generate_FDB_8_default_config)
+  def test_WINNF_FT_S_FDB_8(self, config_filename):
+    """FSS Database Update: Scheduled Database Update."""
+
+    # Load the configuration file
+    config = loadConfig(config_filename)
+
+    # Step 1: Trigger daily activities
+    self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
```
kate-harrison: You need to use a different trigger since we are attempting to test the *scheduled* CPAS behavior in this test case.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1240 @@
...
+                                    config['conditionalRegistrationData'])
+
+    # TODO
+    # Step 3: Create FSS database which includes atleast one FSS site near location 'X'.
+
+    # TODO
+    # Step 4: Inject the FSS database URL into the UUT
+
+    # Step 5: Trigger daily activities
+    self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
```
kate-harrison: You need to wait until a pre-specified time, not trigger CPAS.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.1        Checking no. of grant requests (grant size).     # Very light checking of the config file.     self.assertEqual(len(config['grantRequests']), 3)

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.2        DPA Database Update for DPAs which cannot be monitored by an ESC

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.4        Added       'expectedResponseCodes':   [0, 501],      in config data structure

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB4.     Used the expectedResponseCodes from the config file during checking   heartbeat response.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB4.     Used the expectedResponseCodes from the config file during checking   heartbeat response.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.6     Added a 2-D matrix for response codes in config data structure:     'expectedResponseCodes': [ [(400,), (0, 400)],  [(0,), (0, 400)]],     Both the lists correspond to the response codes expected in grants 'G1' and   'G2' respectively by devices 'A' and 'B'.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.6     Expected response codes for 'G1' grant will be checked.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.6     grantIds and cbsdIds for 'G1' grants having responseCode as 0 will be   fetched out from the grant response    'G1' and  used to construct   relinquishment request in step 6.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.7     changed to "device_a". (for all the test cases).

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.7        changed to "F_ID_a" and "F_ID_b".

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.7     Added below check .self.assertEqual(len(config['expectedResponseCodes']),   2)

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.7     Loading CAT A devices only.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.7     Removed conditional registration data from config data structure and   pre-loading .

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.7     Using the index corresponding to each registration request against the   registration response.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Please   let us know which API to be used in case of scheduled CPAS?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.8     Added 'cpasInitiateTimer': 100, in config data structure     and     sleep(config['cpasInitiateTimer']) in test case definition step 5.

jni2000:
Kate, thanks for reviewing. We have addressed all your recent review comments except the one on teh scheduled CPAS.  Please can you let us know which API to be used in case of scheduled CPAS? Thanks and please continue review.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -949,7 +966,9 @@ def generate_FDB_6_default_config(self, filename):
         'grantRequests': [
             [grant_g1_a, grant_g1_b],
             [grant_g2_a, grant_g2_b]],
-        'expectedResponseCodes': [(400,), (0,)],
+        'expectedResponseCodes': [
+            [(400,), (0, 400)],
```
kate-harrison: The check should be for 0 OR 400, not both. Please check the spec.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1021,52 +1045,36 @@ def test_WINNF_FT_S_FDB_6(self, config_filename):
...
     """Generates the WinnForum configuration for FDB.7"""
 
     # Load devices info
-    device_a = json.load(
-        open(os.path.join('testcases', 'testdata', 'device_b.json')))
-    device_b = json.load(
-        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+    device_a_0 = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_a_1 = device_a_0
```
kate-harrison: Why are you using the same device twice? Why not just use device_c?

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1078,18 +1086,15 @@ def test_WINNF_FT_S_FDB_7(self, config_filename):
 
     # Load the configuration file
     config = loadConfig(config_filename)
+    
+    # Very light checking of the config file.
+    self.assertEqual(len(config['expectedResponseCodes']), 2)
```
kate-harrison: This should be equal to the number of devices, not 2.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1188,6 +1193,7 @@ def generate_FDB_8_default_config(self, filename):
     config = {
         'registrationRequests': [device_a, device_b],
         'grantRequests': [grant_g_a, grant_g_b],
+        'cpasInitiateTimer': 100,
```
kate-harrison: We will wait until a specified time has passed, not a specific duration. This time should be specified at the top of this file. You can use 2am Pacific for now but please put a "TODO" comment nearby that it will need to be changed once SAS Admins agree on the timing.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
kate-harrison: Please check the section "Trigger SAS to execute “daily activities” as scheduled" in the testing API doc.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
kate-harrison: I think you misunderstand: this test case it to check that the SAS executes CPAS at the correct time *without* being triggered, so you *cannot* trigger it to run CPAS in this test case.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -670,6 +681,9 @@ def test_WINNF_FT_S_FDB_4(self, config_filename):
     # Load the configuration file
     config = loadConfig(config_filename)
 
+    # Very light checking of the config file.
+    self.assertEqual(len(config['grantRequests']), 2)
```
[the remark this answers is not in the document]
jni2000: Updated   as per the comment

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Updated   as per the comment

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Updated,   device 'c' is used now

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: FDB.7     Added:     self.assertEqual(len(config['registrationRequests']),   len(config['expectedResponseCodes']))

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Added   sleep until daily activities scheduled

jni2000:
Kare

We have responded your latest review comments, please take a look. Thanks.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -16,8 +16,12 @@
 import os
 import sas
 import sas_testcase
+from datetime import datetime
+from time import sleep 
 from util import configurable_testcase, writeConfig, loadConfig, addCbsdIdsToRequests
 
+#SAS UUT executes its daily activities at configured time ex- 2, means 2 AM everyday
+SCHEDULED_CPAS_TIMER = 2
```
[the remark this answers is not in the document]
jni2000: Updated,   introduced CPAS_END_TIME

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -949,19 +979,22 @@ def generate_FDB_6_default_config(self, filename):
...
-        'expectedResponseCodes': [(400,), (0,)],
+        'expectedResponseCodes': [
+            [(400,), (400,)],
+            [(0,), (0,)]],
         'conditionalRegistrationData': conditionals
         # TODO
         # Need to add data base configurations
     }
-    writeConfig(filename, config)
+    writcases/WINNF_FT_S_FDB_testcase.pyeConfig(filename, config)
```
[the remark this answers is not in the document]
jni2000: reverted   the change

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1200,9 +1220,13 @@ def test_WINNF_FT_S_FDB_8(self, config_filename):
...
     config = loadConfig(config_filename)
+    
+    # Very light checking of the config file.
+    self.assertEqual(len(config['registrationRequests']),
+                     len(config['grantRequests']))
 
-    # Step 1: Trigger daily activities
-    self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
+    # Step 1: Trigger for enabling "schedule daily activities"
+    self._sas_admin.TriggerEnableScheduledDailyActivities()
```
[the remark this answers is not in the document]
jni2000: updated   in "sas.py"

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1219,8 +1243,19 @@ def test_WINNF_FT_S_FDB_8(self, config_filename):
...
+      scheduled_cpas_time = current_time.replace(day=current_time.day,
+                               hour=SCHEDULED_CPAS_TIMER,
+                               minute=0, second=0, microsecond=0)
+    else :
+      scheduled_cpas_time = current_time.replace(day=current_time.day+1,
+                               hour=SCHEDULED_CPAS_TIMER,
+                               minute=0, second=0, microsecond=0)
+    
+    waiting_time_in_secs = (scheduled_cpas_time - current_time).seconds
+    sleep(waiting_time_in_secs)
```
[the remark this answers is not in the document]
jni2000: Added   second wait based on CPAS end time

jni2000:
Kate, we have responded to your latest review comments, please give a check. Thanks.

src/harness/sas.py
```diff
@@ -230,6 +230,10 @@ def TriggerDailyActivitiesImmediately(self):
     RequestPost('https://%s/admin/trigger/daily_activities_immediately' %
                 self._base_url, None, self._tls_config)
 
+  def TriggerEnableScheduledDailyActivities(self):
```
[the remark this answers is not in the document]
jni2000: Added in   sas_interface.py and fake_sas.py files

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1296 @@
...
+import json
+import os
+import sas
+import sas_testcase
+from datetime import datetime
+from time import sleep
+from util import configurable_testcase, writeConfig, loadConfig, addCbsdIdsToRequests
+
+# START_TIME indicates the schedule start time
+CPAS_START_TIME = 2
```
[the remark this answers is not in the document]
jni2000: Updated   the code based on the comments.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1296 @@
...
+                              grant_request_g,
+                              config['conditionalRegistrationData'])
+
+    # TODO
+    # Step 3: Create FSS database which includes atleast one FSS site near location 'X'.
+
+    # TODO
+    # Step 4: Inject the FSS database URL into the UUT
+
+    # Step 5: Wait until after the completion of scheduled CPAS
```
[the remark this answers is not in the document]
jni2000: updated,   considered the time zone information

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1296 @@
...
+    current_time = datetime.now()
+    if current_time.hour < CPAS_START_TIME:
+      scheduled_cpas_start_time = current_time.replace(day=current_time.day,
+                                                       hour=CPAS_START_TIME,
+                                                       minute=0, second=0,
+                                                       microsecond=0)
+    else:
+      scheduled_cpas_start_time = current_time.replace(day=current_time.day + 1,
```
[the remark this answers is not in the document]
jni2000: updated   the logic

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -0,0 +1,1296 @@
...
+                                    hour=CPAS_END_TIME,
+                                    minute=0, second=0, microsecond=0)
+    else:
+      scheduled_cpas_end_time = scheduled_cpas_start_time.replace(
+                                    day=scheduled_cpas_start_time.day + 1,
+                                    hour=CPAS_END_TIME,
+                                    minute=0, second=0, microsecond=0)
+    completion_time_in_secs = (scheduled_cpas_end_time - scheduled_cpas_start_time).seconds
+
+    # Wait until CPAS is scheduled to start
```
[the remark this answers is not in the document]
jni2000: Added   logs

jni2000:
Kate, thanks for reviewing. We have responded all your latest comments and made necessary changes, please take a look.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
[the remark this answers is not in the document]
kate-harrison: Is there anything wrong with the logic I suggested? It's much less complicated than the logic that's in the code right now.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1248,36 +1255,42 @@ def test_WINNF_FT_S_FDB_8(self, config_filename):
...
+      scheduled_cpas_end_time = scheduled_cpas_start_time + timedelta(hours=CPAS_END_TIME -CPAS_START_TIME)
     else:
-      scheduled_cpas_end_time = scheduled_cpas_start_time.replace(
-                                    day=scheduled_cpas_start_time.day + 1,
-                                    hour=CPAS_END_TIME,
-                                    minute=0, second=0, microsecond=0)
+      scheduled_cpas_end_time = scheduled_cpas_start_time + timedelta(days=1,
+                                                                      hours=CPAS_END_TIME -CPAS_START_TIME)
+
+    # Completion time in seconds
```
kate-harrison: "Run time" is probably more accurate.

jni2000: changed as suggested

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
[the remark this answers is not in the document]
jni2000: Changed as suggested.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1280,18 +1275,18 @@ def test_WINNF_FT_S_FDB_8(self, config_filename):
       scheduled_cpas_end_time = scheduled_cpas_start_time + timedelta(days=1,
                                                                       hours=CPAS_END_TIME -CPAS_START_TIME)
 
-    # Completion time in seconds
-    completion_time_in_secs = (scheduled_cpas_end_time - scheduled_cpas_start_time).seconds
+    # Run time in seconds
+    run_time_in_secs = (scheduled_cpas_end_time - scheduled_cpas_start_time).seconds
 
     # Wait until CPAS is scheduled to start
     logging.debug('Wait time for scheduled CPAS in (HH:MM:SS) %s',
```
kate-harrison: Small thing: "in" should be "is" (same on line 1287).

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1260,16 +1260,11 @@ def test_WINNF_FT_S_FDB_8(self, config_filename):
...
-      scheduled_cpas_start_time = current_time + timedelta(hours=CPAS_START_TIME - current_time.hour,
-                                                           minutes=-current_time.minute,
-                                                           seconds=-current_time.second)
-    else:
-      scheduled_cpas_start_time = current_time + timedelta(days=1, 
-                                                           hours=CPAS_START_TIME - current_time.hour,
-                                                           minutes=-current_time.minute,
-                                                           seconds=-current_time.second)
+    # wait till scheduled CPAS starts
+    scheduled_cpas_start_time = datetime.datetime.combine(datetime.date.today(), datetime.time(CPAS_START_TIME))
```
kate-harrison: I think you have too many `datetime`s here (the calls are not consistent with line 21). Please fix and check that the code can execute.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -13,17 +13,24 @@
...
 
 import json
+import logging
 import os
 import sas
 import sas_testcase
-from datetime import datetime
+import time
+from datetime import datetime, timedelta
+from pytz import timezone
```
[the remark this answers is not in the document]
jni2000: FDB.8     In README.md Updated:          * pytz (https://pypi.org/project/pytz)          This library allows accurate and cross platform timezone calculations using        Python 2.4 or higher.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Updated as per the comment.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
jni2000: Sorry, my bad. Fixed now.

jni2000:
Kate, we have 
1. added "* pytz (https://pypi.org/project/pytz)" library description to the   README as a new dependency.
2. fixed the errors (sorry, I did not run the whole test case but only the portion modified in the previous commit, thanks for pointing out.)

Please take a look.

README.md
```diff
@@ -101,6 +101,11 @@ support for creating testing mocks.
 This Python library can be installed with <code>pip</code>. It provides
 support for advanced functools such as OrderedDict and lru_cache memoization.
 
+* pytz (https://pypi.org/project/pytz)
```
kate-harrison: We typically update this (https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/README.md) readme, not this file.

jni2000: moved as suggested. thanks

PR #392, opened 2018-01-20
New routines to compute area of a shapely geometry.

sbdt-google:
This computes the area using the Chamberlain/Duquette algorithm
for approximated area on a spheroid.
This module is to be used for PPA reference model to:
 1) avoid new dependency on a non-standard external 'area' module
 2) work directly on shapely geometries which avoids transformation
 back and forth to geojson (as required by 'area' module)

Note: a small improvement has been made on the Chamberlain algorithm
to use both the equatoral and polar earth radius.


sushantparkhi:
Just a general comment. We will be getting the Full Activity Dump from SAS UUT to compute the area, and it will be in GeoJSON Format. I know it is easier to work on a shapely object directly, but we again have to convert the SAS UUT PPA to Shapely object and then call this function. Is it more suitable to have the GeoJSON as an input instead of Shapely Object in this function so that we don't have to do that conversion?

sbdt-google:
The use of this routine is only for the PPA creation procedure under the PR #253.
On that one, you are doing internally the calculation with shapely polygons as it provides the required routines for union, etc..
The previously proposed code was:
 - converting the shapely into a GeoJSON
 - computing the area of the holes (in a non compliant fashion, by the way), and using an external library that would have been need to be added in the requirements
  - converting back to shapely, to continue the processing

By using the proposed routines here, you dont have to these back and forth conversion, and we now bring with Winnforum code the area calculation routines instead of relying on an external module maintained by a single guy somewhere.

I dont know any other place in Winnforum where you need to compute the area of a polygon, except for the holes in the PPA creation process so as to remove them.

Let me know if this is not clear

sushantparkhi:
That makes sense! I will add your routine to the #253 PR and remove the area package. As far as I know, test cases like PCR.1 and PCR.2 will need to compute the area of the GeoJSON provided by SAS UUT in Full Activity Dump(Equation 8.3-1 of WINNF-TS-0061).

sbdt-google:
I dont know about those test requiring an area- will check with Kate.

So if you agree, you can approve this change so that I submit it. So that you can use it in your PPA code easily.

sbdt-google:
Ok after talking to Kate, I understand that you have the need of testing the 10% criteria is successfully passed in the testing environment.

In that case, I propose that I extend slightly this toolbox to work also with GeoJSON input. So all area computation are done in a single place with same methodology.
Do you agree ?

sushantparkhi:
Sounds good to me! Let me know when you will push the change, I will then integrate with #253

jni2000:
Thanks, that sounds to be a great solution. let's move on finishing this so that we can check in the PPA reference model which a lot of other test cases are depending on.

sbdt-google:
I can submit the existing right now so you can proceed with the PPA, and make a small update for the use of the testing environment (early next week).

Just waiting for you to give approval on the current code, if we follow the procedure. As soon as I have it, I will submit

jni2000:
Approved, please go ahead to submit the extension of the toolbox to work also with GeoJSON input.

sbdt-google:
Done, submitted

sbdt-google:
Please check the new PR #438 for the extension to GeoJSON.

PR #396, opened 2018-01-22
Federated Wireless delivered SIQ.12

jni2000:


src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -16,8 +16,9 @@
 import os
 import sas
 import sas_testcase
-from util import winnforum_testcase, getRandomLatLongInPolygon, \
-  makePpaAndPalRecordsConsistent
+import logging
+from util import winnforum_testcase, writeConfig,loadConfig, configurable_testcase ,getRandomLatLongInPolygon, \
```
suresh-palanisamy: Maintain the same order as in master. Let the new functions that we added be in a separate line.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    spectrum_inquiry_3 = json.load(
+        open(os.path.join('testcases', 'testdata', 'spectrum_inquiry_0.json')))
+    spectrum_inquiry_3['inquiredSpectrum'] = [{
+        'lowFrequency': '',
+        'highFrequency': 3700000000
+    }]
+    # Create the actual config.
+    devices = [device_a, device_c, device_e]
+    config = {
+        'fccIds': [(d['fccId'], 47) for d in devices],
```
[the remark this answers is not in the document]
jni2000: Moved   the fccid and userid to the testcase.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'userIds': [d['userId'] for d in devices],
+        'registrationRequests': devices,
+        'expectedResponseCodes': [0, 103, 102],
```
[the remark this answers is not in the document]
jni2000: Modified   as per REG.11 testcase

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'pal_record':[pal_record1,pal_record2],
```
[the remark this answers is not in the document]
jni2000: Changed   as per the convention used.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'ppa_record':[ppa_record1,ppa_record2],
+        'spectrum_inquiry':[spectrum_inquiry_1,spectrum_inquiry_2,spectrum_inquiry_3],
+        'm':[[1,8,0,2],[],[],[]],
```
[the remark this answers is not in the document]
jni2000: Added   comment to describe the 'm' variable definition.          

Yes, the test spec has to be updated to consider 5 Mhz and 10 Mhz blocks.   If the SAS responds in either 5 Mhz or 10 Mhz then we can have two sets of M   values. If it provides a mix of 5 and 10 Mhz channels then it gets tricky.   Needs discussion.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'gwpz_requests':[gwpz_e],
```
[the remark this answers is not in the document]
jni2000: Modified   to gwpzRecords

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'fr_cbsd':[3610000000,3620000000]
```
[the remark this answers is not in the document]
jni2000: This   field was added as a configurable field as it was not defined in the test   spec. It is used for verifying the last check step that states    "None of the available channels overlap with the frequency range(s)   specified in FRcbsd." We found that this last checking item in the test case does not make sense as FRcbsd is not defined and if we assume it is the frequency range of the CBSD, it does not make sense to the test case. I have raise a clarification on this in this week's weekly meeting, but unfortunately did not get chance to discuss. We have also asked fro clarification from WinnForum WG.4. Will update accordingly when this is settled.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    config = loadConfig(config_filename)
+    logging.debug("%s test configuration file referred ")
+    #Verify fccId ,User Id , expectedResponseCodes and spectrum inquiry objects are equal to registration objects
+    self.assertEqual(len(config['fccIds']),len(config['registrationRequests']))
+    self.assertEqual(len(config['userIds']),len(config['registrationRequests']))
+    self.assertEqual(len(config['spectrum_inquiry']),len(config['registrationRequests']))
+    self.assertEqual(len(config['expectedResponseCodes']),len(config['registrationRequests']))
+   
+    # Check if N1 >=0 
+    self.assertGreaterEqual(len(config['gwpz_requests']),0)
```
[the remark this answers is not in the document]
jni2000: Corrected   the check. Modified to assertGreater

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+    # Check if N2 >0 
+    self.assertGreater(len(config['fccIds']),0)
+
+    # Check if N3 >=0 
+    self.assertGreaterEqual(len(config['pal_record']),0)
```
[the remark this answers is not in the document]
jni2000: Corrected   the check. Modified to assertGreater

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+    # Load info abot N1 GWBL
+    for gwpz_req in config['gwpz_requests']:
+      self._sas_admin.InjectWisp(gwpz_req)
+
+    # Whitelist N2 FCC IDs.
+    for fcc_id  in config['fccIds']:
```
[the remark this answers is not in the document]
jni2000: The   InjectFccId function uses a default value if the max EIRP is not included in   the request.          The usage here has been corrected to include both the fields as per the   convention followed.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+      self._sas_admin.InjectFccId({
+          'fccId': fcc_id,
+      })
+
+    # Whitelist N2 user IDs.
+    for user_id in config['userIds']:
+      self._sas_admin.InjectUserId({'userId': user_id})
+
+    # Register N2 CBSDs.
+    dp_cert = os.path.join('certs', 'dp_client.cert')
```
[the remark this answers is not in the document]
jni2000: Removed   dp cert and dp key to let the default cert and key to be used.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+   # Load N3 PPA
+    for pal_record in config['pal_record']:
+      self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+    for ppa_record in config['ppa_record']:
+     zone_id = self._sas_admin.InjectZoneData({"record": ppa_record})
+
+ 
+    #Trigger CPAS activity
+    if (len(config['gwpz_requests']) > 0 or len(config['pal_record'])):
```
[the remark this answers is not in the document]
jni2000: Modified   checks of len(gwpzRecords) and len(palRecords) to be consistent

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+    #Load N2 spectrum inquiry requests
+    spectrum_inquiry = config['spectrum_inquiry']
+    for idx, spectrum_inq in enumerate(config['spectrum_inquiry']):   
+      spectrum_inquiry[idx]['cbsdId'] = cbsd_ids[idx] 
+    request = {'spectrumInquiryRequest': spectrum_inquiry}
+    response_spec_inq = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse']
+    
+    #Check Spectrum Inquiry Object
+    no_of_pal_channels = 0
```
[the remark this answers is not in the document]
jni2000: Thanks.   These variables are now reset to zero at the start of every loop.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    no_of_gaa_channels = 0
+    self.assertEqual(len(response_spec_inq),len(config['spectrum_inquiry']))
+    for idx,response in enumerate(response_spec_inq):
+       self.assertEqual(response['response']['responseCode'], config['expectedResponseCodes'][idx])
```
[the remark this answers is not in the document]
jni2000: Included   a validation for available Channels not present for failed inquiry responses

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+       if (response['response']['responseCode'] == 0) :
+          self.assertEqual(response['cbsdId'], cbsd_ids[idx])
+          self.assertTrue('availableChannel' in response)
+          for channel in response['availableChannel']:
+            self.assertEqual(channel['ruleApplied'], 'FCC_PART_96')
+            if channel['channelType'] == 'PAL':
+              no_of_pal_channels = no_of_pal_channels + 1
+            else:
+              self.assertTrue(channel['channelType'] == 'GAA')
```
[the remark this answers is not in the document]
jni2000: Modified   to assertEqual()

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+              no_of_gaa_channels = no_of_gaa_channels + 1  
+            self.assertNotEqual(channel['frequencyRange']['lowFrequency'],config['fr_cbsd'][0])
```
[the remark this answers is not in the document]
jni2000: Aligned   the test verification steps to match the order  as specified in the test specification   except for FRcbsd verification. 

However, the definition of FRcbsd and its usage in verification requires clarification.   The step “None of the available channels overlap with the frequency range(s)   specified in FRcbsd” will be addressed after the clarification.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -592,3 +593,220 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+            self.assertNotEqual(channel['frequencyRange']['highFrequency'],config['fr_cbsd'][1])
+          logging.debug("Number of gaa channels found in response for spectrum request %d is %d",idx,no_of_gaa_channels)
```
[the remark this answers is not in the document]
jni2000: Modified   to GAA and PAL

jni2000:
we just responded most of the review comments and made all necessary changes, except the last check item specified in the test case, which need to be clarified by WG.4. We will address that when clarification is available. Please continue review . Thanks.

kate-harrison:
Sorry for the delay. WG4 is in the process of changing the text for this test case (I think Masoud and Takashi will have the latest details); until that is closer to final, I don't think we can make meaningful progress on the test code. Once we finalize the test spec for SIQ.12 and you have updated the code to match, please let me know and I will happily take a look.

jni2000:
Yes, I got the update from Masoud and we are also in process of modifying the test case. Will update soon.

James Ni
Director of SAS System
978.996.7480

kate-harrison:
Please note that the SIQ.12 update has not been approved (or even reviewed) by WG4 so it should not be considered final. If you have other work that you can do first, I strongly suggest concentrating on that instead so you don't end up needing to rework this code again.

jni2000:
Yes, absolutely.

James Ni
Director of SAS System
978.996.7480

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+            self.assertEqual(channel['ruleApplied'], 'FCC_PART_96')
+
+          # Check that the GAA channel frequency ranges do not overlap with FR1cbsd ranges
+          for channel in response['availableChannel']:
+            if channel['channelType'] == 'GAA':
+                for fr1_range in config['fr1Cbsd'][index]:
+                   
+                   # Check if the channel range is lower than the low freq of FR1cbsd range or if it is higher 
+                   # than the high freq. If not then it is considered to be overlapping
+                   if ((channel['frequencyRange']['lowFrequency'] < fr1_range['frequency']['lowFrequency']) and
```
gilschieber: Do we need the first of these two checks? We are looking at channels for a valid response, so lowFrequency < highFrequency.
Only checking that highFrequency <= lowFreq of fr1_range should be enough no?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+            if channel['channelType'] == 'GAA':
+                for fr1_range in config['fr1Cbsd'][index]:
+                   
+                   # Check if the channel range is lower than the low freq of FR1cbsd range or if it is higher 
+                   # than the high freq. If not then it is considered to be overlapping
+                   if ((channel['frequencyRange']['lowFrequency'] < fr1_range['frequency']['lowFrequency']) and 
+                       (channel['frequencyRange']['highFrequency'] <= fr1_range['frequency']['lowFrequency'])):
+                      pass
+                   else:
+                      if  ((channel['frequencyRange']['lowFrequency'] >= fr1_range['frequency']['highFrequency']) and
```
gilschieber: Same comment here.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+                   # Check if the channel range is lower than the low freq of FR1cbsd range or if it is higher 
+                   # than the high freq. If not then it is considered to be overlapping
+                   if ((channel['frequencyRange']['lowFrequency'] < fr1_range['frequency']['lowFrequency']) and 
+                       (channel['frequencyRange']['highFrequency'] <= fr1_range['frequency']['lowFrequency'])):
+                      pass
+                   else:
+                      if  ((channel['frequencyRange']['lowFrequency'] >= fr1_range['frequency']['highFrequency']) and 
+                       (channel['frequencyRange']['highFrequency'] > fr1_range['frequency']['highFrequency'])):
+                          pass 
+                      else:
```
gilschieber: Could we use if / elif /else instead of having these over-indented conditions?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+    # Make PAL and PPA record consistent
+    ppa_record1, pal_record1 = makePpaAndPalRecordsConsistent(
+        ppa_record1, [pal_record1], pal_low_frequency1, pal_high_frequency1,
+        pal_user_id1)
+    ppa_record2, pal_record2 = makePpaAndPalRecordsConsistent(
+        ppa_record2, [pal_record2], pal_low_frequency2, pal_high_frequency2,
+        pal_user_id2) 
+
+    # Move device_a into the first PPA zone
```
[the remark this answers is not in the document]
jni2000: Modified   to device_b

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    device_b['installationParam']['latitude'], device_b['installationParam'][
+        'longitude'] = getRandomLatLongInPolygon(ppa_record1)
+
+    device_b['installationParam']['latitude'] = round(device_b['installationParam']['latitude'],6)
+    device_b['installationParam']['longitude'] = round(device_b['installationParam']['longitude'],6)
```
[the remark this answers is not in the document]
jni2000: This   device was not intented to be in the cluster list of the PPA. Hence it need   not be inside the PPA. So we have assigned a lat/long outside the polygon and   near the boundary. This function getRandomLatLongInPolygon is no longer used.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    # Step2&3: Register N2 CBSDs
+    # Check registration response
+    # The assertRegistered function does the Inject FCC ID and user ID for the registration requests
+    cbsd_ids = self.assertRegistered(config['registrationRequestsN2'],config['conditionalParameters'])
+
+    # Step4: Load N3 PPA
+    for pal_record in config['palRecordsN3']:
+      self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+    for ppa_record in config['ppaRecordsN3']:
+     zone_id = self._sas_admin.InjectZoneData({"record": ppa_record})
```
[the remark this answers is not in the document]
jni2000: Added   one more space.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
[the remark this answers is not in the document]
jni2000: This   device is not intended to be in the cluster list of the PPA. We have moved it   out of the PPA to a location near the boundary of the PPA.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    spectrum_inquiry_list = config['spectrumInquiryRequestsN2']
+    addCbsdIdsToRequests(cbsd_ids, spectrum_inquiry_list)
+ 
+    request = {'spectrumInquiryRequest': spectrum_inquiry_list}
+    response = self._sas.SpectrumInquiry(request)['spectrumInquiryResponse']
+
+    # Check the response contains N2 SpectrumInquiryResponse objects
+    self.assertEqual(len(response),len(config['spectrumInquiryRequestsN2']))
+
+    for index,response in enumerate(response):
```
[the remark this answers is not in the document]
jni2000: Made all   indentations uniform. The indentation is now set to 2 similar to other test   cases.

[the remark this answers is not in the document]
jni2000: That is   correct. We have changed it. 'request' and 'response' variable names have   been changed to spectrum_inquiry_request and spectrum_inquiry_response to   avoid variable name conflict in the 'for' loop.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+       # If the corresponding request contained a valid cbsdId, the
+       # response shall contain the same cbsdId.
+       if 'cbsdId' in spectrum_inquiry_list[index]:
+        if spectrum_inquiry_list[index]['cbsdId'] in cbsd_ids:
+          self.assertEqual(response['cbsdId'], spectrum_inquiry_list[index]['cbsdId'])
+        else:
```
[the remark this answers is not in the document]
jni2000: This   code was written to address the following statement in the spec.     "o The responseCode shall match the corresponding expected response   code listed in the configuration file. "          This piece of code is written similar to the one used in DRG_6 test   case.          The code also takes care of the case where an 'invalid' cbsdId is sent in   the request. In this case the response should not contain cbsdId.          Hope this is ok:)

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,229 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+          self.assertFalse('cbsdId' in response)
+
+       # Check response code matches the expected response code
+       self.assertIn(response['response']['responseCode'],config['expectedResponseCodes'][index])
```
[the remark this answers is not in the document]
jni2000: This   change was done based on a previous comment from Kate " kate-harrison on   Jan 22 Contributor               I think we need to allow for the possibility that multiple codes may be   acceptable for a single response. This was done in REG.11 by using a list of   sets."          I think for some error scenarios like missing parameters or invalid   parameters different SASs might return different allowed response   codes.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: I think   it is redundant and can be removed.           The redundant checks have been removed and it has been simplified using an   assertTrue statement.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: I think   it is redundant and can be removed.           The redundant checks have been removed and it has been simplified using an   assertTrue statement.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Modified   the statements to assertTrue

jni2000:
gilschieber,

We have done the revision based on your review comments. Please take the next round check and approve if no more concerns.

Thanks.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    # Step1: Load information about N1 GWBZs
+    for gwpz_record in config['gwpzRecordsN1']:
+      self._sas_admin.InjectWisp(gwpz_record)
+
+    # Step2&3: Register N2 CBSDs
+    # Check registration response
+    # The assertRegistered function does the Inject FCC ID and user ID for the registration requests
+    cbsd_ids = self.assertRegistered(config['registrationRequestsN2'],config['conditionalParameters'])
+
+    # Step4: Load N3 PPA
```
gilschieber: I believe a step is missing here? 
From spec, we should have a cluster list defined in the config file (even if the cluster list may be empty). And then check if any device is in the cluster list here. If so, need to update PPA records with cbsd_id of these devices before injecting Zone data.

You can refer to GRA_17 test that uses similar flow.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+      # response shall contain the same cbsdId.
+      if 'cbsdId' in spectrum_inquiry_list[index]:
+        if spectrum_inquiry_list[index]['cbsdId'] in cbsd_ids:
+          self.assertEqual(response['cbsdId'], spectrum_inquiry_list[index]['cbsdId'])
+        else:
+          self.assertFalse('cbsdId' in response)
+
+      # Check response code matches the expected response code
+      self.assertIn(response['response']['responseCode'],config['expectedResponseCodes'][index])
+      if (response['response']['responseCode'] == 0) :
```
gilschieber: Please add a blank line here, as it's the start of item 3 if the checklist.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        # Check if the 'availableChannel' parameter is present
+        self.assertTrue('availableChannel' in response)
```
gilschieber: I'd suggest to insert here the else condition line ('availableChannel' is not present if the response code is not SUCCESS). So that this item of the test check list if more easy to read.

Even if we need to repeat the if condition to test the following items.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    
+    pal_low_frequency1 = 3570000000
+    pal_high_frequency1 = 3580000000
+    pal_user_id1 = "pal_user_1"
+    pal_low_frequency2 = 3600000000
+    pal_high_frequency2 = 3610000000
+    pal_user_id2 = "pal_user_2"
+
+    # Make PAL and PPA record consistent
+    ppa_record1, pal_record1 = makePpaAndPalRecordsConsistent(
```
gilschieber: makePpaAndPalRecordsConsistent() returns pal_records as a list.
Please update name to plural. Same below.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        ppa_record1, [pal_record1], pal_low_frequency1, pal_high_frequency1,
+        pal_user_id1)
+    ppa_record2, pal_record2 = makePpaAndPalRecordsConsistent(
+        ppa_record2, [pal_record2], pal_low_frequency2, pal_high_frequency2,
+        pal_user_id2) 
+
+    # Move device_b near to the first PPA zone
```
gilschieber: Why is it located near a PPA zone? Does it mean you want to restrict its GAA frequencies?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'lowFrequency': 3610000000,
+        'highFrequency': 3590000000
+    }]
+    spectrum_inquiry_2['cbsdId'] = "INVALID_CBSD_ID_12345"
+
+    # 3. Spectrum Inquiry: Missing parameter low frequency and missing cbsdId
+    spectrum_inquiry_3 = json.load(
+        open(os.path.join('testcases', 'testdata', 'spectrum_inquiry_0.json')))
+    spectrum_inquiry_3['inquiredSpectrum'] = [{
+        'lowFrequency': '',
```
gilschieber: We expect lowFrequency to be an integer, here it is an (empty) string. Better to delete all line.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+    # Create the actual config.
+    devices = [device_b, device_c, device_e]
+    conditionals = {
+        'registrationData': [conditionalParameters]
+    }
+
+    config = {
+        'registrationRequestsN2': devices,
+        'conditionalParameters' : conditionals,
```
gilschieber: Use same name as in GRA_17? conditionalRegistrationData?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'expectedResponseCodes': [(0,), (103,), (102,)],  
+        'palRecordsN3':[pal_record1,pal_record2],
+        'ppaRecordsN3':[ppa_record1,ppa_record2],
+        'spectrumInquiryRequestsN2':[spectrum_inquiry_1,spectrum_inquiry_2,spectrum_inquiry_3],
+        'gwpzRecordsN1':[gwpz_e],
+        'fr1Cbsd': [ [{'frequency' :{'lowFrequency': 3600000000,'highFrequency': 3610000000}},
```
gilschieber: I don't understand where these values come from.
For example, fr2Cbsd should correspond to the PAL frequencies responses, but here no Cbsd is part of a cluster list and PPA zone, so fr2Cbsd should be empty.
Also, the values of fr1Cbsd are not consistent with this config. Please check spec.

Also, please reformat to simplify reading.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    self.assertGreater(len(config['registrationRequestsN2']),0)
+
+    # Step1: Load information about N1 GWBZs
+    for gwpz_record in config['gwpzRecordsN1']:
+      self._sas_admin.InjectWisp(gwpz_record)
+
+    # Step2&3: Register N2 CBSDs
+    # Check registration response
+    # The assertRegistered function does the Inject FCC ID and user ID for the registration requests
+    cbsd_ids = self.assertRegistered(config['registrationRequestsN2'],config['conditionalParameters'])
```
gilschieber: We could not have conditionalParameters in the config. Please add a check (similar to lines 1174, 1175 of GRA tests).

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    """[Configurable] Spectrum inquiry in the presence of protected entities."""
+
+    config = loadConfig(config_filename)
+
+    # Light checking of the config file
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['registrationRequestsN2']))
+    self.assertEqual(len(config['expectedResponseCodes']),len(config['spectrumInquiryRequestsN2']))
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['fr1Cbsd']))
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['fr2Cbsd']))
+    self.assertGreater(len(config['registrationRequestsN2']),0)
```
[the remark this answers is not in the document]
jni2000: I think   the number of palrecords and pparecords can be different. For example, if we   load three ppa records then each ppa record could have different sets of pal   records.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+      self._sas_admin.InjectWisp(gwpz_record)
+
+    # Step2&3: Register N2 CBSDs
+    # Check registration response
+    # The assertRegistered function does the Inject FCC ID and user ID for the registration requests
+    cbsd_ids = self.assertRegistered(config['registrationRequestsN2'],config['conditionalParameters'])
+
+    # Step4: Load N3 PPA
+    for pal_record in config['palRecordsN3']:
+      self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
```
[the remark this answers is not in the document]
jni2000: We are   following the the existing convention here. We can discuss this in the working group together on this. For now the codes works. Let's deal with the delivery first.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Ok.   Good catch. We have have updated it similar to GRA_17

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Added a   blank line.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: I agree   that it will make it easier to read but it will also make it to deviate from   the test spec. An explanation has to be added on the reason for the   deviation.          I think it is better to leave it the way it is and aligned to the spec.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -16,8 +16,10 @@
 import os
 import sas
 import sas_testcase
-from util import winnforum_testcase, getRandomLatLongInPolygon, \
-  makePpaAndPalRecordsConsistent, filterChannelsByFrequencyRange
+import logging
+from util import winnforum_testcase, writeConfig,loadConfig, configurable_testcase ,getRandomLatLongInPolygon, \
+  makePpaAndPalRecordsConsistent,generateCpiRsaKeys, generateCpiEcKeys, convertRequestToRequestWithCpiSignature, \
```
jcereced: we are not doing any cpi signature conversions on this TC, could we remove the 3 cpi imports?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,234 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    del device_b['installationParam']
+
+    # N2 spectrum inquiry requests
+    # 1. Spectrum Inquiry: 3550-3700 .
+    spectrum_inquiry_1 = json.load(
+        open(os.path.join('testcases', 'testdata', 'spectrum_inquiry_0.json')))
+    spectrum_inquiry_1['inquiredSpectrum'] = [{
+        'lowFrequency': 3550000000,
+        'highFrequency': 3570000000
+    }]
```
jcereced: should this range be wider so we could get a PAL channel in the response if we move the CBSD in side the PPA?

jni2000:
Javier

Thanks for the review comments. We are addressing you newest unaddressed ones right now. We would like to do it all together. If you have not finished the review yet. please let me know and we can wait. Thanks

jcereced:
Done reviewing, just trying to exercise the code to make sure all corners are covered.  Will try running it again when you guys are done

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +843,219 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        for channel in response['availableChannel']:
+          if channel['channelType'] == 'PAL':
+            isFrequencyIncludedInFr2Cbsd = False
+            for fr2_range in config['fr2Cbsd'][index]:
+              if ((channel['frequencyRange']['lowFrequency'] ==
+                             fr2_range['frequency']['lowFrequency']) and
+                  (channel['frequencyRange']['highFrequency'] ==
+                             fr2_range['frequency']['highFrequency'])):
+                isFrequencyIncludedInFr2Cbsd = True
+            self.assertTrue(isFrequencyIncludedInFr2Cbsd, "PAL channel is not available in Fr2Cbsd list")
```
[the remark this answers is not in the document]
jni2000: Updated   as per the comment

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Updated,   now 'pal_record1' changed to 'pal_records1'

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: latitude,longitude   of device_b parameters are updated to be within PPA zone

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: updated,   removed the line

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Modified,   'conditionalParameters' to 'conditionalRegistrationData'

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Modified   the frequency parameters, now fr1Cbasd and fr2cbsd values are consistent

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: we are   invoking 'assertRegistered' method, which in turn take cares of preloading   the conditional data.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Modified,   removed the cpi related methods

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: updated,   increased the channel size to '3580000000'

jni2000:
Javier

we finished responding all your review comments and made corresponding changes. Please review again. If no more concerns, please approve. Thanks

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,232 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    config = {
+        'registrationRequestsN2': devices,
+        'conditionalRegistrationData' : conditionals,
+        'expectedResponseCodes': [(0,), (103,), (102,)],  
+        'palRecordsN3':[pal_records1,pal_records2],
+        'ppaRecordsN3':[{'ppaRecord':ppa_record1,
+                         'ppaClusterList': ppa_cluster_list_1},
+                        {'ppaRecord':ppa_record2,
+                         'ppaClusterList': ppa_cluster_list_2}],
+
```
jcereced: To add a cbsd to the cluster list of a PPa please refer to SIQ3 alredy checked in.

(from SIQ3)
 # Add devices 1 and 3 to the cluster list of the PPA
    ppa_record['ppaInfo']['cbsdReferenceId'] = [cbsd_ids[0], cbsd_ids[2]]

You have to insert the cbsdReferenceID(s) in ppaInfo

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,232 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'ppaRecordsN3':[{'ppaRecord':ppa_record1,
+                         'ppaClusterList': ppa_cluster_list_1},
+                        {'ppaRecord':ppa_record2,
+                         'ppaClusterList': ppa_cluster_list_2}],
+   
+        'spectrumInquiryRequestsN2':[spectrum_inquiry_1,spectrum_inquiry_2,spectrum_inquiry_3],
+        'gwpzRecordsN1':[gwpz_e],
+        'fr1Cbsd': [ [{'frequency' :{'lowFrequency': 3550000000,'highFrequency': 3560000000}}, 
+                       {'frequency' :{'lowFrequency': 3560000000,'highFrequency': 3570000000}}, 
+                       ],[],[] ],
```
jcereced: i'll have to look at this again tomorrow with fresh eyes, but I think the values for fr1 are incorrect.

From the TC:

o	None of the frequencyRange parameters in the availableChannel parameters marked as available for GAA use overlap with the frequency range(s) specified in FR1cbsd,n2, where FR1cbsd,n2 is a list of the frequency ranges specifed in the configuration file for n2-th CBSD (1≤ n2 ≤ N2), and is consistent with the information of protected entities

The only protected entity near the CBSD is the PPA that we moved it into, so shouldn't fr1 in this case be 3570-3580 (the frequency of the PPA)?

jni2000: We have address this issue in the update submit. Modified FR1cbsd to be consistent with the protected entities.As here, the only protected entity where the CBSD belongs is PPA, made FR1cbsd to be consistent with PPA frequency ranges.

jni2000:
Javier

We addressed your last comment. Please review.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    self.assertGreater(len(config['registrationRequestsN2']),0)
+
+    # Step1: Load information about N1 GWBZs
+    for gwpz_record in config['gwpzRecordsN1']:
+      self._sas_admin.InjectWisp(gwpz_record)
+
+    # Step2&3: Register N2 CBSDs
+    # Check registration response
+    # The assertRegistered function does the Inject FCC ID and user ID for the registration requests
+    cbsd_ids = self.assertRegistered(config['registrationRequestsN2'],config['conditionalRegistrationData'])
```
gilschieber: I am not sure that config is always expected to have a conditionalRegistrationData field.
If it is not present, this would fail. 
Please take into account this case (like in GRA_17), something like:

if ('conditionalRegistrationData' in config) and (config['conditionalRegistrationData']): 
  self.assertRegistered(config['registrationRequestsN2'],config['conditionalRegistrationData'])
else:
  self.assertRegistered(config['registrationRequestsN2'])

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+    # Update PPA records with devices' CBSD IDs and Inject zone data
+    if ('ppaRecordsN3' in config) and (config['ppaRecordsN3']):
+      for ppa in config['ppaRecordsN3']:
+        if len(ppa['ppaClusterList']) != 0 : 
+          ppa['ppaRecord']['ppaInfo']['cbsdReferenceId'] = []
+        for device_index in ppa['ppaClusterList']:
+          ppa['ppaRecord']['ppaInfo']['cbsdReferenceId'].append(cbsd_ids[device_index])
```
gilschieber: Shouldn't this 'for' loop be inside the previous if condition?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+
+    # Light checking of the config file
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['registrationRequestsN2']))
+    self.assertEqual(len(config['expectedResponseCodes']),len(config['spectrumInquiryRequestsN2']))
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['fr1Cbsd']))
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['fr2Cbsd']))
+    self.assertGreater(len(config['registrationRequestsN2']),0)
+
+    # Step1: Load information about N1 GWBZs
+    for gwpz_record in config['gwpzRecordsN1']:
```
gilschieber: As for other cases, we should first check that gwpzRecordsN1 field exist in the config before using it.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    # Update PPA records with devices' CBSD IDs and Inject zone data
+    if ('ppaRecordsN3' in config) and (config['ppaRecordsN3']):
+      for ppa in config['ppaRecordsN3']:
+        if len(ppa['ppaClusterList']) != 0 : 
+          ppa['ppaRecord']['ppaInfo']['cbsdReferenceId'] = []
+        for device_index in ppa['ppaClusterList']:
+          ppa['ppaRecord']['ppaInfo']['cbsdReferenceId'].append(cbsd_ids[device_index])
+
+    # Step4: Load N3 PPA
+    for pal_record in config['palRecordsN3']:
```
gilschieber: You are checking above if ppaRecordsN3 is present in config, why not here also for palRecordsN3?

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+      self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+    for ppa in config['ppaRecordsN3']:
```
gilschieber: Similar comment. Check also here presence of ppaRecordsN3 like above. Or insert in the ' if' condition line 981.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+      zone_id = self._sas_admin.InjectZoneData({"record": ppa['ppaRecord']})
+
+ 
+    # Step5: Trigger CPAS activity
+    if (len(config['gwpzRecordsN1']) > 0 or len(config['palRecordsN3']) > 0):
```
gilschieber: Same problem here...

I propose to add an initial check on the fields (that are not mandatory) of the config file, and set these fields to None if not present. So we can avoid having to check every field.

Smthg like:
if 'field' not in config:
  config['field'] = None

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    ppa_record1, pal_records1 = makePpaAndPalRecordsConsistent(
+        ppa_record1, [pal_record_0], pal_low_frequency1, pal_high_frequency1,
+        pal_user_id1)
+    ppa_record2, pal_records2 = makePpaAndPalRecordsConsistent(
+        ppa_record2, [pal_record_1], pal_low_frequency2, pal_high_frequency2,
+        pal_user_id2) 
+    
+    ppa_cluster_list_1 = [0]
+    ppa_cluster_list_2 = []
+    # Move device_b inside the  PPA zone
```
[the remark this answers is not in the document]
jni2000: Modified   to PPA1

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,231 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'airInterface': device_b['airInterface'],
+        'installationParam': device_b['installationParam'],
+        'measCapability': device_b['measCapability']
+    }
+    del device_b['cbsdCategory']
+    del device_b['airInterface']
+    del device_b['installationParam']
+
+    # N2 spectrum inquiry requests
+    # 1. Spectrum Inquiry: 3550-3700 .
```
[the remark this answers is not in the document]
jni2000: Modified   to 3550-3580

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Yes   you are right.Added an if and else condition as suggested

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: The   functionality of adding "cbsdReferenceId" would be same even if the   "for" loop is outside the "if" condition as   "for" condition would not be executed if ppa['ppaClusterList'] is   empty.          As suggested, moved "for" loop inside if condition to avoid   confusion

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Added   "if" condition to verify the existence of gwpzRecordsN1

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Added   "if" condition to verify the existence of palRecordsN3

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Moved   this above under the check for ppaRecordN3

jni2000:
Javier

We have addressed your latest comments and made corresponding fixes. Please check and approve if no more concerns.

Thanks.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,237 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        'expectedResponseCodes': [(0,), (103,), (102,)],  
+        'palRecordsN3':[pal_records1,pal_records2],
+        'ppaRecordsN3':[{'ppaRecord':ppa_record1,
+                         'ppaClusterList': ppa_cluster_list_1},
+                        {'ppaRecord':ppa_record2,
+                         'ppaClusterList': ppa_cluster_list_2}],
+   
+        'spectrumInquiryRequestsN2':[spectrum_inquiry_1,spectrum_inquiry_2,spectrum_inquiry_3],
+        'gwpzRecordsN1':[gwpz_e],
+        'fr1Cbsd': [ [{'frequency' :{'lowFrequency': 3570000000,'highFrequency': 3580000000}}
```
kate-harrison: Please "pretty print" format this so that it's more readable.

kate-harrison: Does the additional layer of indirection -- from the key 'frequency' -- add any value? Does it make the test code easier to write? If yes, please rename to "frequencyRange" to match with the SAS-CBSD spec. Otherwise, please remove.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,237 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+        
+    # Light checking of the config file
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['registrationRequestsN2']))
+    self.assertEqual(len(config['expectedResponseCodes']),len(config['spectrumInquiryRequestsN2']))
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['fr1Cbsd']))
+    self.assertEqual(len(config['spectrumInquiryRequestsN2']),len(config['fr2Cbsd']))
+    self.assertGreater(len(config['registrationRequestsN2']),0)
+
+    # Step1: Load information about N1 GWBZs
+    if ('gwpzRecordsN1' in config) and (config['gwpzRecordsN1']):
```
kate-harrison: You can assume that this is included in the config, although it may have zero records. Same for all of the other fields except conditionalRegistrationData.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,237 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    devices = [device_b, device_c, device_e]
+    conditionals = {
+        'registrationData': [conditionalParameters]
+    }
+
+    config = {
+        'registrationRequestsN2': devices,
+        'conditionalRegistrationData' : conditionals,
+        'expectedResponseCodes': [(0,), (103,), (102,)],  
+        'palRecordsN3':[pal_records1,pal_records2],
```
kate-harrison: This should be just a list, not a list of lists. Having it as a list of lists leads to another bug later, where you only take the first value of each sub-list (which causes no problems with this default config but would cause problems with a more generic config).

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,237 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    # Step4: Load N3 PPA
+    if ('palRecordsN3' in config) and (config['palRecordsN3']):
+      for pal_record in config['palRecordsN3']:
+        self._sas_admin.InjectPalDatabaseRecord(pal_record[0])
+ 
+    # Update PPA records with devices' CBSD IDs and Inject zone data
+    if ('ppaRecordsN3' in config) and (config['ppaRecordsN3']):
+      for ppa in config['ppaRecordsN3']:
+        if len(ppa['ppaClusterList']) != 0 : 
+          ppa['ppaRecord']['ppaInfo']['cbsdReferenceId'] = []
```
kate-harrison: This field needs to exist, even if the list is empty.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,237 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+            if ((channel['frequencyRange']['lowFrequency'] >= inquired_spectrum['lowFrequency']) and 
+                (channel['frequencyRange']['lowFrequency'] < inquired_spectrum['highFrequency']) and
+                (channel['frequencyRange']['highFrequency'] > inquired_spectrum['lowFrequency']) and
+                (channel['frequencyRange']['highFrequency'] <= inquired_spectrum['highFrequency'])):
+              isFrequencyIncludedInRange = True
+          self.assertTrue(isFrequencyIncludedInRange, "Available channel is not in requested frequency range")
+
+      else:
+        # Check 'availableChannel' is not present if the response code is not SUCCESS
+        self.assertTrue('availableChannel' not in response)
```
kate-harrison: `self.assertFalse('availableChannel' in response)` to make debugging easier.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,237 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+              # than the high freq. If not then it is considered to be overlapping
+              self.assertTrue(((channel['frequencyRange']['highFrequency'] <= fr1_range['frequency']['lowFrequency']) or 
+                               (channel['frequencyRange']['lowFrequency'] >= fr1_range['frequency']['highFrequency'])), 
+                                "GAA channel overlaps with FR1cbsd")
+
+        # Check the PAL channels frequency ranges overlap with FR2cbsd
+        for channel in response['availableChannel']:
+          if channel['channelType'] == 'PAL':
+            isFrequencyIncludedInFr2Cbsd = False
+            for fr2_range in config['fr2Cbsd'][index]:
```
kate-harrison: I think you have to use something like `assertChannelsContainFrequencyRange`. I think it is valid for the SAS to respond with e.g. two 5 MHz channels instead of (or in addition to) one 10 MHz channel.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,237 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    spectrum_inquiry_3['cbsdId'] = "REMOVE"
+
+    # Create the actual config.
+    devices = [device_b, device_c, device_e]
+    conditionals = {
+        'registrationData': [conditionalParameters]
+    }
+
+    config = {
+        'registrationRequestsN2': devices,
```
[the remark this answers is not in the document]
janetliuboston: Agree. Made all the configurable parameter names consistent with other configurable parameters by removing the Nx.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
janetliuboston: Re-formatted this part.

janetliuboston: Agreed. The additional layer of indirection is not needed. Removed here and in relevant places.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
janetliuboston: Removed the extraneous check and assume this field is there, even if there are zero records.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
janetliuboston: Is it okay to have a list of lists, but iterate the list of lists in later code? The palRecords list should match the ppaRecords list, so each palRecords list is associated with each ppaRecord.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
janetliuboston: Took out the check for ppaClusterList > 0, to always default this field to an empty list.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
janetliuboston: Agreed. Modified.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
janetliuboston: Used assertChannelsContainFrequencyRange. May need some clarification if the changes are not as expected. Thanks!

janetliuboston:
Thanks for the comments, Kate! The PR is ready for re-review.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,235 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+    devices = [device_b, device_c, device_e]
+    conditionals = {
+        'registrationData': [conditionalParameters]
+    }
+
+    config = {
+      'registrationRequests': devices,
+      'conditionalRegistrationData': conditionals,
+      'expectedResponseCodes': [(0,), (103,), (102,)],
+      'palRecords': [pal_records1, pal_records2],
```
kate-harrison: Each of these elements is a list, so you will end up with [ [pal1], [pal2] ] which is not quite right. I *think* you can do `pal_records1 + pal_records2` to concatenate the two lists.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
kate-harrison: Actually, I don't think you can assume that they are matched in that way. One PAL record can have two associated PPAs, which would break this scheme. I believe the best approach is to have a single list of PAL records which is not linked directly to the PPAs -- we don't use the linkage anyway, and if we were using it, it would be best recovered by using the PAL IDs inside the PPA record.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -841,6 +842,235 @@ def test_WINNF_FT_S_SIQ_11(self):
...
+              # than the high freq. If not then it is considered to be overlapping
+              self.assertTrue(((channel['frequencyRange']['highFrequency'] <= fr1_range['lowFrequency']) or
+                               (channel['frequencyRange']['lowFrequency'] >= fr1_range['highFrequency'])),
+                                "GAA channel overlaps with FR1cbsd")
+
+        # Check the PAL channels frequency ranges overlap with FR2cbsd
+        for channel in response['availableChannel']:
+          if channel['channelType'] == 'PAL':
+            for fr2_range in config['fr2Cbsd'][index]:
+              self.assertChannelsContainFrequencyRange([channel], fr2_range)
```
kate-harrison: Suppose fr2_range is [3550, 3560] and the SAS returns [ [3550, 3555], [3555, 3560] ]. I don't think it would pass this test case since `channel` will contain e.g. `[3550, 3555]`.

Can you pull out only the PAL channels, then do `self.assertChannelsContainFrequencyRange(pal_channels, fr2_range)`? Although that might cause trouble if there are multiple PAL channels for one CBSD :/

jcereced: Kate,

do you mean that the SAS would respond with 3550-3555 PAL and 3555-3560 PAL (i.e two separate 5 M PAL channels)?

I thought I heard/read somewhere that PAL channels are supposed to be 10M..... if that is not what you meant then disregard.


We rerun the latest code and ti's working fine

kate-harrison: Yes, I meant that the SAS could respond with two separate 5 MHz PAL channels. Masoud and I disagree on this exact point (5 MHz channels): he argues that since the PAL is 10 MHz, the Grants/SIQ responses for PAL should also be 10 MHz, whereas I argue that the PAL user should have at least as much flexibility as the GAA user (which *can* ask for a 5 MHz Grant). We haven't yet reached agreement on this issue, and I hope we can avoid needing to do so.

Anyway, we *do* agree in a different scenario: consider the case where there are two 10 MHz PPAs (both with this CBSD on their cluster list) on different channels, let's say 3550-3560 and 3600-3610. In that case, Masoud and I agree that the SAS should mark both as PAL channels. However, the current code would fail the SAS in that scenario. Do you agree?

I think the right logic here is "for each availableChannel marked PAL, there exists an fr2 which contains that frequency range." I believe that matches the test spec and should allow a correctly-implemented SAS to pass.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
janetliuboston: Thank you. Agree with these good points, and made the fix to have one PAL records list without links to PPAs.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jcereced: so you are saying that before Janet's latest change if the config file was set such as fr2:

'fr2Cbsd': [
        [{'lowFrequency': 3550000000, 'highFrequency': 3560000000},
     {'lowFrequency': 36000000000, 'highFrequency': 36100000000}
],

it would not have worked?

but now with her latest change it'll work?

kate-harrison: Not quite. I'm saying that in the hypothetical scenario (i.e. a new config) where we have two PPAs on those channels (both with CBSD X on their cluster list) and we use fr2Cbsd as you wrote, the current code would fail a correctly-implemented SAS UUT which responded that both of those were PAL channels available to CBSD X.

janetliuboston: Added a utility to check that a given channel is in a list of frequency ranges and used it here instead. This PR is ready for re-review. Thanks!

src/harness/sas_testcase.py
```diff
@@ -167,6 +167,23 @@ def assertChannelsContainFrequencyRange(self, channels, frequency_range):
...
+  def assertChannelIncludedInFrequencyRanges(self, channel, frequency_ranges):
+    """Check if the channel lies within the list of frequency ranges.
+
+    Args:
+      channel: A dictionary containing frequencyRange,
+               which is a dictionary containing lowFrequency and highFrequency
+      frequency_ranges: A list of dictionaries containing
+                        lowFrequency and highFrequency
+    """
+    isFrequencyIncludedInRange = False
```
kate-harrison: `is_channel_included_in_range` (both to match the name `channel` and to use `snake_case` which is our convention).

janetliuboston: Fixed the case to snake_case.

janetliuboston:
Fixed the case in additional location.

janetliuboston:
Sounds good. Thank you, Kate! Will wait for Javi's approval before submitting.

janetliuboston:
Thank you, Javi and Kate!

On Mar 7, 2018, at 12:33 PM, jcereced <notifications@github.com<mailto:notifications@github.com>> wrote:


@jcereced approved this pull request.

Looks good. Ran it again and works fine.

thanks!

PR #397, opened 2018-01-22
Fix antenna gain when beamwidth is 0 or 360 degrees.

sbdt-google:
The spec allows for such extrema values without clear specification
of the behavior. Clearly when those happens, we should consider the
antenna as omnidirectional.

### DIFF
--- a/src/harness/reference_models/antenna/antenna.py
+++ b/src/harness/reference_models/antenna/antenna.py
@@ -88,7 +88,7 @@ def GetStandardAntennaGains(hor_dirs, ant_azimuth=None, ant_beamwidth=None, ant_
     hor_dirs:       Ray directions in horizontal plane (degrees).
                     Either a scalar or an iterable.
     ant_azimut:     Antenna azimuth (degrees).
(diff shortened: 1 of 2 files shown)


src/harness/reference_models/antenna/antenna.py
```diff
@@ -99,7 +99,8 @@ def GetStandardAntennaGains(hor_dirs, ant_azimuth=None, ant_beamwidth=None, ant_
   is_scalar = np.isscalar(hor_dirs)
   hor_dirs = np.atleast_1d(hor_dirs)
 
-  if ant_beamwidth is None or ant_azimuth is None:
+  if (ant_beamwidth is None or ant_azimuth is None or
+      ant_beamwidth == 0 or ant_beamwidth == 360):
```
sushantparkhi: Looks good to me! Just a small typo in docstring! **ant_beamwidth:  Antenna 3dB `curoff->cutoff` beamwidth (degrees)**

sbdt-google: Thanks I fixed the typo

PR #402, opened 2018-01-24
Federated wireless delivered PCR test cases

jni2000:
**## Please the reviewer pay attention to the following notes:**

This pull request is intentionally raised from the "fwPCR-off-ppa_ref_model-branch" to "ppa_ref_model" branch for the following reasons:

1. The "ppa_ref_model" review is still pending to be finished and
2. all the PCR test cases depend on the PPA reference model
3. the "fwPCR-off-ppa_ref_model-branch" is branched off from "ppa_ref_model" to avoid unnecessary future conflict and issues
4. This pull request really meant to allow others to review the test cases earlier

After the review approval, codes in this "fwPCR-off-ppa_ref_model-branch" will be first merged into "ppa_ref_model" branch and then merge into master.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_PCR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_PCR_testcase.py
@@ -45,10 +45,66 @@ import json
 import os
 import sas
 import sas_testcase
+import ConfigParser
 from reference_models.ppa_model.ppa_ref_model import ppa_creation_model_byList
-from util import winnforum_testcase, configurable_testcase, writeConfig, loadConfig
+from util import configurable_testcase, writeConfig, loadConfig
 from sas import _RequestGet
 from shapely.geometry import Polygon
+
+
+def IsSubString(SubStrList, Str):
+  '''''
+     SubStrList=['json']
+     Str='device_a.json'
+     IsSubString(SubStrList,Str)#return True (or False)
+     '''
+  flag = True
+  for substr in SubStrList:
+    if not (substr in Str):
+      flag = False
+  return flag
+
+
+def GetFileList(FindPath, FlagStr=[]):
+  '''''
+     get the file name
+     FlagStr=['json'] # the file name should contains these string
+     FileList=GetFileList(FindPath,FlagStr)
+     '''
+  FileList = []
+  FileNames = os.listdir(FindPath)
+  if (len(FileNames) > 0):
+    for fn in FileNames:
+      if (len(FlagStr) > 0):
+        if (IsSubString(FlagStr, fn)):
+          FileList.append(fn)
+      else:
+        FileList.append(fn)
+
+  if (len(FileList) > 0):
+    FileList.sort()
+  return FileList
+
+def loadMultipleCbsds(self, number):
+  key = ['device']
+  path = 'testcases/testdata/'
+  device_list = GetFileList(path, key)
+  devices = []
+  for device in device_list:
+    device_data = json.load(
+      open(os.path.join('testcases', 'testdata', device)))
+    devices.append(device_data)
+    number -= number
+    if number == 0:
+      return devices
+
+
+def GetSasUnderTestUrl():
+  config_parser = ConfigParser.RawConfigParser()
+  config_parser.read(['sas.cfg'])
+  sas_uut = config_parser.get('SasConfig', 'BaseUrl')
+  return sas_uut
+
 
 class PPACreationTestcase(sas_testcase.SasTestCase):
 
@@ -59,8 +115,7 @@ class PPACreationTestcase(sas_testcase.SasTestCase):
   def tearDown(self):
     pass
 
-
-  def generate_PCR_1_default_config(self, filename):
+  def generate_PCR_1_default_config(self, filename, number):
     # Step 1: Load PAL Database Record
     pal_record = json.load(
       open(os.path.join('testcases', 'testdata', 'pal_record_0.json')))
@@ -77,26 +132,17 @@ class PPACreationTestcase(sas_testcase.SasTestCase):
     pal_record['userId'] = "pal_user"
     pal_record['palId'] = "pal/10-2017/123456789/A"
 
-    # Step 2: Load the device_a in DP Test Harness
-    device_a = json.load(
-      open(os.path.join('testcases', 'testdata', 'device_a.json')))
-    device_a['installationParam']['latitude'] = 39.0119
-    device_a['installationParam']['longitude'] = -98.4842
-    device_a['userId'] = pal_record['licensee']
+    # Step 2: Load N devices
+    devices = loadMultipleCbsds(number)
 
-    # Step 2: Load the device_b in DP Test Harness
-    device_b = json.load(
-      open(os.path.join('testcases', 'testdata', 'device_b.json')))
-    device_b['installationParam']['latitude'] = 39.139836
-    device_b['installationParam']['longitude'] = -97.770094
-    device_b['userId'] = pal_record['licensee']
+    for device in devices:
+      device['userId'] = pal_record['licensee']
 
-    url = 'sas_testUUT_url will be provided in the future'
+    url = GetSasUnderTestUrl()
 
     config = {
         'url': url,
-        'device_a': device_a,
-        'device_b': device_b,
+        'devices': devices,
         'pal_record': pal_record
     }
     writeConfig(filename, config)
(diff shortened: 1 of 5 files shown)


src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+        'device_b': device_b,
+        'pal_record': pal_record
+    }
+    writeConfig(filename, config)
+
+  @configurable_testcase(generate_PCR_1_default_config)
+  def test_WINNF_FT_S_PCR_1(self, config_filename):
+    # Load the Config file
+    config = loadConfig(config_filename)
+    url = config['url']
```
kate-harrison: Is there value to pulling these out? It seems valuable to keep track of what's coming from a config.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+      open(os.path.join('testcases', 'testdata', 'device_b.json')))
+    device_b['installationParam']['latitude'] = 39.139836
+    device_b['installationParam']['longitude'] = -97.770094
+    device_b['userId'] = pal_record['licensee']
+
+    url = 'sas_testUUT_url will be provided in the future'
+
+    config = {
+        'url': url,
+        'device_a': device_a,
```
kate-harrison: This should take a configurable number of devices, not just 2. Please double-check the test spec.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+    # Step 2: register device_b in SAS UUT
+    self._sas_admin.InjectFccId({'fccId': device_b['fccId']})
+    self._sas_admin.InjectUserId({'userId': device_b['userId']})
+    request = {'registrationRequest': [device_b]}
+    response = self._sas.Registration(request)['registrationResponse'][0]
+    # Check registration response of device_b in SAS UUT
+    self.assertEqual(response['response']['responseCode'], 0)
+    del request, response
+    # Step 3: Triggers SAS UUT to create a PPA boundary
+    cbsdIds = [device_a['cbsdSerialNumber'],device_b['cbsdSerialNumber']]
```
kate-harrison: This is not how CBSD IDs are created...

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+    palIds = [pal_record['palId']]
+    response = self._sas_admin.TriggerPpaCreation(cbsdIds, palIds)
```
kate-harrison: This does not match the API described in our testing API doc (which is also in the description of this function).

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+    self.assertEqual(response['response']['responseCode'], 0)
```
kate-harrison: This is not the format of the response from this API.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+    device_b = json.load(
+      open(os.path.join('testcases', 'testdata', 'device_b.json')))
+    device_b['installationParam']['latitude'] = 39.139836
+    device_b['installationParam']['longitude'] = -97.770094
+    device_b['userId'] = pal_record['licensee']
+
+    url = 'sas_testUUT_url will be provided in the future'
+
+    config = {
+        'url': url,
```
kate-harrison: This is not part of the test case config since this is already given in sas.cfg.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+    devices_list = [device_a, device_b]
+    pal_record_list = [pal_record]
+    pal_low_frequency = pal_record['channelAssign']['primaryAssignmen']['lowFrequency']
+    pal_high_frequency = pal_record['channelAssign']['primaryAssignmen']['highFrequency']
+    polygon_result = ppa_creation_model_byList(devices_list, pal_record_list, pal_record['userId'], pal_low_frequency, pal_high_frequency)
+    polygon = polygon_result['features'][0]['geometry']['coordinates'][0]
+    polygon = Polygon([tuple(i) for i in polygon])
+    polygon_dump = Polygon([tuple(i) for i in self.polygon_dump])
+    polygon_overlap = polygon.intersection(polygon_dump)
+    self.assertGreaterEqual(polygon_overlap.area, polygon.area * 0.9)
```
kate-harrison: This does not match the latest test spec.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,671 @@
...
+    response = self._sas_admin.TriggerPpaCreation(cbsdIds, palIds)
+    self.assertEqual(response['response']['responseCode'], 0)
+    self.assertNotEqual(response['ppaId'], None)
+    # Step 4: Triggers the SAS UUT to generate the Full Activity Dump
+    self._sas_admin.TriggerCreateActivityDump()
+    # Step 5: SAS Test Harness uses the SAS-SAS protocol to retrieve the PPA zone
+
+    full_activity_dump = _RequestGet('https://%s' %
+                (url + '/dump'))
+    self.polygon_dump = _RequestGet('https://%s' %
```
kate-harrison: (1) This will not retrieve the data in the format that is expected later.
(2) This assumes that the URL provided by the SAS UUT is not prefixed with "https", which cannot be assumed.
(3) We should not be setting member variables in a test case.

kate-harrison:
Please also split this PR into smaller PRs, as has been suggested several times before. If we can review PCR.1 first (separately) and agree on an approach, then it will be much faster to review the remaining test cases.

jni2000:
Thanks, Kate, and sorry for not carefully checking the errors you mentioned. I am asking our engineers to work on them now. Will be careful not to allow this from happening again.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: Hi Kate,
Do you mean I need to use the random method to generate random number of cbsds?

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: Do you mean I do not need to define variable like "url"? I should directly use config['url']?

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: I will make changing for the CBSD ID generation

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: For the API description, it will return the string ppaId. However, what if the ppa creation fail? Will it return null?

xwu0823: Do we need a dict here to wrap the results into it? It could contains the response code and the ppaId fields.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: In the function descriptions, it seems return JSON response received from server. I think it could return the ppa dump object? Is there any other ways to get the ppa object from SAS UUT?

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: I think the test spec is to calculate the overlap of two polygons and make sure the intersection is equalorgreater 90%?

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: I did not notice any functions in util.py could help me load any numbers of cbsds? Are there any other test cases do the similar things?

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
xwu0823: Here I noticed there is an example of using _RequestGet() it has 'https' as a part of url?

 def _SasRequest(self, method_name, request, ssl_cert=None, ssl_key=None):
    return _RequestGet('https://%s/%s/%s/%s' %
                       (self._base_url, self._sas_version, method_name, request),
                       self._tls_config.WithClientCertificate(
                           ssl_cert or self._GetDefaultSasSSLCertPath(),
                           ssl_key or self._GetDefaultSasSSLKeyPath()))

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: url is   assigned with value as sas_admin URL

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: As of   now there is no helper function     is defined to configure N random number of CBSD's at runtime     We assume N is 2 here.if you want to configure more CBSD devices we can   manually add it default.config     and expected to run the tests as work around          we have to define N and how does the different combination of CBSD looks   like whether it is similar CBSD device or different CBSD devices and what is   the the proportion          Once CBSD Objects changes are finalized we will try use CBSD object to   generate CBSD devices dynamically.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: array of   CBSDIds received in Registration response will be passed as argument to   triggerPPAcreation API

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: will   be addressed by checking ppaId fields as outcome of triggerPPACreation API

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: it will   be addressed by checking array of ppaId in response

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: URL is   removed and assumed that SAS UUT url will be used      wherever applicable

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: Updated   the correct calculation based on latest test spec

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
jni2000: Added   the new method GetSasDumpRecord and      GetSasPPADumpRecord in sas.py to pull FAD records from SAS UTT and fetch   PPA zone from dump record to     calculate maximum ppa boundary

jni2000:
We have responded to all the review comments and make corrections to errors and misunderstanding of spec. etc. Also, we are now focusing on PCR.1 and will add the others when this is sufficiently reviewed. There are a couple of loosing parts to be finalized when PPA reference model is approved. Will take care of these later. Please help continue reviewing the PCR.1

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,178 @@
...
+    self._sas_admin.InjectUserId({'userId': device_a['userId']})
+    self._sas_admin.InjectFccId({'fccId': device_b['fccId']})
+    self._sas_admin.InjectUserId({'userId': device_b['userId']})
+
+    # Do registeration and check registration response of device_a and device_b in SAS UUT
+    cbsd_ids = self.assertRegistered([device_a, device_b])
+
+    # Step 3: Triggers SAS UUT to create a PPA boundary
+    palIds = [pal_record_1['palId']]
+    ppaIds = self._sas_admin.TriggerPpaCreation(cbsd_ids, palIds)
```
kate-harrison: This does not match the testing API that we have defined. Please review https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/sas_interface.py#L365

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # Triggers PPA creation reference model
+    polygon_result = ppa_creation_model_byList(devices_list, pal_record_list, pal_record_1['userId'], pal_low_frequency, pal_high_frequency)
+    polygon = polygon_result['features'][0]['geometry']['coordinates'][0]
+    polygon = Polygon([tuple(i) for i in polygon])
+    polygon_dump = Polygon([tuple(i) for i in self.polygon_dump])
+    polygon_overlap = polygon.symmetric_difference(polygon_dump)
+
+    # Check the Equation 8.3.1 in Test Specfification is satisified w.r t [n.12, R2-PAL-05]
+    self.assertLessEqual(polygon_overlap.area, polygon.area * 0.1)
```
kate-harrison: This does not match the test spec.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -0,0 +1,178 @@
...
+        break
+
+    # Checking whether SAS dump data contains the ppa zone url
+    self.assertIsNotNone(ppa_record_url)
+    self.polygon_dump = self._sas.GetSasPPADumpRecord(ppa_record_url)
+
+    # Step 6: Using PPA Creation Reference Model to calculate the maximum PPA
+    devices_list = [device_a, device_b]
+    pal_record_list = [pal_record_1]
+    pal_low_frequency = pal_record_1['channelAssign']['primaryAssignmen']['lowFrequency']
```
kate-harrison: These are not valid keys for the `pal_record_1` dictionary, so this code will not even be able to execute.

jni2000:
Thanks, Kate. We are in process of wrapping and cleaning up the codes that were implemented by an inexperienced engineer and we are making corrections right now. I will follow up and make sure the quality of the delivery.

James Ni
Director of SAS System
978.996.7480

jni2000:
we have reworked on these test cases from scratch and will submit in a new PR.

PR #408, opened 2018-01-30
New DRG test case using new ID: 6

harinineelu:
- Also includes the default.config
Note:
1. Need clarification on Step 5: CHECK: "If the corresponding request
contained a valid cbsdId the response shall contain the same cbsdId".
(I'll make the change in Line 410, once there is clarity. Right now the
test fails at that point, expecting a 103 and the actual being 102).
2. util.py is the same as in PR #407

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_DRG_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_DRG_testcase.py
@@ -352,9 +352,9 @@ class DeregistrationTestcase(sas_testcase.SasTestCase):
     # Very light checking of the config file.
     self.assertEqual(
         len(config['registrationRequest']), len(config['grantRequest']))
-    self.assertTrue(
+    self.assertEqual(
         len(config['grantRequest']), len(config['heartbeatRequest']))
-    self.assertTrue(
+    self.assertEqual(
         len(config['heartbeatRequest']), len(config['deregistrationRequest']))
     self.assertEqual(
         len(config['deregistrationRequest']),
@@ -389,9 +389,9 @@ class DeregistrationTestcase(sas_testcase.SasTestCase):
     del request, responses
 
     # Step 5: First deregistration request
-    degregister_request = config['deregistrationRequest']
-    addCbsdIdsToRequests(cbsd_ids, degregister_request)
-    request = {'deregistrationRequest': degregister_request}
+    deregister_request = config['deregistrationRequest']
+    addCbsdIdsToRequests(cbsd_ids, deregister_request)
+    request = {'deregistrationRequest': deregister_request}
     responses_1 = self._sas.Deregistration(request)['deregistrationResponse']
     # Check the deregistration response
     self.assertEqual(len(responses_1), len(config['expectedResponseCodes']))
@@ -404,17 +404,17 @@ class DeregistrationTestcase(sas_testcase.SasTestCase):
                     expected_response_codes)
       # "If the corresponding request contained a valid cbsdId, the
       # response shall contain the same cbsdId."
-      if 'cbsdId' in degregister_request[i]:
-        if degregister_request[i]['cbsdId'] in cbsd_ids:
-          self.assertEqual(response['cbsdId'], degregister_request[i]['cbsdId'])
+      if 'cbsdId' in deregister_request[i]:
+        if deregister_request[i]['cbsdId'] in cbsd_ids:
+          self.assertEqual(response['cbsdId'], deregister_request[i]['cbsdId'])
         else:
           self.assertFalse('cbsdId' in response)
 
     # Step 6: Send the Deregistration request from Step 5 again
-    request = {'deregistrationRequest': degregister_request}
+    request = {'deregistrationRequest': deregister_request}
     responses_2 = self._sas.Deregistration(request)['deregistrationResponse']
     # Check the deregistration response
-    self.assertEqual(len(responses_2), len(degregister_request))
+    self.assertEqual(len(responses_2), len(deregister_request))
     self.assertEqual(len(responses_2), len(responses_1))
     # If the corresponding responseCode in the previous Deregistration Response
     # Message was 102, the responseCode shall be 102.
--- a/src/harness/testcases/WINNF_FT_S_DRG_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_DRG_testcase.py
@@ -13,11 +13,14 @@
 #    limitations under the License.
 from datetime import datetime
 import json
+import logging
 import os
 
 import sas
 import sas_testcase
-from util import winnforum_testcase
+from util import winnforum_testcase, configurable_testcase, writeConfig, \
+  loadConfig, addCbsdIdsToRequests, addGrantIdsToRequests
+
 
 class DeregistrationTestcase(sas_testcase.SasTestCase):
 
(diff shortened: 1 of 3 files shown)


src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -269,8 +272,161 @@ def test_WINNF_FT_S_DRG_5(self):
...
+    for device in config['registrationRequest']:
+      self._sas_admin.InjectUserId({'userId': device['userId']})
+
+    # Step 2 & 3: Register devices and get grants
+    cbsd_ids, grant_ids = self.assertRegisteredAndGranted(
+        config['registrationRequest'], config['grantRequest'],
+        config['conditionalRegistrationData'])
+
+    # Step 4: First Heartbeat Request
+    heartbeat_request = config['heartbeatRequest']
```
jni2000: As this step (step4) expects success for all heartbeat requests, the function assertHeartbeatsSuccessful can be used.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -269,8 +272,161 @@ def test_WINNF_FT_S_DRG_5(self):
...
+    # Heartbeat requests (where testcode will insert cbsdId and grantId)
+    heartbeat_a = {'operationState': 'GRANTED'}
+    heartbeat_c = {'operationState': 'GRANTED'}
+    heartbeat_b = {'operationState': 'GRANTED'}
+
+    # Deregistration requests (where testcode will fill in cbsdId)
+    deregister_a = {}
+    # device_c's CBSD ID will be removed due to the 'REMOVE' keyword.
+    deregister_c = {'cbsdId': 'REMOVE'}
+    deregister_b = {}
```
jni2000: Perhaps we can make one of these deregistration requests to contain an invalid cbsdId value.

jni2000:
Comments to the contributor's Note:

1.	Need clarification on Step 5: CHECK: "If the corresponding request
contained a valid cbsdId the response shall contain the same cbsdId".
(I'll make the change in Line 410, once there is clarity. Right now the
test fails at that point, expecting a 103 and the actual being 102).

[FW] step5 mentions that the cbsdId in the deregistration request should match the cbsdId in the response. The code in line 401 already addresses this requirement.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
harinineelu: While I agree that this is a good place to use that helper function, I slightly prefer the way it is now since its much clearer as to what the requests really are. (Also because we resend the same request in Step 8, according to the spec.) So, keeping it as is.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
harinineelu: Good point! Done.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -269,8 +272,175 @@ def test_WINNF_FT_S_DRG_5(self):
...
+
+  @configurable_testcase(generate_DRG_6_default_config)
+  def test_WINNF_FT_S_DRG_6(self, config_filename):
+    """[Configurable] Array Deregistration."""
+
+    config = loadConfig(config_filename)
+    # Very light checking of the config file.
+    self.assertEqual(
+        len(config['registrationRequest']), len(config['grantRequest']))
+    self.assertTrue(
```
jni2000: should this be assertEqual   instead of assertTrue ?

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -269,8 +272,175 @@ def test_WINNF_FT_S_DRG_5(self):
...
+        len(config['grantRequest']), len(config['heartbeatRequest']))
+    self.assertTrue(
```
jni2000: should this be assertEqual   instead of assertTrue ?

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -269,8 +272,175 @@ def test_WINNF_FT_S_DRG_5(self):
...
+    request = {'heartbeatRequest': heartbeat_request}
+    responses = self._sas.Heartbeat(request)['heartbeatResponse']
+    self.assertEqual(len(responses), len(heartbeat_request))
+    # Check heartbeat response
+    for i, response in enumerate(responses):
+      self.assertEqual(response['response']['responseCode'], 0)
+    del request, responses
+
+    # Step 5: First deregistration request
+    degregister_request = config['deregistrationRequest']
```
jni2000: Typo in variable name   degregister_request

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -269,8 +272,175 @@ def test_WINNF_FT_S_DRG_5(self):
...
+      self.assertIn(response['response']['responseCode'],
+                    expected_response_codes)
+      # "If the corresponding request contained a valid cbsdId, the
+      # response shall contain the same cbsdId."
+      if 'cbsdId' in degregister_request[i]:
+        if degregister_request[i]['cbsdId'] in cbsd_ids:
+          self.assertEqual(response['cbsdId'], degregister_request[i]['cbsdId'])
+        else:
+          self.assertFalse('cbsdId' in response)
+
```
jni2000: The spec mentions this   additional check "If responseCode == SUCCESS, verify that the response   contains a valid cbsdId.". I see that this check is redundant, but just   for the sake aligning with the test spec we should either add an additional   check here or have this statement removed from the test spec.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
jni2000: Ok. I agree.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
harinineelu: Done.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
harinineelu: Proposal to have this removed is underway. :)

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
harinineelu: Fixed.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
harinineelu: Fixed.

harinineelu:
Thanks for the review. I've fixed all. Could you pls check.

src/harness/testcases/WINNF_FT_S_DRG_testcase.py
jni2000: Thanks. we have verified the latest submission and have no more comments. I think this PR can be merged once the changes to test spec is approved.

jni2000:
Thanks. we have verified the latest submission and have no more comments. This PR can be merged now, we can pick up the change once the change to test spec is approved.

harinineelu:
Merged #408

PR #411, opened 2018-02-03
Fix flat-earth test mode.

sbdt-google:
The flat earth test mode of the terrain driver allows for simulation on
flat earth with zero altitude.
Unfortunately using zero altitude is detected by some model (E-Hata) as
the sea, and they apply a special correction increasing the pathloss.
To avoid confusion, the flat earth mode will now return a fixed altitude
of 10 meters to simulate being inland.

This does not affect normal operation as this is used only for tests.


sbdt-google:
Because this is a simple fix for internal test only code, I will push it without review,
If any concern, tell me and we can roll back.

PR #412, opened 2018-02-06
Update siq 4 and helper fuction

RedaHossino:
Fix SIQ 4 and treat a corner case for helper function assertChannelsContainFrequencyRange

### DIFF
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -141,9 +141,18 @@ class SasTestCase(sas_interface.SasTestcaseInterface, unittest.TestCase):
       time.sleep(10)
     signal.alarm(0)
     
-  def assertChannelsContainFrequencyRange(self, channels, frequency_range):
+  def assertChannelsContainFrequencyRange(self, available_channels, frequency_range):
+    channels = []
+    # fill channels with the ranges which are not entirely included in any other range.
+    for ch in available_channels:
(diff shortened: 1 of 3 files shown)


src/harness/sas_testcase.py
```diff
@@ -142,16 +142,16 @@ def TriggerDailyActivitiesImmediatelyAndWaitUntilComplete(self):
     signal.alarm(0)
     
   def assertChannelsContainFrequencyRange(self, channels, frequency_range):
-    channels.sort(key=lambda ch: (ch['frequencyRange']['lowFrequency'], \
-                                  ch['frequencyRange']['highFrequency']), reverse = False)
+    channels.sort(key=lambda ch: (ch['frequencyRange']['lowFrequency']), reverse = False)
```
gilschieber: You should keep the previous sort (sorting by freq_low but also by freq_high when same freq_low).
With this new sort, tests will fail for example with:

      channels = [{'frequencyRange' : {'lowFrequency': 3550, 'highFrequency': 3565}},
                          {'frequencyRange' : {'lowFrequency': 3550, 'highFrequency': 3560}}, 
                          {'frequencyRange' : {'lowFrequency': 3565, 'highFrequency': 3575}}]

because the new sort has no effect.
And the lowFrequency of third channel is not Less or Equal to highFrequency of second channel (line 151 assert).

RedaHossino: thank you for this comment I corrected it please review the changes.

gilschieber:
Looks good to me.

PR #417, opened 2018-02-13
Obsolete: Federated Wireless GRA.5 and GRA.6

jni2000:
we have totally reworked GRA.5 and GRA.6 and also took FAD.2 out. FAD.2 is kept in the original PR 368 for maintaining the review history per Kate's request.


jni2000:
Here is the original review comments and our responses. Unfortunately, they were not preserved in the original when recreating this branch to split FAD.2 out. --


1 | kate-harrison   commented 4 days ago     I think the best (only?) way to preserve the comment history is to keep   FAD.2 and the SAS Test Harness in this PR. Otherwise the comments will be   split across two locations. I think it shouldn't be hard to keep FAD.2 here   and move GRA.* to another PR? |   | Done | jni2000 commented 4 days ago     Will do carefully.     GRA.5 and GRA.6 will be move out of this PR and keep this FAD.2 and test   harness only per Kare's suggestion.
-- | -- | -- | -- | --
2 | RedaHossino 5 days ago  Contributor     I think this is device cat B so at least installation param should be   preloaded and deleted from registration request. |   | Done | the   test case does not require that we have to use cat B device, maybe change to   Cat A and keeping the installation parameter is simpler for this test   case.
3 | RedaHossino 5   days ago  Contributor     the grant request should contain the Cbsd Id that the SAS UUT sent in the   registration response     Xi, I think you misunderstand the comments. The CBSD ID is in the      registration response and we need to retrieve it and fill into the grant   request to be sent to the SAS UUT. |   | Done | CBSDIDs received in registration   response will be updated in grant request
4 | I   think you need to inject the user Ids and the Fcc Ids also for the devices(a   and c) because the SAS UUT will receive them in the dump. |   | Done | no, the fcc id   and user id shall be the ones to be registered to the SAS UUT, i.e. b and d   in this case.
5 | RedaHossino   5 days ago  Contributor     isn't it better to do one HB request with both grants? |   | Done | Not   relevant to GRA 5 and GRA 6 tests
6 | RedaHossino   5 days ago  Contributor     the spec required one HB request for both grants. |   | Done | Not   relevant to GRA 5 and GRA 6 tests

kate-harrison:
Thanks for pulling this into a separate PR! Once #368 is submitted and merged into this PR, I will take a look at this PR.I think the review should go quite quickly at that point :)

jni2000:
We finished reworking on GRA.5 and GRA.6. Also FAD.2 is taken out. Please review.

kate-harrison:
Please see my comment above regarding finishing PR #368 first. Others are welcome to review before then, but I personally will not start reviewing this PR until #368 is submitted and the changes merged with this PR. It makes no sense to review the same changes in two separate PRs -- and indeed you are likely to get conflicting review comments -- and GRA.5,6 will be very quick to review once the code in #368 is finalized so I don't think there will be any adverse impact on the overall schedule.

jni2000:
we have reworked GRA.5 and 6 and will submit a new PR.

PR #419, opened 2018-02-15
Updated the Make PPA and PAL Consistent Function

sushantparkhi:
Added an ability to update the Pal Record separately or Pal and Ppa Record together. This will be useful in PCR Test cases or in the execution of PPA Reference Model.

### DIFF
--- a/src/harness/util.py
+++ b/src/harness/util.py
@@ -152,29 +152,28 @@ def getRandomLatLongInPolygon(ppa):
 
 def makePalRecordsConsistent(pal_records, low_frequency, high_frequency,
                              user_id, fcc_channel_id="1",
-                             previous_year_date=None, next_year_date=None):
+                             start_date=None, end_date=None):
   """Make Pal object consistent with the inputs
 
     Args:
-      pal_records: (list) A list of PAL Records in the form of dictionary
-      which has to be associated with the PPA.
-      low_frequency: (number) The Primary Low Frequency for PAL.
-      high_frequency: (number) The Primary High Frequency for PAL.
-      user_id: (string) The userId from the CBSD.
+      pal_records: (list) A list of PAL Records in the form of dictionary.
+      low_frequency: (number) The Primary Low Frequency in Hz for PAL.
+      high_frequency: (number) The Primary High Frequency in Hz for PAL.
+      user_id: (string) The userId to put in PAL Records.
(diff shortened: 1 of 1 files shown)


src/harness/util.py
```diff
@@ -150,28 +150,31 @@ def getRandomLatLongInPolygon(ppa):
...
     Args:
-      ppa_record: (dictionary) A dictionary containing PPA Record.
       pal_records: (list) A list of PAL Records in the form of dictionary
       which has to be associated with the PPA.
       low_frequency: (number) The Primary Low Frequency for PAL.
       high_frequency: (number) The Primary High Frequency for PAL.
       user_id: (string) The userId from the CBSD.
       fcc_channel_id: (string) The FCC-supplied frequency channel identifier.
-
+      previous_year_date: (string) Pal license start date, generally set as one year
```
kate-harrison: Nit: "PAL" instead of "Pal"

kate-harrison: Why not just call it "start_date" directly? The description can give guidance on how to set it. We might want to test with an expired PAL, and a more general name would make that test case less confusing. (Although I'm not sure such a test is actually in the spec right now...)

src/harness/util.py
```diff
@@ -202,6 +205,34 @@ def makePpaAndPalRecordsConsistent(ppa_record, pal_records, low_frequency,
...
+def makePpaAndPalRecordsConsistent(pal_records, ppa_record, low_frequency,
+                                   high_frequency, user_id,
+                                   fcc_channel_id="1"):
+  """Make PPA and PAL object consistent with the inputs
+
+    Args:
+      pal_records: (list) A list of PAL Records in the form of dictionary
+      which has to be associated with the PPA.
+      ppa_record: (dictionary) A dictionary containing PPA Record.
+      low_frequency: (number) The Primary Low Frequency for PAL.
```
kate-harrison: Can you add that it's in Hz? I've seen some Hz vs. MHz errors lately in other code :)

src/harness/util.py
```diff
@@ -202,6 +205,34 @@ def makePpaAndPalRecordsConsistent(ppa_record, pal_records, low_frequency,
...
+      high_frequency: (number) The Primary High Frequency for PAL.
+      user_id: (string) The userId from the CBSD.
```
kate-harrison: Nit: "user ID to put in the PAL records" (more general/accurate)

PR #432, opened 2018-02-22
Check config before preloading REG conditionals.

harinineelu:
If conditionalRegistrationData in config, then preload REG conditionals.


harinineelu:
Merged #432

PR #434, opened 2018-02-23
fix the wrong certificate type for SCS.10 and SDS.10 per Comscope review comments

jni2000:
fix the wrong certificate type for SCS.10 and SDS.10 per Comscope review comments


jni2000:
thanks for the review and quick response.

PR #437, opened 2018-02-26
# update PPA record2 coordinates to be CCW

RedaHossino:


src/harness/testcases/testdata/ppa_record_2.json
```diff
@@ -20,18 +20,18 @@
...
-				  ],
-				  [
-						-88.68280,
-						30.56470
-				  ],
 				  [
 						-88.8407,
 						30.5848
 				  ],
+          [
```
kate-harrison: Please make the indentation consistent with the other lat/long points.

RedaHossino: I updated the indentations of the zone section, is it ok like this?

PR #439, opened 2018-02-27
SCS.17,18,19, SDS.17,18,19

jni2000:
Dear reviewers

We understand that we are still pending on answers/confirmation from WG.2 and ITS on the decisions about CRL server for blacklisting a CBSD certificate. We choose to submit this pull request is for you to help reviewing the part that are known and clear while waiting for the answers.

What we do 

- for SCS/SDS/SSS.11 is that we created a mock CRL server within the test codes for blacklist the CBSD certificate
- for SCS.16-19, SDS.16-19 and SSS.16 we take the short-live certificate approach. This is one of the option as suggested in the test spec.

We respect all different opinions on how to proceed with these test cases before having the WG.2 and ITS answers. As we stated, we would like to start the review of other not-affected portion ahead of time while wait for the answers.

So please help review.


jni2000:
we will use this PR to focus on the dependency resolved SCS/SDS.17,18,19 submission. The ITS has approved using short-lived certificate for these test cases.

SCS.11,16, SDS.11,16, SSS.11,16 will be submitted in separate PRs when all questions are answered.

jni2000:
As ITS already approved that we can use short-lived certificate in test cases SCS/SDS.17,18,19, we removed SCS/SDS/SSS.11 and 16 out of this PR to allow focusing on the finalized ones. Please review the implementation of SCS/SDS.17,18,19. Thanks.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -252,3 +254,152 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
 
     # Check registration response
     self.assertEqual(response['response']['responseCode'],104)
+
+  def generate_SCS_17_default_config(self, filename):
+    """Generates the WinnForum configuration for SCS_17. """
+    # Create the configuration for short lived invalid client cert/key,wait timer
+
+    config = {
+      'clientCert': self.getCertFilename("short_lived_client.cert"),
```
kate-harrison: The cert should be created during the test case execution. Otherwise ITS will have to do some manual steps to run this test case (and any manual step for them increasing our total testing time disproportionately, since now it required them to monitor the tests much more closely and coordinate them more carefully).

src/harness/util.py
```diff
@@ -408,3 +408,15 @@ def filterChannelsByFrequencyRange(channels, freq_range):
       and
       channel['frequencyRange']['highFrequency'] <= freq_range['highFrequency']
   ]
+def countdown(time_to_wait):
```
kate-harrison: Just use sleep -- there's no need to reinvent the wheel here.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Added   creation of certificate before test case execution

src/harness/util.py
jni2000: As per   suggestion used sleep

jni2000:
Kate

We have addressed you latest two comments and made necessary changes. Please review. Thanks.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -252,3 +256,218 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
...
+      os.path.abspath(inspect.getfile(inspect.currentframe())))
+
+    # Absolute path to the certs directory
+    path = os.path.join(testcases_dir, '../certs')
+
+    # Change directory to the certs directory
+    os.chdir(path)
+
+    # Build short lived certificate command
+    command = "./generate_short_lived_certs.sh  CBSD  short_lived_client " + " 2 "
```
kate-harrison: The cert should be generated as part of the test case, not as part of the default config. The function to generate the default config will only be run once, but you need to generate a short-lived cert every time.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -252,3 +256,218 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
 
     # Check registration response
     self.assertEqual(response['response']['responseCode'],104)
+
+  def generate_SCS_17_default_config(self, filename):
+    """Generates the WinnForum configuration for SCS_17. """
+    # Create the configuration for short lived invalid client cert/key,wait timer
+
+    # Get current working directory
+    cwd = os.getcwd()
```
kate-harrison: Is it really necessary to switch dirs?

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: As per   suggestion moved the  creation of   certificate at beginning of the test case. A utility function to create the   certificates has been added to the SecurityTestCase class and is invoked   within the test cases.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: By   switching directory it will create certificate in certs directory. This the   same directory structure can be used in the new script and the it can also be   executed independently from the certs directory if needed.

jni2000:
Kate, we have addressed you latest two comments, please give a check. Thanks.

src/harness/security_testcase.py
```diff
@@ -250,3 +250,38 @@ def _InfoCb(conn, where, ok):
       self.assertEquals(client_ssl.get_peer_finished(), None)
     finally:
       client_ssl.close()
+
+  def create_short_lived_certificate(self, client_type, cert_name, cert_expiry_time):
```
kate-harrison: Please use the same naming style as other functions in this file.

src/harness/security_testcase.py
```diff
@@ -250,3 +250,38 @@ def _InfoCb(conn, where, ok):
...
+    """Generates short lived certificate for SCS/SDS 17,18 & 19
+
+    The function uses the root ca, cbsd ca and proxy ca certificates generated by
+    "generate_fake_certs.sh" script and creates a short lived client certificate
+    based on the input parameters.
+
+    Args:
+      client_type: Type of the device (CBSD or DomainProxy)
+      cert_name: Certificate file name to be created. The private key shall also have
```
kate-harrison: Should this include the ".cert" suffix? Either way, the comment is misleading. Please clarify.

src/harness/security_testcase.py
```diff
@@ -250,3 +250,38 @@ def _InfoCb(conn, where, ok):
...
+        same name.
+      cert_expiry_time: Expiry time of the short lived certificate in minutes
```
kate-harrison: Probably "cert_duration_minutes" would make more sense. The current name suggests that you are giving a timestamp.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -257,117 +256,63 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
...
-  def test_WINNF_FT_S_SCS_17(self,config_filename):
+  @winnforum_testcase
+  def test_WINNF_FT_S_SCS_17(self):
     """Invalid certificate following an approved registration request
        Checks that SAS UUT response with fatal alert message.
     """
-    # Read the configuration
-    config = loadConfig(config_filename)
+    device_cert_name = "short_lived_client"
+    expiry_timer = 2  # in minutes
```
kate-harrison: "cert_duration_minutes"

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -257,117 +256,63 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
...
+
+    self.create_short_lived_certificate("CBSD", device_cert_name, expiry_timer)
+
+    device_cert = self.getCertFilename(device_cert_name + ".cert")
+    device_key = self.getCertFilename(device_cert_name + ".key")
+    wait_timer = (expiry_timer * 60) + 10  # In seconds
```
kate-harrison: Define below, closer to where it's used.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -257,117 +256,63 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
...
     self.assertTlsHandshakeSucceed(self._sas_admin._base_url, ['AES128-GCM-SHA256'],
-                                   config['clientCert'], config['clientKey'])
+                                   device_cert, device_key)
     logging.info("TLS Handshake Succeeded")
 
     # Load the device_a file
     device_a = json.load(
         open(os.path.join('testcases', 'testdata', 'device_a.json')))
 
     with security_testcase.CiphersOverload(self._sas, ['AES128-GCM-SHA256'],
```
kate-harrison: This feels like overkill. Can you just use the normal registration function?

src/harness/certs/README.md
```diff
@@ -109,6 +109,9 @@ Required certificates are:
 * `client_inapplicable.[cert|key]`: leaf CBSD device inapplicable fields certificate
   Used on security test test_WINNF_FT_S_SCS_15.
 
+* `short_lived_client.[cert|key]`: Leaf CBSD certificate that will expire within 5 minutes of creation
```
kate-harrison: No longer accurate.

src/harness/certs/generate_fake_certs.sh
```diff
@@ -64,7 +64,7 @@ openssl ca -cert root-ecc_ca.cert -keyfile private/root-ecc_ca.key -in sas-ecc_c
 echo "\n\nGenerate 'cbsd_ca' certificate/key"
 openssl req -new -newkey rsa:4096 -nodes \
     -reqexts cbsd_ca  -config ../../../cert/openssl.cnf \
-    -out cbsd_ca.csr -keyout private/cbsd_ca.key \
+    -out cbsd_ca.csr -keyout cbsd_ca.key \
```
kate-harrison: Why change this?

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -252,3 +255,137 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
 
     # Check registration response
     self.assertEqual(response['response']['responseCode'],104)
+
+  @winnforum_testcase
+  def test_WINNF_FT_S_SCS_17(self):
+    """Invalid certificate following an approved registration request
+       Checks that SAS UUT response with fatal alert message.
```
pmarks-net: This 3-space indentation is inconsistent with the other docstrings.  In general, the second line should be blank:
https://www.python.org/dev/peps/pep-0008/#doc

(This applies throughout the pull request.)

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -252,3 +255,137 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
...
+    with security_testcase.CiphersOverload(self._sas, ['AES128-GCM-SHA256'],
+                                           device_cert,
+                                           device_key):
+      self.assertRegistered([device_a])
+
+    logging.info("CBSD device is in registered state")
+    logging.info("Waiting for %s secs so the certificate will become invalid" % wait_timer)
+
+    # sleep for waitTimer so that certificate will become invalid
+    time.sleep(wait_timer)
```
pmarks-net: People will be disincentivized from running a test that sleeps for 2+ minutes.  Can this be done reliably in less time?

Here's an idea for a better algorithm:
- Assert that the first request succeeds.
- Retry the request in a loop, with a 1-second `sleep`.
- Assert that the number of iterations before failure is within reasonable bounds.

That will allow the test to finish ASAP, and immediate feedback will make it easier to tune the certificate lifetime.

src/harness/security_testcase.py
jni2000: Changed   to camelCase.

src/harness/security_testcase.py
jni2000: It is   just the name for the certificate and private key without the .cert suffix.   The parameter description has been updated accordingly.

src/harness/security_testcase.py
jni2000: Updated   accordingly

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Updated   accordingly

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Moved   this code to the place where it is used.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: This   approach is used so that the device_cert and device_key configured for this   test case is used. The same approach is used for SCS_18 test case where   assertRegisteredAndGranted method is used with the test case specific device   certificate and key instead of the default ones. This approach follows the   ones used in test cases SCS1 to SCS5.

src/harness/certs/README.md
jni2000: Removed   this file from the PR, as the short lived certificates for devices and domain   proxy are created at run time and they are no longer needed to be generated   using the "generate_fake_certs.sh" script before the tests are run.

src/harness/certs/generate_fake_certs.sh
jni2000: The   change was done so that the private keys are not removed at the end of the   script. The private keys of the case are required for signing the short-lived   certificates.          The change has been reverted and the CA private keys are stored under the   'private' directory. Now the 'private' directory is not removed at the end of   the script. This is done to be consistent with the simple CRL server   implementation PR (this directory is also not deleted in this PR). For   certificate revocation and for creating short lived certificates this   directory is needed.

src/harness/certs/generate_short_lived_certs.sh
```diff
@@ -0,0 +1,78 @@
...
+    echo "The intermediate domain proxy ca file not found"
+    return -1
+}
+# Argument1 : Type (CBSD, DomainProxy)
+if [ "$1" == "CBSD"  ]; then
+    generate_cbsd_short_lived_certificate  $2 $3
+elif [ "$1" == "DomainProxy"  ]; then
+    generate_dp_short_lived_certificate   $2 $3
+else
+    echo "wrong option other than (CBSD, Domain_Proxy)"
```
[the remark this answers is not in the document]
jni2000: Updated   correctly.

src/harness/certs/generate_short_lived_certs.sh
```diff
@@ -0,0 +1,78 @@
...
+
+function generate_dp_short_lived_certificate()
+{
+#   Argument1: Name of the certificate to be generated
+#   Argument2: Time (in minutes)
+
+# Create certificate for test case which use short lived domain proxy certificate\
+    if [ -f proxy_ca.cert -a -f proxy_ca.key ]; then
+        echo "Create short lived certificate"
+        current_time=`date -u  +%y%m%d%H%M%SZ`
```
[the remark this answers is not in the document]
jni2000: Removed.

src/harness/testcases/WINNF_FT_S_SDS_testcase.py
```diff
@@ -264,3 +266,142 @@ def test_WINNF_FT_S_SDS_15(self,config_filename):
...
+
+    # Load the grant_0 file
+    grant_0 = json.load(
+      open(os.path.join('testcases', 'testdata', 'grant_0.json')))
+
+    with security_testcase.CiphersOverload(self._sas, ['AES128-GCM-SHA256'],
+                                           domain_proxy_cert,
+                                           domain_proxy_key):
+      cbsd_ids,grant_ids = self.assertRegisteredAndGranted([device_a],[grant_0])
+    logging.info("CBSD device  is in Granted State")
```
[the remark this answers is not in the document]
jni2000: Removed.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -13,10 +13,13 @@
 #    limitations under the License.
 
 import json
+import logging
 import os
 import security_testcase
+import time
 from OpenSSL import SSL
-from util import winnforum_testcase,configurable_testcase, writeConfig,loadConfig
+from util import winnforum_testcase,configurable_testcase, writeConfig, loadConfig
```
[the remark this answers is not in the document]
jni2000: Added   the line feed to achieve readability and to be consistent with the convention   used in the rest of the files.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Updated   accordingly

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: The   time duration has been reduced. Certificate expiry time has been changed from   2 minutes to 1 minute.          As we know the time at which the certificate will expire, I think we do not   have to run requests in a loop until it fails. Also the test case description   does not expect multiple TLS handshakes to happen. The current algorithm just   waits until the time the certificate expires (a 5 sec buffer is also added)   and then tries the TLS handshake again and verifies that the TLS handshake   fails.

jni2000:
Kate and Paul

We have addressed your review comments and made necessary changes/corrections. Additionally, we have also addressed some of the minor comments like unwanted space, uniform indentation,. based on the other PR comments. Please take a look and move on with the CRL server PR. Thanks.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
kate-harrison: This approach is used for those (SCS.1-5) test cases because we need to specifically test the ciphersuite properties of the SAS. This is not the case in this test.

But I see your point that this is the easiest way to use the existing helper functions without further modification. Please use the original list of ciphersuites and add a note about why you are doing it this way.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -256,136 +257,137 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
...
+    operation_states = ['GRANTED']
 
-    operation_states = [('GRANTED')]
-
-    with security_testcase.CiphersOverload(self._sas, ['AES128-GCM-SHA256'],
-                                           device_cert,
-                                           device_key):
-      transmit_expire_time = self.assertHeartbeatsSuccessful(cbsd_ids,grant_ids,operation_states)
+    # Send the Heartbeat request for the Grant of CBSD to SAS UUT.
+    with security_testcase.CiphersOverload(self._sas, ['AES128-GCM-SHA256'], device_cert, device_key):
```
kate-harrison: You could combine this with the previous `with` statement, no?

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: We   agree. We have added a comment to clarify. Made changes to use the original   list of ciphersuites instead of hardcoding it to   "['AES128-GCM-SHA256']".

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
jni2000: Yes. We   can combine them.           The two statements are combined under the    same "with" block.

src/harness/certs/generate_short_lived_certs.sh
```diff
@@ -0,0 +1,75 @@
...
+        -out $1.cert -outdir ./root \
+        -policy policy_anything -extensions oper_req_sign -config ../../../cert/openssl.cnf \
+        -batch -notext -create_serial -utf8 -days 1185 -md sha384 -enddate $enddate_value
+    return 0
+  fi
+  echo "The intermediate Domain Proxy CA file not found."
+  return -1
+}
+# Argument1 : Type (CBSD, DomainProxy).
+if [ $1 == "CBSD" ]; then
```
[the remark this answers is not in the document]
jni2000: Thanks.   The shell variables are now double quoted.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -252,3 +256,138 @@ def test_WINNF_FT_S_SCS_15(self,config_filename):
...
+
+    # Create a short lived certificate signed by intermediate CBSD CA.
+    self.createShortLivedCertificate("CBSD", device_cert_name, cert_duration_minutes)
+
+    # Get the absolute path of the short lived certificate created.
+    device_cert = self.getCertFilename(device_cert_name + ".cert")
+    device_key = self.getCertFilename(device_cert_name + ".key")
+
+    # Successful TLS Handshake
+    self.assertTlsHandshakeSucceed(self._sas_admin._base_url, ['AES128-GCM-SHA256'],
```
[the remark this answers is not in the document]
jni2000: The   test specification expects a particular TLS handshake to either succeed or   fail. Hence for most of the tests this is mutually exclusive and if the   alternate happens (TLS succeeds when we expect a failure) then we assert and   raise an exception. However a few steps in the test spec allows the alternate   to happen. In such cases we catch the exception and perform the rest of the   steps for that test case.          I think as the test spec is very clear on the expected TLS outcome we   should be fine with having two independent functions

jni2000:
Kate and Paul

We have addressed your latest review comments. Please take a look and let's move forward if no more concerns.

Thanks

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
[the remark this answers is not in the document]
pmarks-net: > The test specification expects a particular TLS handshake to either succeed or fail.

Great, so where is the code that reports whether a particular TLS handshake succeeds or fails?  Unless both cases are testing the same code, I cannot trust the test.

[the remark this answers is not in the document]
jni2000: At this   step the test case expects TLS handshake to succeed. Hence the method   assertTlsHandshakeSucceed is invoked here. If the TLS handshake fails then   the method assertTlsHandshakeSucceed will raise an exception and the test   case will fail as expected. The code that checks whether a TLS handshake   failed is present in the method assertTlsHandshakeSucceed. Test case SCS10 is   an example where both success and failure are handled.

jni2000:
Paul

We have answered your question. Please take a look and move forward.

Thanks.

James

src/harness/certs/generate_short_lived_certs.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  # Create certificate for test case which use short lived certificate.
+  if [ -f cbsd_ca.cert -a -f private/cbsd_ca.key ]; then
+    echo "Create short lived certificate"
+    current_time=`date -u +%y%m%d%H%M%SZ`
+    offset=$2
+    enddate_value=$(date -u -d "now + $offset minutes" '+%y%m%d%H%M%SZ')
+
+    # Generate normal operation short_lived certificate/key.
+    echo "\n\nGenerate 'short_lived' certificate/key"
```
pmarks-net: This needs `echo -e` to work correctly:

    $ echo "\ntest"
    \ntest
    $ echo -e "\ntest"
    
    test

jni2000:
Thanks, Paul, will take a look of them and get back to you soon.

On Mar 23, 2018 3:02 PM, Paul Marks <notifications@github.com> wrote:

@pmarks-net commented on this pull request.

________________________________

In src/harness/certs/generate_short_lived_certs.sh<https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/439#discussion_r176836488>:

> +echo -n 'unique_subject = no' >> index.txt.attr
+
+function generate_cbsd_short_lived_certificate()
+{
+  # Argument1: Name of the certificate to be generated.
+  # Argument2: Time (in minutes).
+
+  # Create certificate for test case which use short lived certificate.
+  if [ -f cbsd_ca.cert -a -f private/cbsd_ca.key ]; then
+    echo "Create short lived certificate"
+    current_time=`date -u +%y%m%d%H%M%SZ`
+    offset=$2
+    enddate_value=$(date -u -d "now + $offset minutes" '+%y%m%d%H%M%SZ')
+
+    # Generate normal operation short_lived certificate/key.
+    echo "\n\nGenerate 'short_lived' certificate/key"


This needs echo -e to work correctly:

$ echo "\ntest"
\ntest
$ echo -e "\ntest"

test

src/harness/certs/generate_short_lived_certs.sh
jni2000: fixed both places. Thanks for the findings.

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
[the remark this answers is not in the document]
pmarks-net: I moved this to https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues/488

jni2000:
Thanks, Paul, we will keep eyes on and help work on #488.

PR #441, opened 2018-02-27
fix SCS.10 and SDS.10 issues

jni2000:
This PR is to address the issues found by Tahashi of Sony.

[1] SCS.10
-line 234 :    -policy policy_anything -extensions sas_req_sign -config ../../../cert/openssl.cnf \

  Comment: 
This will create cert of the wrong usage (serverAuth). 
I’ve Looked into the openssl.cnf file and I believe the correct way is: 

  -policy policy_anything -extensions wrong_cbsd_req_sign -config ../../../cert/openssl.cnf \

[FW] We agree with this comment and the TLS authentication will fail for the wrong usage. We will use sas_client_mode_req_sign instead of wrong_cbsd_req_sign. I think that will be correct approach. Both wrong_cbsd_req_sign and sas_client_mode_req_sign are exactly the same. Both will create a SAS certificate with usage clientAuth. sas_client_mode_req_sign
was added to address the usage issue.

[2] SDS.10
-line 306:    -policy policy_anything -extensions sas_req_sign -config ../../../cert/openssl.cnf \

  Comment: 
SDS.10 would be the same situation. 

[FW] Similar to the previous comment sas_client_mode_req_sign will be used and we do not have to create  a wrong_oper_req_sign section.

### DIFF
--- a/cert/openssl.cnf
+++ b/cert/openssl.cnf
@@ -510,15 +510,6 @@ TEST = critical, ASN1:NULL
 
 policyIdentifier = 2.16.840.1.114412.2.1
  CPS.1="https://www.digicert.com/CPS"
-
-[ wrong_cbsd_req_sign ]
-subjectKeyIdentifier = hash
-authorityKeyIdentifier = keyid:always,issuer
-basicConstraints = CA:FALSE
-keyUsage = critical, digitalSignature, keyEncipherment
-extendedKeyUsage = clientAuth
-certificatePolicies=@cps,ROLE_SAS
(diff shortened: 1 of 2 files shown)


jni2000:
Sorry for late. 
I’ve reviewed, this is OK. I approve this PR.

Thanks. 

Best Regards,
Takashi

jni2000:
Kate, this fix has been reviewed and approved by Takashi. Please can you help remove the merge blocker as he has no direct access to github? Thanks.

PR #442, opened 2018-02-27
Add a new utility routine GridPolygon().

sbdt-google:
Useful to efficiently grid a zone from a GeoJSON or a Shapely
polygon (PPA, DPA, ...).

### DIFF
--- a/src/harness/reference_models/geo/utils.py
+++ b/src/harness/reference_models/geo/utils.py
@@ -71,10 +71,10 @@ def GridPolygon(poly, res_arcsec):
 
   res = res_arcsec / 3600.
   bounds = poly.bounds
-  lng_min = int(bounds[0] / res) * res
-  lat_min = int(bounds[1] / res) * res
-  lng_max = int(bounds[2] / res + 1) * res
-  lat_max = int(bounds[3] / res + 1) * res
+  lng_min = np.floor(bounds[0] / res) * res
+  lat_min = np.floor(bounds[1] / res) * res
+  lng_max = np.ceil(bounds[2] / res + 1) * res
+  lat_max = np.ceil(bounds[3] / res + 1) * res
   mesh_lng, mesh_lat = np.mgrid[lng_min:lng_max:res,
                                 lat_min:lat_max:res]
   points = np.vstack((mesh_lng.ravel(), mesh_lat.ravel())).T
--- a/src/harness/reference_models/geo/utils.py
+++ b/src/harness/reference_models/geo/utils.py
@@ -73,8 +73,8 @@ def GridPolygon(poly, res_arcsec):
   bounds = poly.bounds
   lng_min = np.floor(bounds[0] / res) * res
   lat_min = np.floor(bounds[1] / res) * res
-  lng_max = np.ceil(bounds[2] / res + 1) * res
-  lat_max = np.ceil(bounds[3] / res + 1) * res
+  lng_max = np.ceil(bounds[2] / res) * res + res/2
+  lat_max = np.ceil(bounds[3] / res) * res + res/2
   mesh_lng, mesh_lat = np.mgrid[lng_min:lng_max:res,
                                 lat_min:lat_max:res]
   points = np.vstack((mesh_lng.ravel(), mesh_lat.ravel())).T
--- a/src/harness/reference_models/geo/utils.py
+++ b/src/harness/reference_models/geo/utils.py
@@ -12,7 +12,7 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-"""Utility routines.
+"""Utility geometry routines.
 """
 
 import shapely.geometry as sgeo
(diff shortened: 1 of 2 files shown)


src/harness/reference_models/geo/utils.py
```diff
@@ -42,6 +42,48 @@ def GeoJsonToShapelyGeometry(geo):
...
+  lat_min = int(bounds[1] / res) * res
+  lng_max = int(bounds[2] / res + 1) * res
+  lat_max = int(bounds[3] / res + 1) * res
+  mesh_lng, mesh_lat = np.mgrid[lng_min:lng_max:res,
+                                lat_min:lat_max:res]
+  points = np.vstack((mesh_lng.ravel(), mesh_lat.ravel())).T
+  pts = poly.intersection(sgeo.asMultiPoint(points))
+  if isinstance(pts, sgeo.Point):
+    return [(pts.x, pts.y)]
+  return [(p.x, p.y) for p in pts]
```
arhannan: what gets returned if isinstance(pts, sgeo.Point) is False?

sbdt-google: FYI, I have done a further review cycle on the PAT.
If Point is false, then it is a MultiPoint and it returns a list of (lon,lat) tuples.

src/harness/reference_models/geo/utils.py
```diff
@@ -42,6 +42,48 @@ def GeoJsonToShapelyGeometry(geo):
...
+    # Optional block: for MultiPolygons, we process per polygon
+    # to avoid inefficiencies if polygons largely disjoint.
+    pts = ops.unary_union(
+        [sgeo.asMultiPoint(GridPolygon(p, res_arcsec))
+         for p in poly])
+    return [(p.x, p.y) for p in pts]
+
+  res = res_arcsec / 3600.
+  bounds = poly.bounds
+  lng_min = int(bounds[0] / res) * res
```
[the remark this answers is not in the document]
sbdt-google: I made the change, although it wont have any effect: 
because if the min of the division with the step (lat_min/res) is -100.5, it means you actually have to start putting a point at -100, and not at -101 which is outside of the polygon for sure.
But for the sake of extra safeguard, i did the change.

sbdt-google:
Change done, see my comment.

src/harness/reference_models/geo/utils.py
```diff
@@ -71,10 +71,10 @@ def GridPolygon(poly, res_arcsec):
 
   res = res_arcsec / 3600.
   bounds = poly.bounds
-  lng_min = int(bounds[0] / res) * res
-  lat_min = int(bounds[1] / res) * res
-  lng_max = int(bounds[2] / res + 1) * res
-  lat_max = int(bounds[3] / res + 1) * res
+  lng_min = np.floor(bounds[0] / res) * res
+  lat_min = np.floor(bounds[1] / res) * res
+  lng_max = np.ceil(bounds[2] / res + 1) * res
```
arhannan: np.ceil(bounds[2] / res) * res

src/harness/reference_models/geo/utils.py
```diff
@@ -71,10 +71,10 @@ def GridPolygon(poly, res_arcsec):
...
+  lat_max = np.ceil(bounds[3] / res + 1) * res
```
arhannan: np.ceil(bounds[3] / res ) * res

src/harness/reference_models/geo/utils.py
sbdt-google: I am sorry but I dont think  this is correct: 
Imagine that
  bounds[2] = 10 
  res = 1 degree (3600 arcsec)
Then you have:
  lng_max = 10 
Now when we build the array of points to check, we do not include lng_max in the test, although it corresponds to a point at the border.

If you want to save testing for points which are obviously outside, then my previous version with int() on the left and int( ... + 1) on the right is the most accurate. 
Now with the floor, we have one more set of points on the left that we include on the test although they are outside. And on the right the ceil() usually introduce one serie of more points as well.

sbdt-google:
Thanks for the review

src/harness/reference_models/geo/utils.py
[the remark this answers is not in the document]
arhannan: It does have an effect. Say we have 360 arcsec resolution
if lng_min = -100.5, and lng_max = -100, int will be -100. thus skipping -100.5, .6, .7, .8, .9 
while floor, -101 -100 respectively. grid will include -100.5, .6, .7, .8, .9

arhannan:
The protection points returned by the function will include interior holes. Do we want to include these  in the protection points?

src/harness/reference_models/geo/utils.py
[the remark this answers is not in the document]
sbdt-google: I think you got confused, by the meaning of the -100 in my example: it was not the latitude but the ratio of latitude and res: latitude/res.

Taking back your example of real_lng_min=-100.5 and real_lng_max=100, with res_arcsec = 360:

   res = 360/3600 = 1/10
Case int:
   lng_min = int(-100.5 * 10) * 1/10 = int(-1005) * 1/10 = -100.5
   => we start correctly with the -100.5 sample
Case floor:
  lng_min = floor(-100.5 * 10) * 1/10 = -100.5  
   => same result

Lets now take real_lng_min = -100.55. The samples shall still be -100.5, -100.6, ...
Case int:
  lng_min = int(-100.55*10) * 1/10 = int(-1005.5)  * 1/10 = -1005   *1/10. = -100.5
  => we start from the correct sample
Case floor
  lng_min = floor(100.55*10)*1/10 = floor(-1005.5) * 1/10 = -1006 * 1/10. = -100.6
 => we start with -100.6 which is obviously outside

sbdt-google:
Why do you say so ? It should not.
If the polygoin defines a hole, then the MultiPoint intersected with rthe Polygon will only return the MultiPoint with the Point inside the Polygon (hole being considered).

If there is no unit test on this, I will add one.

arhannan:
we are using a unary_union of points returned by GridPolygon

src/harness/reference_models/geo/utils.py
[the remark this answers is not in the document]
arhannan: Thanks for the clarifications

sbdt-google:
I think this is because in PAT test, you did not take into account the requirement that the routines takes a GeoJSON geometry, and not a list of coordinates (you are passing an array of coordinates).

Moreover your GeoJSON for PPA are incorrectly specified (invalid GeoJSON).

I will change this PR slightly to make a test that the input dict is a correct GeoJSON geometry.

sbdt-google:
Ariful, I updated the code in the PR #449 : 
 - new unit test for case with holes
 - make the routine more robust against bad input.
Can you check the PR there, and I can submit right away,

Issue #398, opened 2018-01-22
Injected ESC data is not quite correct


kate-harrison:
According to WinnForum requirements and SAS admin agreements, we will use the installationParam object from SAS-CBSD 2.0. In particular, the field `azimuthRadiationPattern` will be required for SAS-SAS exchange of ESC sensors and the `antenna*` fields are no longer needed (in this use case).

The existing [ESC sensor record](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/testdata/esc_sensor_record_0.json) is using installationParam from SAS-CBSD 1.* and hence is not consistent with the latest SAS admin agreements.

PR #444, opened 2018-02-27
Use the installationParam object from SAS-CBSD 2.0

harinineelu:
Refer: https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues/398


harinineelu:
Thanks for reviewing, Janet :) 

Regarding the elevationRadiationPattern, have not added that since it is optional and can be added later if needed.

janetliuboston:
Sounds good then! Thanks, Harini!

harinineelu:
Merged #444

PR #445, opened 2018-02-27
Fix issue with POST/GET in sas.py

gilschieber:


gilschieber:
Fixing issue introduced in #414

PR #446, opened 2018-02-27
Remove all test cases which use the old IDs

harinineelu:
Remove all test cases which use the old IDs (and old test spec).


janetliuboston:
Looks good. Thanks.

harinineelu:
Merged #446

PR #450, opened 2018-02-28
NEW SSS test cases with new ID: 1 to 5

gilschieber:


src/harness/testcases/WINNF_FT_S_SSS_testcase.py
```diff
@@ -18,10 +18,67 @@
 
 SAS_CERT = os.path.join('certs', 'sas.cert')
 SAS_KEY = os.path.join('certs', 'sas.key')
+SAS_URL = 'https://fake.sas.url.not.used.org'
```
jni2000: The SAS_URL should also include the version number of the SAS.

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
```diff
@@ -18,10 +18,67 @@
...
 
+  @winnforum_testcase
+  def test_WINNF_FT_S_SSS_1(self):
+    """New registration with TLS_RSA_WITH_AES_128_GCM_SHA256 cipher.
+
+    Checks that SAS UUT response satisfy cipher security conditions.
+    Checks that a SAS registration with this configuration succeed.
+    """
+    self.doSasTestCipher('AES128-GCM-SHA256',
+                         client_cert=SAS_CERT, client_key=SAS_KEY,
```
jni2000: Should we make the cert and key as configurable parameters instead of global variables?

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
```diff
@@ -18,10 +18,67 @@
...
+                         client_url=SAS_TH_URL)
```
jni2000: Should we make the CA certificate chain also as one of the arguments?

src/harness/security_testcase.py
```diff
@@ -171,6 +172,29 @@ def doTestCipher(self, cipher, client_cert=None, client_key=None):
...
+                                   client_cert, client_key)
+
+    # Does a regular SAS registration
+    self.SasReset()
+    certificate_hash = getCertificateFingerprint(client_cert)
+    self._sas_admin.InjectPeerSas({'certificateHash': certificate_hash,
+                                   'url': client_url})
+    self._sas_admin.TriggerFullActivityDump()
+    with CiphersOverload(self._sas, [cipher], client_cert, client_key):
+      self._sas.GetFullActivityDump(client_cert, client_key)
```
jni2000: I think we should ensure that the GetFullActivityDump is successful. If not it should assert error should be raised.

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
gilschieber: Done.

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
gilschieber: As we don't expect to run these tests with different set of certificates, I think there is no point to use a configurable test case here.

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
gilschieber: The CA certificate chain is only related to the server, so I think it should not change across testcases (in particular these ones where we test only cipher).

src/harness/security_testcase.py
gilschieber: You're right. But this is already done.
The GetFullActivityDump calls _SasRequest that calls_RequestGet, and inside this function there is an 
assert http_code == 200, http_code
This should be enough I think.

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
gilschieber: Another point is that these tests are checking the algorithm, not the certs, so we can use the default certs the same way we use them for every other test. Let me know your thoughts so that we can go on with this one. Thanks.

src/harness/fake_sas.py
```diff
@@ -202,6 +202,30 @@ def GetEscSensorRecord(self, request, ssl_cert=None, ssl_key=None):
       # Return Empty if invalid Id
       return {}
 
+  def GetFullActivityDump(self, version, ssl_cert=None, ssl_key=None):
```
jni2000: This test case looks like copied from the content of FAD.1 PR. The below comments are related to those portion. Please use for reference:
1. The parameters ssl_cert=None, ssl_key=None arguments doesm't seem to be used
2. Please add some docstring description for this function to help understand the usage of this function.

src/harness/fake_sas.py
```diff
@@ -202,6 +202,30 @@ def GetEscSensorRecord(self, request, ssl_cert=None, ssl_key=None):
...
+    response = json.loads(json.dumps({
+        'files': [
+            {'url': "https://raw.githubusercontent.com/Wireless-Innovation-Forum/\
+             Spectrum-Access-System/master/schema/empty_activity_dump_file.json",
```
jni2000: Alignment: one space in the lines 209, 213,217 & 221 should be removed to align correctly.

src/harness/security_testcase.py
```diff
@@ -171,6 +172,29 @@ def doTestCipher(self, cipher, client_cert=None, client_key=None):
...
       self.assertRegistered([device_a])
 
+  def doSasTestCipher(self, cipher, client_cert, client_key, client_url):
+    """Does a cipher test as described in SSS tests 1 to 5 specification.
+
+    Args:
+      cipher: the cipher openSSL string name to test.
+      client_cert: SAS client certificate file in PEM format to use.
+      client_key: associated key file in PEM format to use.
+      client_url: base URL of the peer client SAS.
```
jni2000: The test harness would be the client SAS right? May be ‘client SAS’ can be renamed to ‘peer SAS’ or ‘SAS UUT’?

src/harness/security_testcase.py
```diff
@@ -171,6 +172,29 @@ def doTestCipher(self, cipher, client_cert=None, client_key=None):
     with CiphersOverload(self._sas, [cipher], client_cert, client_key):
       self.assertRegistered([device_a])
 
+  def doSasTestCipher(self, cipher, client_cert, client_key, client_url):
+    """Does a cipher test as described in SSS tests 1 to 5 specification.
+
+    Args:
+      cipher: the cipher openSSL string name to test.
+      client_cert: SAS client certificate file in PEM format to use.
```
jni2000: Please update the description of client_cert and client_key. They are actually path to the file and not the contents of the file in PEM format right?

src/harness/security_testcase.py
```diff
@@ -171,6 +172,29 @@ def doTestCipher(self, cipher, client_cert=None, client_key=None):
...
+      client_key: associated key file in PEM format to use.
+      client_url: base URL of the peer client SAS.
+    """
+
+    # Using pyOpenSSL low level API, does the SAS UUT server TLS session checks.
+    self.assertTlsHandshakeSucceed(self._sas_admin._base_url, [cipher],
```
jni2000: I think this method assertTlsHandshakeSucceed has to be invoked with the client_url. You could also consider building the SAS_TH_URL from the sas.cfg . Both the baseurl and the SAS version are configured in that file.

src/harness/security_testcase.py
jni2000: This method GetFullActivityDump would also not use the client_url. I think we should either modify these methods take the URL as input or have the functions use the sas.cfg file and not have the SAS_TH_URL defined and passed as an argument from the test case

src/harness/fake_sas.py
gilschieber: 1. This is completely consistent with all of the other functions in this file. If this is wrong, let's address in a separate PR which cleans up all functions.
2. Not needed in fake_sas.py, the sas_interface.py contains the corresponding documentation.

src/harness/fake_sas.py
gilschieber: Alignment corrected.

gilschieber:
Comments addressed, can you take a look?

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
jni2000: we do not hold strong position on this comment as long as the test code servers the requirements. If in case we need them be configurable, we can patch fixes.

src/harness/testcases/WINNF_FT_S_SSS_testcase.py
jni2000: ok, agreed.

src/harness/security_testcase.py
jni2000: ok, thanks.

src/harness/fake_sas.py
jni2000: ok, we can do as what you suggested later.

jni2000:
Looks good to us. If others agree, we approve this PR.

src/harness/security_testcase.py
jni2000: Norbert,

We think this issue still needs to be fixed in the SSS test cases

The TLS handshake and GetFullActivityDump is done on URLs ._sas_admin._base_url). GetFullActivityDump and _sas._base_url respectively. But he URL (client_url) that is injected using inject peer sas is completely different. 

The client url is set to https://fake.sas.url.not.used.org/v1.2 while the ._sas_admin._base_url and ._sas._base_url are based on the sas.cfg file and use localhost:9000

Please note that all the security test cases including the ones that we submitted would need the fix. I think the fix is to not use the global variable “SAS_URL” but to automatically construct the URL based on the sas.cfg.

Perhaps if other reviewers agree then we can do one patch and fix all the SSS test cases after everything is merged.

PR #454, opened 2018-03-04
remove duplicate getCertificateFingerprint

jni2000:
remove duplicate getCertificateFingerprint


jni2000:
Norbert/Kate

We found that there are two duplicate getCertificateFingerprint here, probably caused by one of our merge. We are fixing this now. Please review. Thanks.

jni2000:
thanks, for reviewing.

PR #455, opened 2018-03-05
Change device_c height type from AMSL to AGL

RedaHossino:


kate-harrison:
Just curious, what's the reason for this change? I think we originally tried to have a mix of AGL and AMSL so that the WinnForum default configs would be at least slightly more interesting.

RedaHossino:
Actually the reason of this change is that the landcover in the location of this CBSD is ~= 82m, so the height(4.5m) will not be valid.
if we need to keep the AMSL type we should change the height to be haat < 6 m, right?

kate-harrison:
Isn't the HAAT large but negative (and thus definitely < 6 m) in that case?

PR #459, opened 2018-03-06
Fix palRecords in GRA.17 - config.

harinineelu:
Also updated the copyright year and ordered the test cases.


harinineelu:
Merged #459

PR #461, opened 2018-03-07
adding "cbsd/" prefix in generateCbsdReferenceId()

jni2000:
adding "cbsd/" prefix in generateCbsdReferenceId() helper function per reviewer's suggestion.


jni2000:
thanks.

PR #463, opened 2018-03-08
Added Exception for PPA Census Tract MultiPolygon

sushantparkhi:


src/harness/reference_models/ppa/ppa.py
```diff
@@ -177,6 +177,12 @@ def PpaCreationModel(devices, pal_records):
   # after Census Tract Clipping
   contour_union = ops.cascaded_union(device_polygon)
   ppa_polygon = _ClipPpaByCensusTract(contour_union, pal_records)
+
+  if ppa_polygon.is_empty:
+    raise Exception("Empty Polygon is generated, please check the inputs.")
+  if ppa_polygon.geom_type == "MultiPolygon":
```
sbdt-google: Why isnt it supported ? My understanding is that you can easily come up wih MultiPolygons, and this should not be an issue. If it is not supported down the line, we should fix there instead.

sbdt-google: Offline discussion has answered this question: the protocol spec does not support multipolygon although it should be easy to support.

sushantparkhi:
@sbdt-google Sorry missed the json.dumps part while pushing the code. I've added that and updated the Readme.md.

sushantparkhi:
@sbdt-google @kate-harrison Can you re-approve this PR with the Latest Change, then I will merge this to master?

PR #476, opened 2018-03-15
Misc cleanup and unit tests

sbdt-google:
This is a cleanup PR with no impact to  current modules. 
List of changes: 
 - provides a fake propagation model to enable more advanced unit testing on ref models
 - create as an example a unit test for the PPA module
 - centralize all global drivers instance into one single geo/drive.py module (instead of being spread in wf_itm.py and ppa.py)
 - support compilation propagation models in MacOs
 - Fix non relative path in antenna_test.py module
 - Add official NLCD land cover codes reference
 - update README to reflect some of the recent devlopments


jni2000:
will do today..

jni2000:
I review the parts that I am that I understand, looks good. More review is in progress.

src/harness/reference_models/examples/benchmark/time_dpa.py
```diff
@@ -0,0 +1,257 @@
...
+Notes:
+  - multiprocessing facility not tested on Windows (use 1 process if issues)
+  - if warning reported on cached tiles swapping, increase the cache size.
+"""
+# TODO(sbdt): add reporting for memory usage
+from collections import namedtuple
+import multiprocessing
+import time
+
+import cartopy.crs as ccrs
```
sushantparkhi: Maybe we need to add this and matplotlib to README.md?

src/harness/reference_models/examples/benchmark/time_dpa.py
```diff
@@ -0,0 +1,257 @@
+"""Benchmark the DPA model timing.
```
sushantparkhi: Winnforum Docstring?

src/harness/reference_models/examples/benchmark/time_dpa.py
```diff
@@ -0,0 +1,257 @@
...
+from collections import namedtuple
+import multiprocessing
+import time
+
+import cartopy.crs as ccrs
+import matplotlib.pyplot as plt
+import numpy as np
+
+from reference_models.move_list_calc import move_list
+from reference_models.geo import zones
```
sushantparkhi: Somehow I am not able to find zone.py in this branch? Am I missing something?

src/harness/reference_models/examples/benchmark/time_dpa.py
```diff
@@ -0,0 +1,257 @@
...
+do_inside_urban_area = True
+
+# - Ratio of cat B and different catA
+ratio_cat_b = 0.2
+ratio_cat_a_indoor = 0.4
+
+# - Simulation setup
+#   + Number of parallel processes -
+#     do not exceed your workstation parallel capabilities (number of core -2)
+num_processes = 8
```
sushantparkhi: We can use `multiprocessing.cpu_count()` to set the default value here and subtract it by 2, as suggested in the comments?

src/harness/reference_models/examples/benchmark/time_dpa.py
sbdt-google: No it is missing from this PR.
Let me add them now, and address your other comments.

src/harness/reference_models/examples/benchmark/time_prop.py
```diff
@@ -0,0 +1,90 @@
...
+  res_last = CalcHybrid(0.5, rx_lats[-1], rx_lngs[-1])
+  res_first = CalcHybrid(0.5, rx_lats[-1], rx_lngs[-1])
+  print '## Initial driver report'
+  drive.terrain_driver.stats.Report()
+
+  print ''
+  print '# Propagation calculation time at various distances'
+  for k, dist in enumerate(distances):
+    # ITM
+    times = timeit.repeat(setup = 'import time_prop',
```
sushantparkhi: Should we have a benchmark for ITM with multiple reliabilities?

src/harness/reference_models/examples/entities.py
```diff
@@ -116,24 +123,92 @@ def GenerateCbsdList(n_cbsd, template_cbsd,
...
+  }
+
+
+def GetCbsdGrantRequest(cbsd, min_freq_mhz, max_freq_mhz):
+  """Returns a CBSD grant request from a |Cbsd| object and a min/max frequency.
+  """
+  return {
+      'operationParam': {
+          'operationFrequencyRange': {
+              'lowFrequency': min_freq_mhz * 1e6,
```
sushantparkhi: I think we need to typecast this to int as multiplying it with 1e6 will generate decimal.

src/harness/reference_models/geo/tiles.py
```diff
@@ -36,24 +36,26 @@ def UpdateForTileLoad(self, ilat, ilon):
...
     print "  Avg:{avg} (std={std})".format(
-        avg=np.mean(count_per_tile), std=np.std(count_per_tile))
+        avg=np.mean(counts), std=np.std(counts))
     print "  Min:{min} Max:{max}".format(
-        min=np.min(count_per_tile), max=np.max(count_per_tile))
+        min=np.min(counts), max=np.max(counts))
+
 
 NED_TILES = set([
     ( 6, 162), (44, -81), (67,-165),
```
sushantparkhi: Maybe removing 2 Spaces from here.

src/harness/reference_models/move_list_calc/move_list.py
```diff
@@ -44,14 +44,10 @@
 
 # Import WINNF reference models including propagation, geo, and CBSD antenna gain models
 from reference_models.propagation import wf_itm
-# from reference_models.geo import terrain
```
sushantparkhi: I think iadd is not used anywhere in the file.

src/harness/reference_models/examples/benchmark/time_dpa.py
sbdt-google: This is a "benchmarking code", so the idea is to run it in some environment with
precise parameters. And extrapolate from that.
What i can do, is that value -1 does what you suggest. But instead of 50% of the resources, I would say 75% ?

src/harness/reference_models/examples/benchmark/time_prop.py
sbdt-google: In fact here I have put the average, which means that it does 100 reliabilities internally.

All other more advanced benchmark, we can add later. For your information, each additional reliability calculation takes 0.01% of the time. So:
 -  going from 100 to 1000 reliabilities adds about 10% to the overall curve.
 -  going from 100 to 1 (the median for example) saves about 1% of the time.

src/harness/reference_models/examples/benchmark/time_dpa.py
sbdt-google: Done

src/harness/reference_models/examples/benchmark/time_dpa.py
sbdt-google: Done

src/harness/reference_models/examples/benchmark/time_dpa.py
sbdt-google: It was done on another branch to keep this one only focused.
But you are right, better to add it here. You should see it now

src/harness/reference_models/examples/entities.py
sbdt-google: Yes it is decimal, I dont think the reference models cares, but for sake of generality if we use this purely test code to test/benchmark other pieces which care about that.

src/harness/reference_models/geo/tiles.py
sbdt-google: Sorry, I did not understand which space you mention.

sbdt-google:
Sushant, PTAL I think I have addressed all your comments

src/harness/reference_models/examples/benchmark/time_prop.py
sushantparkhi: I missed the -1 flag!

src/harness/reference_models/geo/tiles.py
sushantparkhi: Here On Line 61 and below it is using 4 Spaces instead of 2 Spaces for Indentation.

sushantparkhi:
Looks good to me!

jni2000:
Thanks, Sushant.

sbdt-google:
Thanks for the quick turn around

PR #296, opened 2017-10-06
NEW FAD 1, Full Activities Dump Test Case

RedaHossino:

PR #477, opened 2018-03-15
Add polygon comparison routines.

sbdt-google:
To be used for PPA comparison according to eq 8.3.1 of test spec.


sbdt-google:
Thanks for the quick review. Just submitted, so you can use it for your PR #296

PR #478, opened 2018-03-15
Fix dpaId - Added an underscore (Eg: 'east_dpa_6')

harinineelu:


jni2000:
Can you please also modify the similar DPA ids in the following files ? We changed them from "east_dpa_1" format to "esst_dpa1" to be in sync with what was used in the test cases you modified here -- asked by reviewers. Doing this all together is better to keep them in sync. Thanks.

testcases/WINNF_FT_S_GRA_testcase.py:   'highFrequency': pal_high_frequency },'dpaId': 'east_dpa4'})
testcases/WINNF_FT_S_HBT_testcase.py:   'dpaId': 'east_dpa6',
testcases/WINNF_FT_S_SIQ_testcase.py:    'dpaId':'east_dpa4'})

harinineelu:
Merged #478

PR #486, opened 2018-03-22
updated the FSS record data with additional information from the FCC

jni2000:
updated the FSS record data with additional information from the FCC FSS database for supporting FSS purging list reference model, this will also be used in FSS protection test cases.


jni2000:
Kate and Javier

Per your suggestion in the review comments to the FSS purge list reference model PR, we update the FSS data record to make it in sync for all the test cases. Please review. Thanks.

src/harness/testcases/testdata/fss_record_0.json
```diff
@@ -8,18 +8,21 @@
...
+        "antennaElevationAngle":39
       },
       "operationParam": {
         "operationFrequencyRange": {
           "lowFrequency": 3650000000,
           "highFrequency": 4200000000
         }
-      }
+      },
+      "ttc": true
```
kate-harrison: I think this needs to be moved to a higher level since we cannot change the IncumbentProtectData object from SAS-SAS v1.0. Please also make sure that the other fields use names that are consistent with the definition of that object, or are defined outside the object.

jni2000:
Thanks, Kate. Reasonable. Which level would you suggest we move it to ? We will make sure the consistency of all the fields.

James Ni

kate-harrison:
I think it has to be moved to the same level as "record" based on the way the object is defined (please look at our SAS testing API doc).

jni2000:
Per today's discussion, we made the following changes
1. removed the "elevation" field -- this is actually the site base elevation from the original FSS database data. Maybe needed and will add in when needed.
2. using "antennaDowntilt" instead of "antennaElevationAngle" to match up with WinnF spec.
3. moved "ttc" up to the record level to avoid changing the IncumbentProtectData object already defined.

src/harness/testcases/testdata/fss_record_0.json
```diff
@@ -8,18 +8,21 @@
         "latitude": 39.2291,
         "longitude":  -100.1,
         "height": 7.6,
+        "elevation": 100,
```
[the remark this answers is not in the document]
sbdt-google: 1. Note that we cannot use the `elevation` information, as it might be completely inconsistent with the terrain database elevation, and then it will break all propagation calculation. We shall use only the height above terrain and use the terrain from the database. Anyway the prop model do not have ability to make sense of such extra 'elevation'.
2. I though the FSS record is defined in the v1.0 of the SAS-SAS database and that we used that for defining the interface, even if it is no more defined in newer SAS-SAS database. 
On that doc, the FSS has been modeled with installationParams similar to the CBSD one. This is why we use antennaDowntilt, and not antennaElevationAngle.

I let @kate-harrison  comment further on that fact that we use the format defined for CBSD for the FSS record. So all non compliant info shall be removed, once this fact is well established among us.

sbdt-google:
Ok, I did not see your last message when writing my first comment.
Please kate @kate-harrison, have a final review on this item if it is ok for you now,

src/harness/testcases/testdata/fss_record_0.json
```diff
@@ -21,5 +21,6 @@
         }
       }
     }
-    ]
+  ],
+  "ttc": true
```
kate-harrison: I think this needs to go one level higher, as discussed in today's call.

jni2000: ok, done, please take a look. We will have to add "incumbentProtectionData" tag for this.

kate-harrison: I think you have misread the spec; please take another look.

Also note that this change will require changing any code which uses this file.

jni2000: the "incumbentProtectionData" defined in the spec. contains the "id", "type", "deploymentParam" etc. which is the contents of the current top level in the data, without adding this "incumbentProtectionData" tag, it will not be a valid JSON. Can you please be more specific how it shall be. Agreed that we need to modify other codes using this data. Will do after this is finalized.

kate-harrison:
The testing API is quite clear about the name of this field. The type is "IncumbentProtectionData" but the name of the field is not.

jni2000:
So if we do as the test API modified today as below, we do not need to make these changes any more, as ttc will be an additional input parameter. If this is what you mean, I will close this PR.

>**FSS**
URL (HTTP POST): $BASE_URL/admin/injectdata/fss
Input:
IncumbentProtectionData record (from SAS-SAS v1.0)
boolean ttc (True if and only if the FSS is of type TT&C)
Output: none

kate-harrison:
What I meant is that the file is not consistent with the API that you pasted in your previous comment. This PR should update the file (and corresponding code) to match the new API definition.

jni2000:
ok here is my proposal
1. modify the data as I proposed here
2. keep the FSS injection API as we defined today with the new "ttc" parameter addition
3. make necessary changes in test codes to capture the data structure change

This the minimum impacting changes I can see. Let me know if you think otherwise.

kate-harrison:
My point is that (1) and (2) are conflicting with each other. The proposal in this CR does not match the FSS injection API as we defined today.

jni2000:
I was thinking
1. use the new data record definition to keep all necessary information.
2. split the new data record and feed the separated "IncumbentProtectionData" and "ttc" into the injection API as we defined today, this will minimize the changes on the SAS UUT side.
3. we do all necessary changes on the test codes side to match up with the new data record definition.

I think this would work.

kate-harrison:
I'm not sure what you mean by (2), or how it would minimize the changes to SAS UUT.

I thought the implementation was very straightforward based on the API change we documented today, and I'm having a hard time understanding the confusion.

jni2000:
My confusion is that if we follow the API we defined today as below, we will have to feed the API with two separated input parameters. While if we modify the fss_record as a single JSON object as is now, we will have to do what I mentioned earlier. 

>**FSS**
URL (HTTP POST): $BASE_URL/admin/injectdata/fss
Input:
IncumbentProtectionData record (from SAS-SAS v1.0)
boolean ttc (True if and only if the FSS is of type TT&C)
Output: none

Or do you mean that we should change the injection API as below?

>**FSS**
URL (HTTP POST): $BASE_URL/admin/injectdata/fss
Input:
fss_data record as is now, i.e. {"incombentProtectionData": {....}, "ttc": true }
Output: none

If we are still not inline, please state explicitly what is your suggestion. Thanks

jni2000:
Changed the FSS record as below
``{
  "record": {
    "id": "incumbent/ibfs/0",
    "type": "FSS",
    "deploymentParam":
    [
      {
        "installationParam": {
          "latitude": 39.2291,
          "longitude":  -100.1,
          "height": 7.6,
          "heightType": "AGL",
          "antennaGain": 16,
          "antennaBeamwidth": 30,
          "antennaAzimuth": 270,
          "antennaDowntilt": -90
        },
        "operationParam": {
          "operationFrequencyRange": {
            "lowFrequency": 3650000000,
            "highFrequency": 4200000000
          }
        }
      }
    ]
  },
  "ttc": true
}

Also modified the following test cases that use the FSS record data.

        modified:   src/harness/testcases/WINNF_FT_S_HBT_testcase.py
        modified:   src/harness/testcases/WINNF_FT_S_PAT_testcase.py
        modified:   src/harness/testcases/WINNF_FT_S_SIQ_testcase.py

Did a rebase to catch up newest master codes. Please review. Thanks

src/harness/testcases/WINNF_FT_S_PAT_testcase.py
```diff
@@ -145,8 +145,9 @@ def generate_FT_S_PAT_default_config(self, filename):
     open(os.path.join('testcases', 'testdata', 'device_a.json')))
 
     # load FSS
-    fss_record_0 = json.load(
+    fss_data = json.load(
```
kate-harrison: I guess the TT&C information doesn't need to be used in this test case, so it's OK to discard it. (No change needed.)

jni2000:
thanks.

PR #487, opened 2018-03-23
Improve geo.utils module.

sbdt-google:
- Add winding check function
- Add winding correction function
- Add better conversion routines to geojson and shapely.
- Fix small bug in PolygonsAlmostEqual routine for dict input


sbdt-google:
Thanks, I will submit then and if others have comment we can always make some fixes later.

PR #489, opened 2018-03-26
initial submission of xPR test cases for review

jni2000:
Please review the initial submission of xPR.1


kate-harrison:
The xPR test cases are templates for other test cases, not test cases themselves. Is there value to having a separate implementation instead of calling MCP.1 directly?

jni2000:
Discussed with our team, we agreed with what Kate commented. We do not need to explicitly implement separate xPR.1 and xPR,2 but implementing the fully configurable MCP.1. We will incorporate the codes into MCP.1. So this PR is not needed and thus closed.

PR #490, opened 2018-03-27
small update related to wait for DPA step

RedaHossino:
# remove unapproved comment 
# update time value of wait for DPA
# update step number


src/harness/testcases/WINNF_FT_S_HBT_testcase.py
```diff
@@ -1065,11 +1065,10 @@ def test_WINNF_FT_S_HBT_12(self):
     if response['response']['responseCode'] == 500:
       return
 
-    # Sleep 240 seconds
-    time.sleep(240)
+    # Step 8
+    time.sleep(300)
```
kate-harrison: Should the spec say 240? (I agree that it currently says 300, and hence this change is correct.)

kate-harrison: *shrugs* I guess? I don't have a strong opinion :)

PR #494, opened 2018-03-28
split the interference module out of IAP per reviewer's suggestion

jni2000:
We split the interference module out of IAP PR as suggested by reviewers. This module is used in FSS purge list reference model and IAP reference model. Please review, Thanks.

### DIFF
--- a/src/harness/reference_models/interference/interference.py
+++ b/src/harness/reference_models/interference/interference.py
@@ -178,8 +178,7 @@ class AggregateInterferenceOutputFormat:
     self.lock = multiprocessing.Lock()
 
   def SetAggregateInterferenceInfo(self, latitude, longitude, interference):
-    self.lock.acquire()
-    try:
+    with self.lock():
       if latitude not in self.aggregate_interference_info:
         self.aggregate_interference_info[latitude] = self.manager.dict()
    
@@ -191,10 +190,7 @@ class AggregateInterferenceOutputFormat:
 
       aggregate_interference_proxy[longitude].append(interference)
       self.aggregate_interference_info[latitude] = aggregate_interference_proxy
-
-    finally:
-      self.lock.release()
-
+      
   def GetAggregateInterferenceInfo(self):
     return self.aggregate_interference_info
 
--- a/src/harness/reference_models/interference/interference.py
+++ b/src/harness/reference_models/interference/interference.py
@@ -88,7 +88,7 @@ ESC_CAT_A_HIGH_FREQ_HZ = 3660.e6
 ESC_LOW_FREQ_HZ = 3550.e6
 ESC_HIGH_FREQ_HZ = 3680.e6
 
-# ESC Channel 21 Center Frequency
+# ESC Channel 21 Centre Frequency
 ESC_CH21_CF_HZ = 36525.e5
 
 # One Mega Hertz
@@ -163,7 +163,7 @@ EscInformation = namedtuple('EscInformation',
 
 def dbToLinear(x):
   """This function returns dBm to mW converted value"""
-  return 10**(x / 10)
+  return 10**(x /float(10))
 
 def linearToDb(x):
   """This function returns mW to dBm converted value"""
@@ -183,7 +183,6 @@ def getProtectedChannels(low_freq_hz, high_freq_hz):
     (low_freq_hz,high_freq_hz). 
   """
   assert low_freq_hz < high_freq_hz, 'Low frequency is greater than high frequency'
-
   channels = np.arange( max(low_freq_hz, 3550*MHZ), min(high_freq_hz, 3700*MHZ), 5*MHZ)
 
   return [(low, high) for low,high in zip(channels, channels+5*MHZ)]
@@ -193,7 +192,7 @@ def findGrantsInsideNeighborhood(grants, protection_point, entity_type):
   """Finds grants inside protection entity neighborhood
 
   Args:
-    grants: a list of grants of type GAA
+    grants: an iterator of grants with attributes latitude and longitude
     protection_point: tuple containing (latitude,longitude) of a protection entity 
     entity_type: enum of type ProtectedEntityType 
   Returns:
@@ -218,13 +217,13 @@ def findGrantsInsideNeighborhood(grants, protection_point, entity_type):
 
 
 def findOverlappingGrants(grants, constraint):
-  """Finds grants inside protection entity neighborhood
+  """Finds grants overlapping with protection entity
    
   Grants overlapping with frequency range of the protection entity are 
   considered as overlapping grants 
 
   Args:
-    grants: a list of grants of type GAA
+    grants: an iterator of grants with attributes latitude and longitude
     constraint: protection constraint of type ProtectionConstraint
   Returns:
     grants_inside: a list of grants, each one being a namedtuple of type
@@ -245,9 +244,10 @@ def findOverlappingGrants(grants, constraint):
     # ESC Passband is 3550-3680MHz
     # Category A CBSD grants are considered in the neighborhood only for 
     # constraint frequency range 3550-3660MHz
-    if ProtectedEntityType.ESC == _DISTANCE_PER_PROTECTION_TYPE[constraint.entity_type]:
-      if grant.cbsd_category == 'A' and constraint.high_frequency > ESC_CAT_A_HIGH_FREQ_HZ:
-        freq_check = False
+    if (constraint.entity_type == ProtectedEntityType.ESC and
+          grant.cbsd_category == 'A' and 
+          constraint.high_frequency > ESC_CAT_A_HIGH_FREQ_HZ):
+          freq_check = False
 
     # Append the grants information if it is inside the neighborhood of
     # protection constraint
@@ -322,10 +322,6 @@ def getAllGrantInformationFromCbsdDataDump(cbsd_data_records, is_managing_sas=Tr
       altitude_cbsd = terrainDriver.GetTerrainElevation(lat_cbsd, lon_cbsd)
       height_cbsd = height_cbsd - altitude_cbsd
 
-    # Sanity check on CBSD antenna height
-    if height_cbsd < 1 or height_cbsd > 1000:
-      raise ValueError('CBSD height is less than 1m or greater than 1000m.')
-
     for grant in grants:
       # Return CBSD information
       cbsd_grant = CbsdGrantInformation(
@@ -370,8 +366,8 @@ def computeInterferencePpaGwpzPoint(cbsd_grant, constraint, h_inc_ant,
     interference: interference contribution(dBm)
   """
 
-  if (cbsd_grant.latitude == constraint.latitude) and \
-    (cbsd_grant.longitude == constraint.longitude): 
+  if (cbsd_grant.latitude == constraint.latitude and
+        cbsd_grant.longitude == constraint.longitude): 
     db_loss = 0
     incidence_angles = wf_itm._IncidenceAngles(hor_cbsd=0, ver_cbsd=0, hor_rx=0, ver_rx=0)
   else:
@@ -452,8 +448,7 @@ def computeInterferenceFssCochannel(cbsd_grant, constraint, fss_info, max_eirp):
   Args:
     cbsd_grant: a namedtuple of type CbsdGrantInformation
     constraint: protection constraint of type ProtectionConstraint
-    fss_info: contains information on antenna height the tangent 
-              and perpendicular components.
+    fss_info: contains fss point and antenna information
     max_eirp: The maximum EIRP allocated to the grant during IAP procedure
   Returns:
     interference: interference contribution(dBm)
@@ -499,20 +494,20 @@ def getFssMaskLoss(cbsd_grant, constraint):
 
   # if lower edge of the FSS passband is less than CBSD grant
   # lowFrequency and highFrequency
-  if constraint.low_frequency < cbsd_grant.low_frequency and\
-         constraint.low_frequency < cbsd_grant.high_frequency:
+  if (constraint.low_frequency < cbsd_grant.low_frequency and
+         constraint.low_frequency < cbsd_grant.high_frequency):
     fss_mask_loss = 0.5 
 
   # if CBSD grant lowFrequency and highFrequency is less than
   # 50MHz offset from the FSS passband lower edge
-  elif cbsd_grant.low_frequency < offset and\
-     cbsd_grant.high_frequency < offset:
+  elif (cbsd_grant.low_frequency < offset and
+     cbsd_grant.high_frequency < offset):
     fss_mask_loss = linearToDb((cbsd_freq_range / MHZ) * 0.25)
 
   # if CBSD grant lowFrequency is less than 50MHz offset and
   # highFrequency is greater than 50MHz offset
-  elif cbsd_grant.low_frequency < offset and\
-            cbsd_grant.high_frequency > offset:
+  elif (cbsd_grant.low_frequency < offset and
+            cbsd_grant.high_frequency > offset):
     low_freq_mask_loss = linearToDb(((offset - cbsd_grant.low_frequency) /
                                                      MHZ) * 0.25)
     fss_mask_loss = low_freq_mask_loss + linearToDb(((cbsd_grant.high_frequency - offset) / 
@@ -521,10 +516,10 @@ def getFssMaskLoss(cbsd_grant, constraint):
   # if FSS Passband lower edge frequency is grater than CBSD grant
   # lowFrequency and highFrequency and
   # CBSD grand low and high frequencies are greater than 50MHz offset
-  elif constraint.low_frequency > cbsd_grant.low_frequency and \
-      constraint.low_frequency > cbsd_grant.high_frequency and \
-      cbsd_grant.low_frequency > offset and\
-             cbsd_grant.high_frequency > offset:
+  elif (constraint.low_frequency > cbsd_grant.low_frequency and
+      constraint.low_frequency > cbsd_grant.high_frequency and 
+      cbsd_grant.low_frequency > offset and
+             cbsd_grant.high_frequency > offset):
     fss_mask_loss = linearToDb((cbsd_freq_range / MHZ) * 0.6)
 
   return fss_mask_loss
@@ -539,8 +534,7 @@ def computeInterferenceFssBlocking(cbsd_grant, constraint, fss_info, max_eirp):
   Args:
     cbsd_grant: a namedtuple of type CbsdGrantInformation
     constraint: protection constraint of type ProtectionConstraint
-    fss_info: contains information on antenna height on 
-              the tangent and perpendicular components.
+    fss_info: contains fss point and antenna information 
     max_eirp: The maximum EIRP allocated to the grant during IAP procedure
   Returns:
     interference: interference contribution(dBm)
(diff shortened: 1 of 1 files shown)


src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+# ESC IAP for Out-of-Band Category A CBSDs in Frequency Range (Hz)
+ESC_CAT_A_LOW_FREQ_HZ = 3550.e6
+ESC_CAT_A_HIGH_FREQ_HZ = 3660.e6
+
+# ESC Passband Frequency Range (Hz)
+ESC_LOW_FREQ_HZ = 3550.e6
+ESC_HIGH_FREQ_HZ = 3680.e6
+
+# ESC Channel 21 Center Frequency
+ESC_CH21_CF_HZ = 36525.e5
```
sbdt-google: Typo

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+
+# Define ESC information, i.e., a tuple with named fields of
+# 'antenna_height', 'antenna_azimuth', 'antenna_gain', 'antenna_pattern_gain'
+EscInformation = namedtuple('EscInformation',
+                                 ['antenna_height', 'antenna_azimuth',
+                                  'antenna_gain', 'antenna_pattern_gain'])
+
+def dbToLinear(x):
+  """This function returns dBm to mW converted value"""
+  return 10**(x / 10)
```
sbdt-google: Better to put the last 10 as a float, to avoid issues if passing an int to the function:
```
10**(x / 10.)
```

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+  channels = np.arange( max(low_freq_hz, 3550*MHZ), min(high_freq_hz, 3700*MHZ), 5*MHZ)
+
+  return [(low, high) for low,high in zip(channels, channels+5*MHZ)]
+
+
+def findGrantsInsideNeighborhood(grants, protection_point, entity_type):
+  """Finds grants inside protection entity neighborhood
+
+  Args:
+    grants: a list of grants of type GAA
```
sbdt-google: "an iterablle of grants with attributes: latitude and longitude"

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+
+    # Check if CBSD is inside the neighborhood of protection constraint
+    if dist_km <= _DISTANCE_PER_PROTECTION_TYPE[entity_type][grant.cbsd_category == 'B']:
+      grants_inside.append(grant)
+
+  return grants_inside
+
+
+def findOverlappingGrants(grants, constraint):
+  """Finds grants inside protection entity neighborhood
```
sbdt-google: Please update the docstring description. It is identical to previous function, altthough it does things diferently.

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+  for grant in grants:
+    # Check frequency range
+    overlapping_bw = min(grant.high_frequency, constraint.high_frequency) \
+                        - max(grant.low_frequency, constraint.low_frequency)
+    freq_check = (overlapping_bw > 0)
+    
+    # ESC Passband is 3550-3680MHz
+    # Category A CBSD grants are considered in the neighborhood only for 
+    # constraint frequency range 3550-3660MHz
+    if ProtectedEntityType.ESC == _DISTANCE_PER_PROTECTION_TYPE[constraint.entity_type]:
```
sbdt-google: This line is wrong. Is it somethink like ?
```
if (constraint.entity_type == ProtectedEntityType.ESC and
     grant.cbsd_category == 'A' and 
    constraint.high_frequency > ESC_CAT_A_HIGH_FREQ_HZ):
  freq_check = False 
```
Also please double check the logic about freq check:
isnt it the low_frequency of the constraint which should be checked ?

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+# 'pointing_elevation' 
+FssProtectionPoint = namedtuple('FssProtectionPoint',
+                                   ['latitude', 'longitude',
+                                    'height_agl', 'max_gain_dbi',
+                                    'pointing_azimuth', 'pointing_elevation'])
+
+
+# Define ESC information, i.e., a tuple with named fields of
+# 'antenna_height', 'antenna_azimuth', 'antenna_gain', 'antenna_pattern_gain'
+EscInformation = namedtuple('EscInformation',
```
sbdt-google: Unused in this file.
But I guess these structures supposed to be used by other modules (IAP, ..) ??
Why not latitude/longitude like in the other protected point ?
Note: I might decide later to move the structure in a separate file for being reused by the DPA protection stuff. (nothing to do for now)

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+
+  Args:
+    low_freq_hz: Low frequency of the protected entity(Hz).
+    high_freq_hz: High frequency of the protected entity(Hz)
+  Returns:
+    An array of protected channel frequency range tuple 
+    (low_freq_hz,high_freq_hz). 
+  """
+  assert low_freq_hz < high_freq_hz, 'Low frequency is greater than high frequency'
+
```
sbdt-google: please remove useless space which do not add to the readability

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+    lat_cbsd = registration.get('installationParam', {}).get('latitude')
+    lon_cbsd = registration.get('installationParam', {}).get('longitude')
+    height_cbsd = registration.get('installationParam', {}).get('height')
+    height_type_cbsd = registration.get('installationParam', {}).get('heightType')
+    if height_type_cbsd == 'AMSL':
+      altitude_cbsd = terrainDriver.GetTerrainElevation(lat_cbsd, lon_cbsd)
+      height_cbsd = height_cbsd - altitude_cbsd
+
+    # Sanity check on CBSD antenna height
+    if height_cbsd < 1 or height_cbsd > 1000:
```
sbdt-google: This does not belong to this piece of code. 
It is currently done in propag models, but there has been agreement among SAS admins that all heights
below 1m (or 1.5m TBD) will be automtically reinitialized to 1m. This is to copde with the fact of errors in reporting
AGL height or inconsistencies in AMSL reported height vs terrain data.
So best is to remove this from here, and let the prop model handle this

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+    cbsd_grant: a namedtuple of type CbsdGrantInformation
+    constraint: protection constraint of type ProtectionConstraint
+    h_inc_ant: reference incumbent antenna height (in meters)
+    max_eirp: The maximum EIRP allocated to the grant during IAP procedure
+    region: Region type of the GWPZ or PPA area
+  Returns:
+    interference: interference contribution(dBm)
+  """
+
+  if (cbsd_grant.latitude == constraint.latitude) and \
```
sbdt-google: Cleaner to use the parenthesis for line continuation:
```
 if (cbsd_grant.latitude == constraint.latitude and
    cbsd_grant.longitude == constraint.longitude):
```

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+
+def computeInterferenceFssCochannel(cbsd_grant, constraint, fss_info, max_eirp):
+  """Compute interference grant causes to a FSS protection point
+  
+  Routine to compute interference neighborhood grant causes to FSS protection 
+  point for co-channel passband
+  Args:
+    cbsd_grant: a namedtuple of type CbsdGrantInformation
+    constraint: protection constraint of type ProtectionConstraint
+    fss_info: contains information on antenna height the tangent
```
sbdt-google: typo

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,648 @@
...
+  offset = constraint.low_frequency - 50.e6
+
+  # Get CBSD grant frequency range
+  cbsd_freq_range = cbsd_grant.high_frequency - cbsd_grant.low_frequency
+
+  fss_mask_loss = 0
+
+  # if lower edge of the FSS passband is less than CBSD grant
+  # lowFrequency and highFrequency
+  if constraint.low_frequency < cbsd_grant.low_frequency and\
```
sbdt-google: use parenthesis instead of \.
Here and below

src/harness/reference_models/interference/interference.py
jni2000: Corrected

src/harness/reference_models/interference/interference.py
jni2000: Updated   as per suggestion

src/harness/reference_models/interference/interference.py
jni2000: Updated   comment

src/harness/reference_models/interference/interference.py
jni2000: Updated   doc string

src/harness/reference_models/interference/interference.py
jni2000: Corrected   the logic. Low_frequency can be same for category A and category B devices.   High frequency is checked because category A high freqeuncy is 3660 whereas   caregory B high freqeuncy is 3680

src/harness/reference_models/interference/interference.py
jni2000: Yes,   EscInformation is used by IAP and Aggregate Interference modules. As   suggested, not changed anything for now

src/harness/reference_models/interference/interference.py
jni2000: Removed   space

src/harness/reference_models/interference/interference.py
jni2000: Removed   height check

src/harness/reference_models/interference/interference.py
jni2000: Updated   as per suggestion

src/harness/reference_models/interference/interference.py
jni2000: Corrected

src/harness/reference_models/interference/interference.py
jni2000: Updated   paranthesis

jni2000:
Surge, thanks for your comments. We have addressed them. Please take a look again and help us wrap this up as soon as possible to allow IAP, FSS purge etc. related work moving forward. Thanks.

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,676 @@
...
+
+  This class is used to generate specified output format for post IAP 
+  allowed interference or for aggregate interference to non-DPAs
+  
+  Creates a data structure to store post IAP allowed interference or 
+  aggregate interference to non-DPAs output in the format of 
+  {latitude : {longitude : [ output_first_5MHz_segment, output_first_5MHz_segment ]}}
+  """
+  def __init__(self):
+    self.manager = multiprocessing.Manager()
```
sbdt-google: You are using a Manager proxy object to be able to share the dict across process, so that I guess we fill it up in parallel.
That is a good idea in general, however I am not sure it will work in this case since we have a inner dict.
See for example: https://stackoverflow.com/questions/9436757/how-does-multiprocessing-manager-work-in-python

We need to do a test. But I guess we can keep this issue in mind, and fix it in a future PR so as not to block this PR.

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,676 @@
...
+    self.aggregate_interference_info = self.manager.dict()
+
+  def SetAggregateInterferenceInfo(self, latitude, longitude, interference):
+    if self.aggregate_interference_info.has_key(latitude) is False:
+      self.aggregate_interference_info[latitude] = {}
+   
+    # Creating a proxy container for mutable dictionary
+    aggregate_interference_proxy = self.aggregate_interference_info[latitude]
+
+    if aggregate_interference_proxy.has_key(longitude) is False:
```
sbdt-google: same as above. Please use the pythonic construct

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,676 @@
...
+      aggregate_interference_proxy[longitude] = []
+
+    aggregate_interference_proxy[longitude].append(interference)
```
sbdt-google: This is this part which may not work with the Manager issue I mentioned earlier

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,676 @@
...
+  Args:
+    grant : namedtuple of type CbsdGrantInformation
+    eirp : EIRP of the grant 
+    channel_constraint: namedtuple of type ProtectionConstraint
+    fss_info: contains fss point and antenna information
+    esc_antenna_info: contains information on ESC antenna height, azimuth, 
+                      gain and pattern gain 
+    region_type: region type of protection PPA/GWPZ area.
+  Returns:
+    interference: Interference caused by a grant(dBm)
```
sbdt-google: description is wrong. This is not dBm as you convert it to linear.

All these routines above compute interference in dBm. Why not returning also in dBm ?
The client code can then itself do the conversion in linear before adding things up. Which he will be able to do more efficiently since it will work on a bunch of them at once

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,641 @@
...
+    and grants
+  """
+  # List of CBSD grant tuples extracted from FAD record
+  grant_objects = []
+
+  # Creating list of cbsds
+  cbsd_list_uut = []
+  cbsd_list_th = []
+
+  for cbsds in sas_uut_fad_object.getCbsdRecords():
```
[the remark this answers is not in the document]
jni2000: Removed   creation of cbsd_list

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,641 @@
...
+    cbsd_list_uut.append(cbsds)
+
+  grant_objects_uut = getAllGrantInformationFromCbsdDataDump(cbsd_list_uut, True)
+
+  for fad in sas_th_fad_objects:
+    for cbsds in fad.getCbsdRecords():
```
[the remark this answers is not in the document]
jni2000: Removed   creation of cbsd_list

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,641 @@
...
+    grant_objects: a list of grants, each one being a namedtuple of type 
+                   CBSDGrantInformation, of all CBSDs from SAS UUT FAD and 
+                   SAS Test Harness FAD
+  """
+
+  grant_objects = []
+
+  # Loop over each CBSD grant
+  for cbsd_data_record in cbsd_data_records:
+    registration = cbsd_data_record.get('registrationRequest')
```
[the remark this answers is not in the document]
jni2000: Changed   to registration. Also updated in cbsd_x.json files in test_data directory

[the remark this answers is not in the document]
jni2000: Updated   to square brackets

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,676 @@
...
+  Creates a data structure to store post IAP allowed interference or 
+  aggregate interference to non-DPAs output in the format of 
+  {latitude : {longitude : [ output_first_5MHz_segment, output_first_5MHz_segment ]}}
+  """
+  def __init__(self):
+    self.manager = multiprocessing.Manager()
+    self.aggregate_interference_info = self.manager.dict()
+
+  def SetAggregateInterferenceInfo(self, latitude, longitude, interference):
+    if self.aggregate_interference_info.has_key(latitude) is False:
```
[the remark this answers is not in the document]
jni2000: Updated   as per suggestion

src/harness/reference_models/interference/interference.py
jni2000: Test   is done on this. It is working as expected.    Below approach mentioned in the python document is used to achieve   this :           "If standard (non-proxy) list or dict objects are contained in a   referent, modifications to those mutable values will not be propagated   through the manager because the proxy has no way of knowing when the values   contained within are modified. However, storing a value in a container proxy   (which triggers a __setitem__ on the proxy object) does propagate through the   manager and so to effectively modify such an item, one could re-assign the   modified value to the container proxy:          # create a list proxy and append a mutable object (a dictionary)     lproxy = manager.list()     lproxy.append({})     # now mutate the dictionary     d = lproxy[0]     d['a'] = 1     d['b'] = 2     # at this point, the changes to d are not yet synced, but by     # updating the dictionary, the proxy is notified of the change     lproxy[0] = d "          Even though we are using proxy container, sometime output is going hay-wire   because of multiple proocesses. Multiprocessing.Lock is used to synchronize   writing to shared dictionary across multiple processes

src/harness/reference_models/interference/interference.py
jni2000: Updated   as per suggestion

src/harness/reference_models/interference/interference.py
jni2000: Description   updated in comment serial number 18

src/harness/reference_models/interference/interference.py
jni2000: Returning   values in dBm. Converting to linear in the caller

jni2000:
Serge and Taran

We have addressed your latest comments and made all suggested fixes/changes. please take a look. Thanks.

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,669 @@
...
+    cbsd_grant: a namedtuple of type CbsdGrantInformation
+    constraint: protection constraint of type ProtectionConstraint
+    h_inc_ant: reference incumbent antenna height (in meters)
+    max_eirp: The maximum EIRP allocated to the grant during IAP procedure
+    region: Region type of the GWPZ or PPA area
+  Returns:
+    interference: interference contribution(dBm)
+  """
+
+  if (cbsd_grant.latitude == constraint.latitude and
```
sbdt-google: Info only (nothing to do): I will probably change the prop model so they handle this case, rather than letting this to the client code of the prop model. Will be more robust all over the test harness

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,669 @@
...
+  def __init__(self):
+    self.manager = multiprocessing.Manager()
+    self.aggregate_interference_info = self.manager.dict()
+    self.lock = multiprocessing.Lock()
+
+  def SetAggregateInterferenceInfo(self, latitude, longitude, interference):
+    self.lock.acquire()
+    try:
+      if latitude not in self.aggregate_interference_info:
+        self.aggregate_interference_info[latitude] = self.manager.dict()
```
sbdt-google: Looks better, should work although I have not tested

src/harness/reference_models/interference/interference.py
```diff
@@ -0,0 +1,669 @@
...
+  {latitude : {longitude : [ output_first_5MHz_segment, output_first_5MHz_segment ]}}
+  """
+  def __init__(self):
+    self.manager = multiprocessing.Manager()
+    self.aggregate_interference_info = self.manager.dict()
+    self.lock = multiprocessing.Lock()
+
+  def SetAggregateInterferenceInfo(self, latitude, longitude, interference):
+    self.lock.acquire()
+    try:
```
sbdt-google: usually one do:
```
with self.lock():
   ...
```
The lock acquire/release will be taken care of in a robust way

sbdt-google:
James & Sushant,
You are all set to submit this PR. Thanks for making the various improvements.

Please note however that as we plug the DPA stuff, we may rearrange the objects to minimize duplicated code, and make some basic refactoring. We will put you as reviewer of course.

jni2000:
Thanks, Surge. We will keep sync'ing up.

src/harness/reference_models/interference/interference.py
jni2000: Agreed. Will keep up with your enhancement. Thanks

src/harness/reference_models/interference/interference.py
jni2000: Thanks for the suggestion. Modified as suggested.

PR #500, opened 2018-03-30
Update SAS UUT Config to Seperate the URLs and version numbers

RedaHossino:
#seperate SAS 4 URLs:

- SAS-SAS (EC & RSA certif)
- CBSD-SAS (EC & RSA certif)
* harness test will choose using EC url or RSA url depending on the type of the cipher used  for the request)

#seperate SAS versions:
- SAS-SAS
- CBSD-SAS


src/harness/sas.cfg
```diff
@@ -1,4 +1,7 @@
 [SasConfig]
-BaseUrl: localhost:9000
-Version: v1.2
-
+AdminAndCbsdSasRsaBaseUrl: localhost:9000
```
kate-harrison: Can you split into `AdminBaseUrl` and `CbsdSasRsaBaseUrl`? It's confusing to have the two mixed together.

src/harness/sas.py
```diff
@@ -82,6 +93,18 @@ def DownloadFile(self, url, ssl_cert=None, ssl_key=None):
                           self._GetDefaultSasSSLCertPath(), ssl_key
                           if ssl_key else self._GetDefaultSasSSLKeyPath()))
 
+  def UpdateSasRequestUrl(self, cipher):
+    if 'ECDSA' in cipher:
+      sas_sas_active_base_url = self._sas_sas_ec_base_url
```
kate-harrison: Don't you need to use `self.sas_sas_active_base_url` (prepend with `self.`)?

src/harness/security_testcase.py
```diff
@@ -161,9 +161,9 @@ def doCbsdTestCipher(self, cipher, client_cert=None, client_key=None):
     """
     client_cert = client_cert or self._sas._GetDefaultCbsdSSLCertPath()
     client_key = client_key or self._sas._GetDefaultCbsdSSLKeyPath()
-
+    self._sas.UpdateCbsdRequestUrl(cipher)
     # Using pyOpenSSL low level API, does the SAS UUT server TLS session checks.
-    self.assertTlsHandshakeSucceed(self._sas_admin._base_url, [cipher],
+    self.assertTlsHandshakeSucceed(self._sas.cbsd_sas_active_base_url, [cipher],
```
kate-harrison: Nice bug fix!

src/harness/security_testcase.py
```diff
@@ -161,9 +161,9 @@ def doCbsdTestCipher(self, cipher, client_cert=None, client_key=None):
     """
     client_cert = client_cert or self._sas._GetDefaultCbsdSSLCertPath()
     client_key = client_key or self._sas._GetDefaultCbsdSSLKeyPath()
-
+    self._sas.UpdateCbsdRequestUrl(cipher)
```
kate-harrison: Do you need to call this (or something similar) again at the end of the function? Otherwise I think you may end up with one test case affecting another and the ordering of the tests becomes important.

src/harness/sas.cfg
RedaHossino: yes sure, I updated the code.

src/harness/sas.py
RedaHossino: done, thnx.

src/harness/security_testcase.py
RedaHossino: doesn't each test has its own instance of class SasImpl and then its own proper instance url variables?

kate-harrison: Oops, good point!

PR #501, opened 2018-03-30
MCP.1 functional portion for review

jni2000:
MCP.1 functional portion initial submit, please review

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_MCP_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_MCP_testcase.py
@@ -12,13 +12,22 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+#import ConfigParser
+from full_activity_dump_helper import getFullActivityDumpSasTestHarness, getFullActivityDumpSasUut
+from full_activity_dump_integration_test import FullActivityDumpIntegrationTest
+from functools import partial
 import json
 import os
+#import pre_iap
+#from reference_models.iap import iap
+#from reference_models.aggregate_interference import aggregate_interference
 import sas
 import sas_testcase
+import sas_objects
 from sas_test_harness import SasTestHarnessServer, generateCbsdRecords, \
     generatePpaRecords
-from util import winnforum_testcase, configurable_testcase, writeConfig, \
+import time    
+from util import winnforum_testcase, configurable_testcase, getCertificateFingerprint, writeConfig, \
   loadConfig, makePpaAndPalRecordsConsistent
 
 
@@ -292,10 +301,10 @@ class MultiConstraintProtectionTestcase(sas_testcase.SasTestCase):
         'conditionalRegistrationData': conditionals,
         'iterationData': [iteration0_config, iteration1_config],
         'sasTestHarnessConfigs': [sas_test_harness_0_config, sas_test_harness_1_config],
-        'domainProxyConfigs': [{'cert': os.path.join('certs', 'dp_1_client.cert'),
-                                'key': os.path.join('certs', 'dp_1_client.key')},
-                               {'cert': os.path.join('certs', 'dp_2_client.cert'),
-                                'key': os.path.join('certs', 'dp_2_client.key')}],
+        'domainProxyConfigs': [{'cert': os.path.join('certs', 'client.cert'),
+                                'key': os.path.join('certs', 'client.key')},
+                               {'cert': os.path.join('certs', 'client.cert'),
+                                'key': os.path.join('certs', 'client.key')}],
         'deltaIap': 2
     }
     writeConfig(filename, config)
@@ -306,3 +315,227 @@ class MultiConstraintProtectionTestcase(sas_testcase.SasTestCase):
        to 3700 MHz to protect configurable IAP-protected entities and DPAs
     """
     config = loadConfig(config_filename)
+
+    #config_parser = ConfigParser.RawConfigParser()
+    #config_parser.read(['sas_test_harness.cfg'])
+    #Headroom = config_parser.get('TestHarnessConfig', 'Headroom')
+    #ProtectionThreshold = config_parser.get('TestHarnessConfig', 'ProtectionThreshold')
+
+    cbsd_agg_list = []
+    fad_test_harnesses_objects = []
+    protected_entities = []
+    sas_test_harness_objects = []
+    domain_proxy_objects = []
+    for index in range(len(config['domainProxyConfigs'])):
+      domain_proxy_objects.append(sas_objects.DomainProxy(config['domainProxyConfigs'][index]['cert'], \
+                                                            config['domainProxyConfigs'][index]['key'], self))
+    # Step 1 : Load DPAs
+    self._sas_admin.TriggerLoadDpas()
+
+    # STEP 2 : ESC inf rms SAS about inactive DPA
+    self._sas_admin.TriggerBulkDpaActivation({'activate': False})
+
+    # Step 3 : creates multiple SAS TH, and Load predefined FAD,CBSD
+    for index, test_harness in enumerate(config['sasTestHarnessConfigs']):
+      # Initialize SAS Test Harness Server instance to dump FAD records
+      sas_test_harness_objects.append(SasTestHarnessServer(config['sasTestHarnessConfigs'][index]['sasTestHarnessName'],
+                                     config['sasTestHarnessConfigs'][index]['hostName'],
+                                     config['sasTestHarnessConfigs'][index]['port'],
+                                     config['sasTestHarnessConfigs'][index]['serverCert'],
+                                     config['sasTestHarnessConfigs'][index]['serverKey'],
+                                     config['sasTestHarnessConfigs'][index]['caCert']))
+
+    for index, test_harness in enumerate(sas_test_harness_objects):
+        # Start the server
+        test_harness.start()
+
+        # informing SAS UUT about SAS Test Harnesses
+        certificate_hash = getCertificateFingerprint(config['sasTestHarnessConfigs'][index]['serverCert'])
+        self._sas_admin.InjectPeerSas({'certificateHash': certificate_hash,
+                                       'url': test_harness.getBaseUrl()})
+
+    # Step 4,5 : Inject IAP protected entities into UUT
+    for iteration_content in config['iterationData']:
+      records = iteration_content['protectedEntities']
+      if records.has_key('fssRecords'):
+        for fss_record in records['fssRecords']:
+          self._sas_admin.InjectFss({'record': fss_record})
+      if records.has_key('gwpzRecords'):
+        for gwpz_record in records['gwpzRecords']:
+          self._sas_admin.InjectWisp({'record': gwpz_record})
+      if records.has_key('escRecords'):
+        for esc_record in records['escRecords']:
+          self._sas_admin.InjectEscSensorDataRecord({'record': esc_record})
+      if records.has_key('ppaRecords'):
+        for ppa_record in records['ppaRecords']:
+          self._sas_admin.InjectZoneData({'record': ppa_record})
+
+      # Step 6,7 : Creating FAD Object and Pull FAD records from SAS UUT
+      if len(iteration_content['sasTestHarnessData']) > 0:
+        fad_uut_object = getFullActivityDumpSasUut(self._sas,
+                                                   self._sas_admin,
+                                                   config['domainProxyConfigs'][0]['cert'],
+                                                   config['domainProxyConfigs'][0]['key'])
+
+        # Pull FAD from SAS Test Harnesses
+        for index, test_harness in enumerate(sas_test_harness_objects):
+          fad_test_harnesses_objects.append(getFullActivityDumpSasTestHarness(
+                                  test_harness.getSasTestHarnessInterface(),
+                                  config['sasTestHarnessConfigs'][index]['serverCert'],
+                                  config['sasTestHarnessConfigs'][index]['serverKey']
+                              )
+          )
+
+      # Step 8 : Trigger CPAS and wait until completion
+      self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
+
+      # Step 9 : Call IAP reference model, calculated at step 12
+      # Collect all the protected Entites
+      protected_entities.append(iteration_content['protectedEntities'])
+
+      # Call Pre IAP model
+      # TODO: To invoke pre-iap model
+      #pre_iap_headRooms = { Headroom }
+      #pre_iap.pre_iap(protected_entities, fad_uut_object, fad_test_harnesses_objects, pre_iap_headRooms)
+
+      entities = []
+      for data in iteration_content['protectedEntities'].values():
+        entities.append(data[0])
+      # Call IAP reference model
+      # TODO: To invoke IAP reference model
+      #if records.has_key('ppaRecords'):
+      #  pal_records = iteration_content['protectedEntities']['palRecords']
+      #  iap_refmodel_margin_list = iap.performIap(entities, fad_uut_object, fad_test_harnesses_objects, pal_records)
+      #else:
+      #  pal_records = ""
+      #  iap_refmodel_margin_list = iap.performIap(entities, fad_uut_object, fad_test_harnesses_objects, pal_records)
+
+      # Step 10 : DP Test Harness Register N(2,k)CBSDs with SAS UUT
+      # Use DP objects to get CBSD registered
+      for domain_proxy_object in domain_proxy_objects:
+        for index, cbsdRequestsWithDomainProxys in enumerate(iteration_content['cbsdRequestsWithDomainProxies']):
+          registration_requests = cbsdRequestsWithDomainProxys['registrationRequests']
+          grant_requests = cbsdRequestsWithDomainProxys['grantRequests']
+
+          # domain_proxy_object.initialize triggers the registration and grant for the records passes as parameters
+          domain_proxy_object.initialize(registration_requests, grant_requests)
+
+          # Step 11 : Send heartbeat request
+          domain_proxy_object.heartbeatRequestForAllGrants()
+
+          # Get the list of CBSDs
+          cbsd_agg_list.append(domain_proxy_object.getCbsdsWithAtLeastOneAuthorizedGrant())
+
+      # Register individual Cbsds
+      for cbsd_record in iteration_content['cbsdRecords']:
+        cbsd_ids, grant_ids = self.assertRegisteredAndGranted([cbsd_record['registrationRequest']],
+                                                              [cbsd_record['grantRequest']])
+        cbsd_agg_list.append(cbsd_ids)
+
+      # Step 12 : Invoke Aggregate Interference Model
+      # TODO: To invoke Aggregate Interference Model
+      #for protected_entity in iteration_content['protectedEntities'].keys():
+        #agg_interference_list = self.callAggregateInterferenceModel(cbsd_agg_list, iteration_content['protectedEntities'])
+
+        # check MCP.1
+        # TODO: iap_refmodel_margin_list input required
+        #self.checkMcpIap(iteration_content['protectedEntities'], agg_interference_list, iap_refmodel_margin_list, config['deltaIap'])
+
+      # Step 13 : Configure SAS Harness with FAD,trigger FAD generation
+      for index, test_harness in enumerate(sas_test_harness_objects):
+        sas_test_harness_dump_records = [iteration_content['sasTestHarnessData'][index]['cbsdRecords']]
+        test_harness.writeFadRecords(sas_test_harness_dump_records)
+
+      # Step 14,15 : Trigger Full Activty Dump and Pull FAD records from SAS UUT
+      fad_uut_object = getFullActivityDumpSasUut(self._sas,
+                                                 self._sas_admin,
+                                                 config['domainProxyConfigs'][0]['cert'],
+                                                 config['domainProxyConfigs'][0]['key'])
+
+      # Step 16 : Trigger CPAS and wait for its completion
+      self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
+
+      # Step 17 : Call IAP reference model
+      # TODO: To invoke DPA Move Reference model
+
+      # Call Pre IAP model
+      # TODO: To invoke Pre-iap reference model
+      #pre_iap.pre_iap(protected_entities, fad_uut_object, fad_test_harnesses_objects, pre_iap_headRooms)
+
+      # Call IAP reference model
+      # TODO: To invoke IAP reference model
+      # iap_refmodel_margin_list = iap.performIap(entities, fad_uut_object, fad_test_harnesses_objects, pal_records)
+
+      # Step 18,19,20 and 21 :
+      # Send heartbeat request for the grants, relinquish the grant, grant request and heartbeat for new grant
+      for domain_proxy_object in domain_proxy_objects:
+        domain_proxy_object.performHeartbeatAndUpdateGrants()
+
+        # Step 22: Calculating the Aggregate interface and IAP reference model invoke.
+        cbsd_agg_list.append(domain_proxy_object.getCbsdsWithAtLeastOneAuthorizedGrant())
+
+      # Invoke Aggregate Interference Model
+      # TODO: To invoke Aggregate Interference Model
+      #for protected_entity in iteration_content['protectedEntities'].keys():
+      #  agg_interference_list = self.callAggregateInterferenceModel(cbsd_agg_list, iteration_content['protectedEntities'])
+
+        # Check MCP.1
+        #self.checkMcpIap(iteration_content['protectedEntities'], agg_interference_list, iap_refmodel_margin_list, config['deltaIap'])
+
+      # Step 23: ESC Test harness enables DPA activations
+      # deactivated the dpaDeactivationList DPA IDs
+      for dpa in iteration_content['dpaActivationList']:
+        self._sas_admin.TriggerDpaDeactivation(dpa)
+
+      # activated the dpaActivationList DPA IDs
+      # step 24: wait for 240 sec if DPA is activated in step 23 else 15 sec
+      for dpa in iteration_content['dpaActivationList']:
+        self._sas_admin.TriggerDpaActivation(dpa)
+        time.sleep(240)
+      else:
+        time.sleep(15)
+
+      # Step 25,26,27 and 28 :
+      # Send heartbeat request for the grants, relinquish the grant, grant request and heartbeat for new grant
+      for domain_proxy_object in domain_proxy_objects:
+        domain_proxy_object.performHeartbeatAndUpdateGrants()
+
+        # Step 29
+        # TODO: Aggregate interference for DPA
+
+        # Step 30:Invoke Aggregate Interference Model
+        cbsd_agg_list.append(domain_proxy_object.getCbsdsWithAtLeastOneAuthorizedGrant())
+
+      # Invoke Aggregate Interference Model
+      # TODO: To invoke Aggregate Interference Model
+      #for protected_entity in iteration_content['protectedEntities'].keys():
+      #  agg_interference_list = self.callAggregateInterferenceModel(cbsd_agg_list, iteration_content['protectedEntities'])
+
+        # check MCP.1
+      #  self.checkMcpIap(iteration_content['protectedEntities'], agg_interference_list, iap_refmodel_margin_list, config['deltaIap'])
+
+      # check MCP.1 DPA
+      # TODO: need to invoke the DPA checkperformHeartbeatAndUpdateGrants
+
+      # checkMcpIap(pe,agg_interference_list,agg_interference_iap_list,pre_iap_headroom,left_over_alloc)
+
+    for test_harness in sas_test_harness_objects:
+      test_harness.shutdown()
+      del test_harness
+
+  def checkMcpIap(self, protected_entity, agg_interference_list, iap_refmodel_margin_list, delta_iap):
+    """Calculating IAP for each FSS and ESC sensor protection point,
+    and for at least 95% of the protection points of each PPA and GWPZ.
+    """
+    pass
+
+  def callAggregateInterferenceModel(self, cbsd_list, protected_entity):
+    """Args:
+         cbsd_list: List of cbsds with Authorized grants
+         protected_entity: A dictionary with multiple key-value pairs where
+         keys are protection entities
+
+       Returns:
+         Aggregate Interference list
+    """
+    pass


src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      domain_proxy_objects.append(sas_objects.DomainProxy(config['domainProxyConfigs'][index]['cert'], \
+                                                            config['domainProxyConfigs'][index]['key'], self))
+    # Step 1 : Load DPAs
+    self._sas_admin.TriggerLoadDpas()
+
+    # STEP 2 : ESC inf rms SAS about inactive DPA
+    self._sas_admin.TriggerBulkDpaActivation({'activate': False})
+
+    # Step 3 : creates multiple SAS TH, and Load predefined FAD,CBSD
+    for index, test_harness in enumerate(config['sasTestHarnessConfigs']):
```
kate-harrison: Can't you just use `test_harness` (which should be `test_harness_config`) directly instead of `config['sasTestHarnessConfigs'][index]`?

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      # Initialize SAS Test Harness Server instance to dump FAD records
+      sas_test_harness_objects.append(SasTestHarnessServer(config['sasTestHarnessConfigs'][index]['sasTestHarnessName'],
+                                     config['sasTestHarnessConfigs'][index]['hostName'],
+                                     config['sasTestHarnessConfigs'][index]['port'],
+                                     config['sasTestHarnessConfigs'][index]['serverCert'],
+                                     config['sasTestHarnessConfigs'][index]['serverKey'],
+                                     config['sasTestHarnessConfigs'][index]['caCert']))
+
+    for index, test_harness in enumerate(sas_test_harness_objects):
```
kate-harrison: Why not do this inside the previous loop?

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
        to 3700 MHz to protect configurable IAP-protected entities and DPAs
     """
     config = loadConfig(config_filename)
+
```
kate-harrison: In other places, I think you assumed that the implementation of MCP was actually in a separate function -- which I think was a good idea. This implementation doesn't seem to match with that, though.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+        # Start the server
+        test_harness.start()
+
+        # informing SAS UUT about SAS Test Harnesses
+        certificate_hash = getCertificateFingerprint(config['sasTestHarnessConfigs'][index]['serverCert'])
+        self._sas_admin.InjectPeerSas({'certificateHash': certificate_hash,
+                                       'url': test_harness.getBaseUrl()})
+
+    # Step 4,5 : Inject IAP protected entities into UUT
+    for iteration_content in config['iterationData']:
```
kate-harrison: Consider making a "do one iteration" function to keep things clean. (Not totally sure if this is the best way to go, but it might be.)

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      records = iteration_content['protectedEntities']
```
kate-harrison: Use a less generic name please.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+          self._sas_admin.InjectWisp({'record': gwpz_record})
+      if records.has_key('escRecords'):
+        for esc_record in records['escRecords']:
+          self._sas_admin.InjectEscSensorDataRecord({'record': esc_record})
+      if records.has_key('ppaRecords'):
+        for ppa_record in records['ppaRecords']:
+          self._sas_admin.InjectZoneData({'record': ppa_record})
+
+      # Step 6,7 : Creating FAD Object and Pull FAD records from SAS UUT
+      if len(iteration_content['sasTestHarnessData']) > 0:
```
kate-harrison: It's cleaner to directly reference the number of SAS THs.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+        fad_uut_object = getFullActivityDumpSasUut(self._sas,
+                                                   self._sas_admin,
+                                                   config['domainProxyConfigs'][0]['cert'],
```
kate-harrison: No, not this config...

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+                                                   config['domainProxyConfigs'][0]['key'])
+
+        # Pull FAD from SAS Test Harnesses
+        for index, test_harness in enumerate(sas_test_harness_objects):
+          fad_test_harnesses_objects.append(getFullActivityDumpSasTestHarness(
+                                  test_harness.getSasTestHarnessInterface(),
+                                  config['sasTestHarnessConfigs'][index]['serverCert'],
```
kate-harrison: This technically works, but it's confusing. Please add a note about why it's OK to do this.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+                                  config['sasTestHarnessConfigs'][index]['serverKey']
+                              )
+          )
+
+      # Step 8 : Trigger CPAS and wait until completion
+      self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
+
+      # Step 9 : Call IAP reference model, calculated at step 12
+      # Collect all the protected Entites
+      protected_entities.append(iteration_content['protectedEntities'])
```
kate-harrison: Is this really the format you want? I don't think it's the format used in e.g. the pre IAP ref model.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+
+      # Call Pre IAP model
+      # TODO: To invoke pre-iap model
+      #pre_iap_headRooms = { Headroom }
+      #pre_iap.pre_iap(protected_entities, fad_uut_object, fad_test_harnesses_objects, pre_iap_headRooms)
+
+      entities = []
+      for data in iteration_content['protectedEntities'].values():
+        entities.append(data[0])
```
kate-harrison: Why only the first?

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+
+    for index, test_harness in enumerate(sas_test_harness_objects):
+        # Start the server
+        test_harness.start()
+
+        # informing SAS UUT about SAS Test Harnesses
+        certificate_hash = getCertificateFingerprint(config['sasTestHarnessConfigs'][index]['serverCert'])
+        self._sas_admin.InjectPeerSas({'certificateHash': certificate_hash,
+                                       'url': test_harness.getBaseUrl()})
+
```
kate-harrison: Where are you loading the content into the SAS THs?

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      # Step 23: ESC Test harness enables DPA activations
+      # deactivated the dpaDeactivationList DPA IDs
+      for dpa in iteration_content['dpaActivationList']:
+        self._sas_admin.TriggerDpaDeactivation(dpa)
+
+      # activated the dpaActivationList DPA IDs
+      # step 24: wait for 240 sec if DPA is activated in step 23 else 15 sec
+      for dpa in iteration_content['dpaActivationList']:
+        self._sas_admin.TriggerDpaActivation(dpa)
+        time.sleep(240)
```
kate-harrison: You sleep once after all of them, not once each.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +315,227 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      else:
```
kate-harrison: Not valid.

jni2000:
Thanks, Kate, for your encouragement to my team, really inspired us a lot. We will address your comments and get back to you as soon as possible.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -292,10 +301,10 @@ def generate_MCP_1_default_config(self, filename):
         'conditionalRegistrationData': conditionals,
         'iterationData': [iteration0_config, iteration1_config],
         'sasTestHarnessConfigs': [sas_test_harness_0_config, sas_test_harness_1_config],
-        'domainProxyConfigs': [{'cert': os.path.join('certs', 'dp_1_client.cert'),
-                                'key': os.path.join('certs', 'dp_1_client.key')},
-                               {'cert': os.path.join('certs', 'dp_2_client.cert'),
-                                'key': os.path.join('certs', 'dp_2_client.key')}],
+        'domainProxyConfigs': [{'cert': os.path.join('certs', 'client.cert'),
```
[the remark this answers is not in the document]
janetliuboston: While testing we realized that there are no certificates named 'dp_client' and 'dp_key', so updated with 'domain_proxy.cert' and 'domain_proxy.key' .

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Agreed, updated as suggested.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Agreed, updated the code with single loop.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Yes, complete MCP steps will be in one function. Updated the code as mentioned.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: We considered this approach, but now MCP steps are put into one function which has similar functionality as "do one iteration" except the initial portions. While adding the IAP and pre-IAP functionalities we will consider "one iteration" function if needed.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Agreed. Modified the name "records" to "protected_entity_records".

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: We do not have any variable for indicating "number of SAS THs" directly, so we used this way.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Updated the cert and key for UUT FAD.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Updated to add a comment.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: This portion of code is used for pre-IAP functionality. Removed now. We will add new code once pre-IAP model is integrated.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: FAD records are added in step 13 by using the "test_harness.writeFadRecord" method.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Updated. Sleep is invoked after the for loop with condition.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
janetliuboston: Updated as per the step 24 in MCP.1 test case.

janetliuboston:
Thank you, Kate. This is ready to be reviewed again.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +311,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+
+          # Step 11 : Send heartbeat request
+          domain_proxy_object.heartbeatRequestForAllGrants()
+
+          # Get the list of CBSDs
+          cbsd_agg_list.append(domain_proxy_object.getCbsdsWithAtLeastOneAuthorizedGrant())
+
+      # Register individual Cbsds
+      for cbsd_record in iteration_content['cbsdRecords']:
+        cbsd_ids, grant_ids = self.assertRegisteredAndGranted([cbsd_record['registrationRequest']],
```
TaranFurnell: assertRegisteredAndGranted supports conditional registration, please add support in this testcase.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +311,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+                                                              [cbsd_record['grantRequest']])
+        cbsd_agg_list.append(cbsd_ids)
```
TaranFurnell: getCbsdsWithAtLeastOneAuthorizedGrant returns Cbsd objects, assertRegisteredAndGranted returns CBSD ids (I think this is a string). So currently 2 different data types are in this list, the CBSD id one probably doesn't have sufficient information for this test.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +311,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+
+      # Step 12 : Invoke Aggregate Interference Model
+      # TODO: To invoke Aggregate Interference Model
+
+      # Check MCP.1
+      # TODO: To invoke checkMcpIap method
+
+      # Step 13 : Configure SAS Harness with FAD,trigger FAD generation
+      for index, test_harness in enumerate(sas_test_harness_objects):
+        sas_test_harness_dump_records = [iteration_content['sasTestHarnessData'][index]['cbsdRecords']]
```
TaranFurnell: Use zip, there's no need to get the index.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +311,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
        to 3700 MHz to protect configurable IAP-protected entities and DPAs
     """
     config = loadConfig(config_filename)
+    test_type = "MCP"
```
TaranFurnell: Nit: The convention in this repo is to use single quotes.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +311,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+        # Invoke Aggregate Interference Model
+        # TODO: To invoke Aggregate Interference Model
+
+        # Check MCP.1
+        # TODO: To invoke checkMcpIap method
+
+        # check MCP.1 DPA
+        # TODO: need to invoke the DPA checkperformHeartbeatAndUpdateGrants
+
+      for test_harness in sas_test_harness_objects:
```
TaranFurnell: I think this is at the wrong level (big file). Currently it is run every iteration.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +311,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+    Return: None
+    """
+    cbsd_agg_list = []
+    fad_test_harnesses_objects = []
+    sas_test_harness_objects = []
+    domain_proxy_objects = []
+    sas_uut_cert = os.path.join('certs/sas.cert')
+    sas_uut_key = os.path.join('certs/sas.key')
+
+    for index in range(len(config['domainProxyConfigs'])):
```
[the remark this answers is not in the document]
jni2000: As   suggested, iterating directly through elements of list

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +311,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      if 'ppaRecords' in protected_entity_records:
+        for ppa_record in protected_entity_records['ppaRecords']:
+          self._sas_admin.InjectZoneData({'record': ppa_record})
+
+      # Step 6,7 : Creating FAD Object and Pull FAD records from SAS UUT
+      if len(iteration_content['sasTestHarnessData']) > 0:
+        fad_uut_object = getFullActivityDumpSasUut(self._sas,
+                                                   self._sas_admin,
+                                                   sas_uut_cert,
+                                                   sas_uut_key)
```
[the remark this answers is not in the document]
jni2000: Removed   the certs, keys while calling getFullActivityDumpSasUut

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
jni2000: Conditionals   are added through admin interface at once in the test case. This was   implemented based on the previous review comments.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
jni2000: As   suggested, modified to use "zip"

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      # TODO: To invoke IAP reference model
+
+      # Step 10 : DP Test Harness Register N(2,k)CBSDs with SAS UUT
+      # Use DP objects to get CBSD registered
+      for domain_proxy_object in domain_proxy_objects:
+        for cbsdRequestsWithDomainProxy in iteration_content['cbsdRequestsWithDomainProxies']:
+          registration_requests = cbsdRequestsWithDomainProxy['registrationRequests']
+          grant_requests = cbsdRequestsWithDomainProxy['grantRequests']
+
+          # domain_proxy_object.initialize triggers the registration and grant for the records passes as parameters
```
TaranFurnell: Comment is out of date.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+       config: Testcase configuration
+       test_type: A string which indicates the type of testcase to be invoked("MCP"/"XPR")
+    Return: None
+    """
+    cbsd_agg_list = []
+    fad_test_harnesses_objects = []
+    sas_test_harness_objects = []
+    domain_proxy_objects = []
+    sas_uut_cert = os.path.join('certs/sas.cert')
+    sas_uut_key = os.path.join('certs/sas.key')
```
TaranFurnell: cert and key are unused.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,214 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+        # Step 23: ESC Test harness enables DPA activations
+        # deactivated the dpaDeactivationList DPA IDs
+        for dpa in iteration_content['dpaActivationList']:
+          self._sas_admin.TriggerDpaDeactivation(dpa)
+
+        # activated the dpaActivationList DPA IDs
+        # step 24: wait for 240 sec if DPA is activated in step 23 else 15 sec
+        for dpa in iteration_content['dpaActivationList']:
+          self._sas_admin.TriggerDpaActivation(dpa)
+        if len(iteration_content['dpaActivationList']):
```
TaranFurnell: why check len? If it's an empty list the result should be the same.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,211 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      if 'escRecords' in protected_entity_records:
+        for esc_record in protected_entity_records['escRecords']:
+          self._sas_admin.InjectEscSensorDataRecord({'record': esc_record})
+
+      if 'ppaRecords' in protected_entity_records:
+        for ppa_record in protected_entity_records['ppaRecords']:
+          self._sas_admin.InjectZoneData({'record': ppa_record})
+
+      # Step 6,7 : Creating FAD Object and Pull FAD records from SAS UUT
+      if len(iteration_content['sasTestHarnessData']) > 0:
```
TaranFurnell: ```if iteration_content['sasTestHarnessData']:```?

Empty lists are falsy.

jni2000: Updated   as per the comment

jni2000:
Taran and Kate, we have responded your latest review comments, please take a look. Thanks.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,239 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+
+      # informing SAS UUT about SAS Test Harnesses
+      certificate_hash = getCertificateFingerprint(test_harness['serverCert'])
+      self._sas_admin.InjectPeerSas({'certificateHash': certificate_hash,
+                                     'url': sas_test_harness_object.getBaseUrl()})
+      sas_test_harness_objects.append(sas_test_harness_object)
+   
+    # Step 4,5 : Inject IAP protected entities into UUT
+    for iteration_content in config['iterationData']:
+      cbsd_agg_list = []
```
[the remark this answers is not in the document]
janetliuboston: Yes, we have to collect the list of CBSD records/objects which are added individually and via domain proxies. The code has been updated with 'cbsds_with_domain_proxies', a list of CBSDs added via domain proxies
'cbsd_agg_list' - a list of CBSDs added individually. 
We have to consolidate both the list and provide as input to aggregate interference model, currently we are woking on this changes. It will be reflected as part of reference models APIs integration.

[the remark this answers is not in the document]
TaranFurnell: Consolidating individual CBSDs and Domain Proxy CBSDs could be a pain, yes. I'm more concerned that handling individual CBSDs will effectively require re-implementing the same thing a Domain Proxy does in the test. I'd like to propose a solution to this.

For each individual CBSD create a Domain Proxy with only that single CBSD and the cert/key as given in the CBSDs config. Add each of these to the list of Domain Proxies.

This way:
1) There is no need to re-implement the register/request grant/heartbeat behavior for individual CBSDs.
2) Support for individual CBSDs will exist implicitly in the test without any need for special consideration. This should simplify the test code a bit.

Does this make sense? and have I missed anything? As far as we can tell it should work as Domain Proxy requests look the same as a CBSD request (Except the cert/keys are different and Domain Proxies can make requests for more than one CBSD)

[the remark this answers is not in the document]
janetliuboston: Yes, got the point. For each individual CBSD, we can create a domain proxy object which in turn will take care of all the CBSD related operations.
Earlier we had a discussion with Kate on the individual CBSDs presence in MCP.1 test case. Please see below excerpt comments from PR405.
 
kate-harrison on Mar 13 
Contributor
I just realized that there's another piece missing: we need the capability for individual CBSDs (which will use their own cert/key pair) to register, get a grant, etc. So in the same way we support multiple DPs, we need to support multiple individual CBSD connections.

Sorry for catching this only at the last minute -- I literally remembered only as I was writing a reply to an unrelated email. 

jni2000 on Mar 13 • edited  
Contributor
AS per the current test spec, MCP.1 test case uses only DP test harness for CBSD’s registration and grants, so we would like to know exactly in which test step we will be using the individual CBSDs ? is there any updates on the MCP.1 test case document? Please let us know your thoughts on the comment.

kate-harrison on Mar 14 
Contributor
I believe that the intention is to use individual CBSDs in every step where "DP test harness" is mentioned: so ITS can use N DPs and M individual CBSDs in each of these steps. The test spec is unfortunately not very clear on this point; you are welcome to double-check with Masoud.

jni2000:
Taran

Please take a look of our response to your latest review comment. Thanks.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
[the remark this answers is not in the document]
TaranFurnell: Yes, individual CBSDs are needed (to clarify in this response "individual CBSD" means a CBSD that is not in a Domain Proxy). My point is individual CBSDs behave the same way as a Domain Proxy with a single CBSD. What I am proposing is the config has individual CBSDs as is, but the test takes each individual CBSD and make a new "domain proxy" (configured with a CBSD cert/key) to handle the behavior and stick it with the rest of the domain proxies. This will simplify the handling logic.

Example (takes the place of lines 459-462)
```python
for cbsd_record in iteration_content['cbsdRecords']:
  proxy = test_harness_objects.DomainProxy(cbsd_record['clientCert'], cbsd_record['clientKey'], self)
  proxy.registerCbsdsAndRequestGrants([cbsd_record['registrationRequest']], [cbsd_record['grantRequest']])
  domain_proxy_objects.append(proxy)
```

Note: Old version of Domain Proxy initialization used for consistency with the current code, will need to be updated.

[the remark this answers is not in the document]
jni2000: Updated the as suggested in the comment, created new DP object for each CBSD.

jni2000:
Taran

We have addressed your latest review comment to our earlier responses and we agreed and made the changes as you suggested. Please continue reviewing. Thanks.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,241 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+
+  def executeMCPOnlyTestSteps(self, iteration_content, domain_proxy_objects, cbsd_as_dp_list):
+    """Executes the steps from Step23 to Step30 for MCP testcase
+
+    Args:
+       iteration_content: A dictionary with multiple key-value pairs that contain iteration data
+       domain_proxy_objects: Domain proxy objects
+       cbsd_as_dp_list: list of Domain Proxy objects, which act like individual CBSD
+    """
+    cbsds_with_domain_proxies = []
```
TaranFurnell: There is another variable called ```cbsds_with_domain_proxies``` that exists higher up. It looks like this is the same kind of data, just after DPA activation. Please change to something like ```cbsds_with_domain_proxies_after_dpa_activation``` so it's clear this is not the same variable.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,241 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+      domain_proxy_object.heartbeatForAllActiveGrants()
+
+      # Get the list of CBSDs
+      cbsds_with_domain_proxies.append(domain_proxy_object.getCbsdsWithAtLeastOneAuthorizedGrant())
+
+    # Register individual Cbsds(Domain Proxy with single CBSD)
+    for cbsd_record in iteration_content['cbsdRecords']:
+      proxy = test_harness_objects.DomainProxy(cbsd_record['clientCert'], cbsd_record['clientKey'], self)
+      proxy.registerCbsdsAndRequestGrants([cbsd_record['registrationRequest']], [cbsd_record['grantRequest']])
+      cbsd_as_dp_list.append(proxy)
```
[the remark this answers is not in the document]
jni2000: Yes, Updated the as suggested in the comments, removed the variable which holds individual CBSDs

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
jni2000: Yes, variable is used for same kind of data, after above changes this variable is no longer required, modified as per the latest changes

jni2000:
Taran

We have addressed your latest two review comments. Please take a look. If no more concerns, can we please move forward to approve ? Thanks

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,238 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+    """Executes the steps from Step1 to Step22 for MCP and XPR testcases
+
+    Args:
+       test_type: A string which indicates the testcase type
+       iteration_content: A dictionary with multiple key-value pairs that contain iteration data
+       sas_test_harness_objects: SAS Test Harnesses objects
+       domain_proxy_objects: Domain proxy objects
+    """
+    fad_test_harnesses_objects = []
+    cbsds_with_domain_proxies = []
```
TaranFurnell: Comment what these variables are used for and/or move them down to the last place they could reasonably be created.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -306,3 +310,238 @@ def test_WINNF_FT_S_MCP_1(self, config_filename):
...
+
+    # Step 9 : Invoke IAP reference model
+    # Call Pre IAP model
+    # TODO: To invoke pre-iap model
+
+    # Call IAP reference model
+    # TODO: To invoke IAP reference model
+
+    # Step 10 : DP Test Harness Register N(2,k)CBSDs with SAS UUT
+    # Use DP objects to get CBSD registered
```
TaranFurnell: nit: ```Use DP objects to register CBSDs``` perhaps?

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -12,13 +12,17 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
+from full_activity_dump_helper import getFullActivityDumpSasTestHarness, getFullActivityDumpSasUut
+from functools import partial
```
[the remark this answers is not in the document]
jni2000: we will keep this for now and remove if indeed not required for all the derived test cases.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
jni2000: changed as suggested.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
jni2000: done as suggested.

PR #503, opened 2018-03-30
Fix small set of issues on previous #476

sbdt-google:
- zones union of elements instead of MultiPolygon
 which keeps only first polygon of inner MultiPolygons
 - update ppa_test with the latest geo.utils
 - fix issue introduced last minute in time_dpa
 - massive speed up of the distribution when urban areas
 - small improvements to the script

### DIFF
--- a/src/harness/reference_models/examples/benchmark/time_dpa.py
+++ b/src/harness/reference_models/examples/benchmark/time_dpa.py
@@ -44,7 +44,7 @@ from reference_models.examples import entities
 dpa_name = 'east_dpa_7'
 
 # - Number of sites to distribute within a given range of the DPA
-num_sites = 45000
+num_sites = 100
 max_dist_cat_a = 160
 max_dist_cat_b = 320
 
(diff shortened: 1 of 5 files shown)


src/harness/reference_models/examples/benchmark/time_dpa.py
```diff
@@ -114,23 +114,25 @@ def PrepareSimulation():
   zone_catb = dpa_zone.buffer(extend_catb_deg).intersection(us_border)
 
   if urban_areas is not None:
+    # simplify the huge urban_areas for quicker inclusion tests
     urban_areas = urban_areas.intersection(zone_catb)
 
   # - Distribute the CBSDs
+  print ' - Cat A indoor'
   cbsds_cat_a_indoor = entities.GenerateCbsdsInPolygon(
       num_sites * ratio_cat_a_indoor,
```
sushantparkhi: Maybe use of 2 Space Indentation instead of 4 here and below

sushantparkhi:
Looks good to me!

src/harness/reference_models/examples/benchmark/time_dpa.py
sbdt-google: Indentation in continuing statement (after "(") is auto done to 4 char, independently of the 2 char of global indentation. I think that is the common way to use, and as supported by most editors python modes

PR #506, opened 2018-04-04
QPR.2 and QPR.6 tests

harinineelu:


jni2000:
We have reviewed this PR.  Since we have learnt that a lot of requirements are changed, and the TS 0061 needs to change. So we will hold on the review till we have answers from the WinnF.

harinineelu:
Thanks Janet, Merged #506

PR #507, opened 2018-04-05
DPA reference model: upgrade for IPR and MCP tests

sbdt-google:
This PR update NIST move list ref model:
 - new module `dpa_mgr` for handling the public interface for move list, aggregated interference calculation and interference check of ref model versus SAS UUT
 - cleaning of the move_list.py module in relation to these changes. The legacy findMoveList routine is still kept to not break any client code using it.
 - integration of the reference_models/common/ package with:
   * common data structures used by DPA but also interference/IAP modules. A future PR will finalize the convergence to these data structures
   * a pool module for managing all multiprocessing in one place
   * a cache module, used for DPA interference check robustness, and which will evolve later for multiprocess cache sharing as well

Proper unit testing will be added later on another PR. The NIST example has been ported to the new franmework (see dpa_mgr_example). As well as the more involved benchmark in examples/benchmark/time_dpa.py.
Tests have been done using these 2 tools that it gives same results as before.


sbdt-google:
Hi Janet,

I made a minor update to support the variable catB distance that will be officially included soon, as well as the azimuth range support.
You can have a quick look before I submit.
Thanks,
Serge

sbdt-google:
Thanks!

PR #508, opened 2018-04-05
Miscellaneous test harness fixes

TaranFurnell:
- Fixed FAD record file size
- Added logging of http server failure (In the very unlikely event that a test harness HTTP server dies it would currently die silently, this makes sure it is logged and attempts to restart)
- Moved generation time fix into fad generation (The earlier solution was a quick fix, this bakes it into the files, also done there to ensure size is accurate)
- Modified test harness interface to access the harness directly (Prevents tests from needing to know the matching client certificate, FAD helper modification will follow to actually make use of this later, for now you can use the getFullActivityDumpSasTestHarness helper function with the cert and key as None) Edit: FAD helper modification now included.

Edit: Additional fixes
- Fixed FAD record checksum
- Switched test harness to use SimpleHTTPRequestHandler
- Updated getFullActivityDumpSasTestHarness to not have certificate/key as arguments.


src/harness/sas_test_harness.py
```diff
@@ -133,54 +133,70 @@ def getSasTestHarnessVersion(self):
...
     return full_activity_dump
 
-  def __createFadRecord(self, encoded_url, record):
+  def __createFadRecord(self, encoded_url, record, file_name):
     """ Creates the files object that includes url,checksum,size,version and recordtype of record
         required to be packed in FAD.
     """
     fad_record = {}
     fad_record['url'] = encoded_url
     fad_record['checksum'] = hashlib.sha1(json.dumps(record)).hexdigest()
```
jcereced: do we have the same issue with the checksum (need to calculate the checksum for the entire file, not just the record)?

jcereced:
Just run the latest changes and this is what I see:




{"files": [{"url": "https://10.250.12.169:9001/v1.2/cbsd/activity_dump_file_cbsd0.json", "checksum": "e8d733ad19eebd574f60346d74781e74f19e2b1e", "version": "v1.2", "recordType": "cbsd", "size": 3404}, {"url": "https://10.250.12.169:9001/v1.2/zone/activity_dump_file_zone1.json", "checksum": "cc1c6bdcfc7dd78409818684bdff170569e70612", "version": "v1.2", "recordType": "zone", "size": 2382}, {"url": "https://10.250.12.169:9001/v1.2/esc_sensor/activity_dump_file_esc_sensor2.json", "checksum": "13ad241ba7686f31116a96a68ab2805fe2b775be", "version": "v1.2", "recordType": "esc_sensor", "size": 43533}], "generationDateTime": "2018-04-05T19:21:24Z", "description": "Load 10.250.12.169 FAD to SAS UUT"}


The numbers don't match for the size with the actual size of the files

I downloaded the individual files and check their size, values below:


cbsd    1673
zone    918
esc      10223

thanks,


Javi

TaranFurnell:
I think I know what might have caused that. The files may not necessarily be being stored as UTF-8, let me take a look and get back to you.

jcereced:
sounds good..thanks!

TaranFurnell:
@jcereced This commit produces the correct checksum on my end, please confirm. I've included some changes to fad testing and the fad helper so you can more easily investigate the output of the requests.

To prevent extra encoding errors the server handler has been changed to a ```SimpleHTTPRequestHandler``` so it will simply serve the exact content and encoding of the generated files. I don't believe this is an issue as the test harness does not need to react to the SAS UUT.

jcereced:
Just ran FAD.2 and I'm getting the correct sizes and checksums...fantastic!!

I want to try a few more of the TCs that get a FAD from the TH but gotta run to a soccer game.  I'll check them first thing tomorrow and mark it as approved if all of them run ok (they should)

TaranFurnell:
Great! Merging now.

PR #509, opened 2018-04-05
FW aggregated interference non-DPA: clean PR.

sbdt-google:
Merging issue have rendered the PR #487 unreadable and not cleanly based on the master.
To avoid any issue a new PR is created with the changes made on PR #487, and that old one will be dropped.


jni2000:
Thanks, Serge and Janet. Another great step.

PR #510, opened 2018-04-05
Restore default HTTP timeout

norbertbarbosa:
There is no reason to use a arbitrary 30s timeout...


janetliuboston:
Agree. Thanks, Norbert!

PR #513, opened 2018-04-06
Small fix to SIQ3

gilschieber:
Injecting GWBL rather than GWPZ.


RedaHossino:
could you check that the distance between the FSS and GWBL < 150 Km(as I checked it is not the case)?
if you prefer I can fix that in another PR ?

gilschieber:
I confirm that the distance between the FSL and GWBL is < 150 kms (around 100 kms).
Can you check again on your side?

RedaHossino:
yes, that's right I double check it and it is good, thank you.

PR #514, opened 2018-04-10
fix PCR.1 errors

jni2000:
fix PCR.1 submission time mis-operation errors.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_PCR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_PCR_testcase.py
@@ -99,7 +99,9 @@ def assertRegConditionalsForPpaRefModel(registration_requests,
           # If the cbsdCategory is not present in registration request then
           # assign it to the cbsdCategory in conditional_params.
           if 'cbsdCategory' not in device:
-            break
+            device.update({'cbsdCategory': conditional_params.get('cbsdCategory',
+                                                                  'B')})
+          break
       # Raise an exception if InstallationParam is not found in the conditionals.
       if not install_param_assigned:
         raise Exception("InstallationParam not found in conditionals for device "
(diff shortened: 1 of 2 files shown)


jni2000:
Kate, please review the fixes for the two errors in PCR.1 submission. Thanks.

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -99,7 +99,9 @@ def assertRegConditionalsForPpaRefModel(registration_requests,
           # If the cbsdCategory is not present in registration request then
           # assign it to the cbsdCategory in conditional_params.
           if 'cbsdCategory' not in device:
-            break
+            device.update({'cbsdCategory': conditional_params.get('cbsdCategory',
+                                                                  'B')})
```
kate-harrison: There is no default CBSD category, so you can't use `get`  here.

janetliuboston: FW:
Would this be an acceptable fix, to NOT use a default category, since cbsdCategory seems to be required for conditionalRegistrationData?

Kate:
I think that `get` still fills in `None` in that case (rather than raising an exception) if the field isn't in conditional_params. It seems better to raise an error -- indicating a bad config -- than to silently ignore the problem and let the error be caught later (which I think it would be?).

janetliuboston:
Addressed comment to propagate error if cbsdCategory does not exist in conditionalRegistrationData.

janetliuboston:
Ready for review. Thanks, Kate!

src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -99,7 +99,11 @@ def assertRegConditionalsForPpaRefModel(registration_requests,
           # If the cbsdCategory is not present in registration request then
           # assign it to the cbsdCategory in conditional_params.
           if 'cbsdCategory' not in device:
-            break
+            if not conditional_params.get('cbsdCategory'):
```
kate-harrison: If you just do `conditional_params['cbsdCategory']`, this will automatically raise a KeyError exception. So you can make this a lot simpler :)

janetliuboston: Simpler is better. :) Submitted. Thanks, Kate!

PR #516, opened 2018-04-10
aggregate interference unit tests

janetliuboston:


janetliuboston:
This PR is ready for review. Thanks!

src/harness/reference_models/examples/testutils.py
```diff
@@ -80,3 +84,57 @@ def __call__(self,
...
+
+  Attributes:
+    ant_gain: Effective antenna gain from CBSD towards the protected entity
+              and vice versa. Default value is configured for interference
+              calculation
+    antenna_height: Antenna height considered for incidence angle calculation
+    reference_bw: Aggregate interference is calculated in reference bandwidth
+  """
+
+  def __init__(self,
```
sbdt-google: Would have been nice to be able to pass a FakePropagationPredictor to the constructor, that is saved
in self.propag_predictor:
``` 
def __init__(self,
                  ant_gain=10,
                  reference_bw=5.e6,
                  propag_predictor=None):
      self.antenna_height = antenna_height
      self.ant_gain = ant_gain
      self.reference_bw = reference_bw
      if propag_predictor is None:
         propag_predictor = FakePropagationPredictor()
      self.propag_predictor = propag_predictor
```

src/harness/reference_models/examples/testutils.py
```diff
@@ -80,3 +84,57 @@ def __call__(self,
...
+               reference_bw=5.e6):
+    self.antenna_height = antenna_height
+    self.ant_gain = ant_gain
+    self.reference_bw = reference_bw
+
+  def __call__(self, grant, eirp, constraint, fss_info=None,
+                   esc_antenna_info=None, region_type=None):
+
+    # Calculate the db loss and incidence angles using the FakePropagationPredictor
+    db_loss, incidence_angles, _ = FakePropagationPredictor().__call__(grant.latitude,
```
sbdt-google: and then here:
```
db_loss, incidence_angles, _ = self.propag_predictor(grant.latitude, ...)
```

src/harness/reference_models/interference/aggregate_interference_test.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+  def setUp(self):
+    self.original_interference = interf.computeInterference
+
+  def tearDown(self):
+    interf.computeInterference = self.original_interference
+
+  def test_AggregateInterferenceFssCochannel(self):
+    expected_interference = {34.4116: {-100.57114: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0,\
```
sbdt-google: Did you check by hand these values are expected, or are they the results of a first run ?

Of course if you pass a specific propag_predictor, with L1 distance (abs(lat2-lat1) + abs(lon2-lon1)), easy to know what is the expected pathloss returned, and then verify by hand/excel that the calculation is ok.

Same comment about below tests about making sure that the reference has been validated/computed by hand or in excel sheet, as it now tells you if there is an issue with the current code. The stored reference then provides regression testing for future update.

jni2000: Have   changed the distance type to L1, calculated the path loss value manually and   verified the same with the executed results. Results are available up on request.

jni2000:
Serge, we have addressed your review comments. Please take a look before we merge. Also, if test results are required, let us know where to save them.

sbdt-google:
Thanks. Looks very good to me.
Since you have the test results, just add your excel  (or text file) with the results in the folder, as is.
You can call it  `aggregate_interference_test_calc_sheet.xlsx`  for example.

jni2000:
Serge, added test results in " src/harness/reference_models/interference/test_results/", please take a look.

sbdt-google:
I am fine, I havent checked the calculation detail as of now, but I trust the math works out ok.
Thanks, you can submit

PR #468, opened 2018-03-10
SCS/SDS/SSS.16 implementation based on WG.2 confirmation -- revoked CA is sub-CA

jni2000:
As WinnF WG.2 and ITS approved to use CRL as the initial certificate blacklisting solution before OCSP is available, the test codes developers working on security related test cases came to the following agreements:

1. Deploy a CRL server (a.k.aThe certificate revocation list distribution point – CDP) in the ITS designated VM environment to serve as a FCC certification dedicated CRL server
2. Certificates, sub-CAs to be blacklisted for all the test cases (e.g. SCS.11 and SCS.16 etc.) will be created and per-loaded into the CRL server appropriately. There is NO need for the test harness codes to blacklist a certificate or sub-CA on-demand. During the test, the test codes will directly use the certificate or sub-CA that are already in the CRL.
3. SAS UUT is assumed to pull the CRL periodically as if in normal operation. There is only a coordination need, that is to make sure SAS UUT has the latest CRL information before running the related test cases.

The SCS.16, SDS.16 and SSS.16 test cases are modified to meeting these agreements as below

1.  the revoked CA is sub-CA that will be inserted into CRL
2. the test cases will use CRL for testing the revoked sub-CA requirement.

PR #447, opened 2018-02-28
Update the configurations and the generation script of the certificates

RedaHossino:
- re add the test extention to the certificates
- remove the keyword critical from the test extension
- add section in opennssl config for cbsd not test certificate
- update devices certificates with fcc ID and Cbsd serial number

PR #521, opened 2018-04-11
Update default client certificate to be a DomainProxy certificate.

norbertbarbosa:
As most of the testcases use batch request, it make sense to use a
domain proxy certificate as default certificate.


src/harness/certs/generate_fake_certs.sh
```diff
@@ -115,17 +125,17 @@ openssl ca -cert sas_ca.cert -keyfile private/sas_ca.key -in sas.csr \
...
 
 # Generate normal operation client certificate/key.
-echo "\n\nGenerate 'client' certificate/key"
+echo -e "\n\nGenerate generic 'client' certificate/key based on domainProxy"
 openssl req -new -newkey rsa:2048 -nodes \
     -reqexts cbsd_req -config ../../../cert/openssl.cnf \
     -out client.csr -keyout client.key \
-    -subj "/C=US/ST=District of Columbia/L=Washington/O=Wireless Innovation Forum/OU=www.wirelessinnovation.org/CN=SAS CBSD Example"
-openssl ca -cert cbsd_ca.cert -keyfile private/cbsd_ca.key -in client.csr \
+    -subj "/C=US/ST=District of Columbia/L=Washington/O=Wireless Innovation Forum/OU=www.wirelessinnovation.org/CN=SAS DP Example"
```
RedaHossino: I think the CN should be with the format
 <FRN of certified Domain Proxy>:<unique identifier chosen by owner of FRN>.
where  An FRN, or FCC registration number, is a 10-digit number that is assigned to a business or individual registering with the FCC.
and  I think in the second part "unique identifier chosen by owner of FRN", we can put any value.

norbertbarbosa: I agree, after looking at https://workspace.winnforum.org/higherlogic/ws/public/download/6020/WINNF-TS-0022%20v1.1.2%20CBRS%20PKI%20Certificate%20Policy.pdf that the CN should have this format.
In fact no 'fake' certificate generated by this script follow the subject specification. Should I update all certificate generated by this file or should it be done in another PR?

In that case, which value of FRN should I use? (it seems that there is no 'fake' value defined for test, as 0000000000). 
Same question apply for the FCC ID required for all CBSD certificates: which value should be used?

As the subject and CN is not really used by the SAS ecosystem (except the CN for server defined as localhost), perhaps adding a comment with a reference to this spec will be enough: it will emphasise the fact that all these certificates are fake for test purpose only...

RedaHossino: yes you are right the current script doesn't use correct CNs, but the SAS should verify that cert's CN has the correct format.
for DP cert: as I understood, there is no way by which the SAS can validate the FRN value(and it is not required), so we can put any 10 digits.
In a previous open PR #447 we changed the CN for the cbsd certificates, and in those certificates the FCC Id & serial number are the same as the FCC Id & serial number of the CBSD, please feel free to copy this part from  that PR.

do you agree with that?

kate-harrison: I propose that we use 0123456789 for the FRN as it is an obviously fake value.

norbertbarbosa: Ok: By using https://apps.fcc.gov/coresWeb/simpleSearch.do I verified that the FRN 0123456789 is not used, so I will use this value.

> but the SAS should verify that cert's CN has the correct format.

Not sure about this... As I understand, if you want to enforce a given FCC_ID/ SN or FRN you have to use the corresponding OIDs. The CN is just part of the unique identifier for the certificate, but there is no testcases checking the CN format.

Anyway, I updated the subject part of all certificates to better follow the spec.
Please take another look

RedaHossino: thank you for the update, 
as I understand according to one email thread "Questions about Certificate", the CN are required to exist and to have the correct format.

don't you think it would be better if the update of the revoking part was in another PR, so other people could test it?

I approve the changes related to subject data, I read all the other changes, they look good but I coudn'l test the revoking part.

norbertbarbosa:
The 'revoking part' was done in PR #468. Here I simply merged with the origin/master.

PR #523, opened 2018-04-12
Prop Models: Implement minimum height of 1m

sbdt-google:
As proposed in WG1 specification change (note: not yet integrated in spec), to use minimum of 1m height.
Also add following changes:
 - remove the exception check for maximum height
 - add a flag parameter to specify if CBSD height is a AMSL height. Using this flag will simplify the test harness code logic, as it wont need to do the test for AMSL and correction in various places. 

Background:
The models are not designed for negative height, and would return bogus results.
Normally AGL height should always be positive (except for CBSD in basement or tunnels or subway), although an error in configuration would go through the SAS-CBSD protocol. 
Problematic also is the height being specified as AMSL. This AMSL would be transformed into an AGL height by removing the terrain height coming from the terrain database, which may be inaccurate, and thus leading to a negative height for purpose of propagation modeling.

To avoid having the bogus result preventing a CBSD interference to be taken into account during incumbents protection, a minimum height of 1m is taken into account.
Most likely, valid underground CBSD will be handled as part of a future extension of the SAS specification, with a specific modeling and specific flag for categorizing those

### DIFF
--- a/src/harness/reference_models/propagation/wf_hybrid.py
+++ b/src/harness/reference_models/propagation/wf_hybrid.py
@@ -160,6 +160,8 @@ def CalcHybridPropagationLoss(lat_cbsd, lon_cbsd, height_cbsd,
                           Value in [0,1]: returns the CDF quantile
                           -1: returns the mean path loss
     region:             Region type among 'URBAN', 'SUBURBAN, 'RURAL'
+    is_height_cbsd_amsl: If True, the CBSD height shall be considered as AMSL (Average
+                         mean sea level).
 
   Returns:
     A namedtuple of:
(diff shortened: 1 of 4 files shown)


jni2000:
I have reviewed the general logic and they looks Google. We are in process of another round of reviewing of the functions.

sbdt-google:
Ok let me know, the change are pretty simple, and in agreement with the change in spec done recently.

jni2000:
Serge, Mo has reviewed the functional portion and all looks good. Please merge in.

sbdt-google:
Thanks. submitted

PR #530, opened 2018-04-16
Update FAD_1 with accepted  TODOs

RedaHossino:
update fad 1 test with the official accepted TODOs(so far):
- add the possibility to have PAL grant in the config
- add check of PAL&PPA of the config file
- update check of PPA CBSD Ids
- add check for winding of geoJSON contour of ppa record.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_FAD_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_FAD_testcase.py
@@ -18,7 +18,7 @@ import json
 import os
 import sas
 import sas_testcase
-from reference_models.geo import vincenty
+from reference_models.geo import vincenty, utils
 from sas_test_harness import SasTestHarnessServer, generateCbsdRecords, \
     generateCbsdReferenceId, generatePpaRecords
 
@@ -178,7 +178,7 @@ class FullActivityDumpTestcase(sas_testcase.SasTestCase):
                             ['operationFrequencyRange']['highFrequency'], 3700000000)
           self.assertEqual( grants_of_cbsd[0]['requestedOperationParam']\
                             ['operationFrequencyRange']['highFrequency'] % 5000000, 0) 
-          # check grant OperationParam    
+          # check grant OperationParam   
           self.assertDictEqual( grants_of_cbsd[0]['operationParam'], \
                                 grant_request[index]['operationParam'])
 
@@ -229,6 +229,10 @@ class FullActivityDumpTestcase(sas_testcase.SasTestCase):
                                                               pal_low_frequency,
                                                               pal_high_frequency,
                                                               device_a['userId'])
+
+      device_a['installationParam']['latitude'], device_a['installationParam'][
+        'longitude'] = getRandomLatLongInPolygon(ppa)
+      ppa['cbsdClusterList'] = [{'cbsdSerialNumber': device_a['cbsdSerialNumber'] , 'fccId': device_a['fccId']}]
       ppas.append(ppa)
       pals.extend(pal)
       # ESC senosrs
@@ -285,27 +289,61 @@ class FullActivityDumpTestcase(sas_testcase.SasTestCase):
       # Register N1 CBSDs.
       request = {'registrationRequest': config['registrationRequests']}
       responses = self._sas.Registration(request)['registrationResponse']
+
       # Check registration responses and get cbsd Id
       self.assertEqual(len(responses), len(config['registrationRequests']))
-      # TODO : to update the test so that it is possible to get a PAL Grant
+      cbsd_ids = []
+      for response in responses:
+        cbsd_ids.append(response['cbsdId'])
+      cbsdIds_of_pal_grant = []
+      # inject PALs and N2 PPAs
+      ppa_ids = []       
+      for pal in config['palRecords']:
+          self._sas_admin.InjectPalDatabaseRecord(pal)                  
+      for ppa in config['ppaRecords']:
+        ppa['ppaInfo']['cbsdReferenceId'] = []
+        for cluster_list in ppa['cbsdClusterList']:
+          for index, cbsd in enumerate(config['registrationRequests']):
+            if cluster_list['fccId'] == cbsd['fccId'] and  cluster_list['cbsdSerialNumber'] == cbsd ['cbsdSerialNumber']:
+              ppa['ppaInfo']['cbsdReferenceId'].append(cbsd_ids[index])
+              break
+        del ppa['cbsdClusterList']
+        ppa_ids.append(self._sas_admin.InjectZoneData({'record': ppa}))
       grants = config['grantRequests']
       for index, response in enumerate(responses):
           self.assertEqual(response['response']['responseCode'], 0)
           grants[index]['cbsdId'] = response['cbsdId']
+          grant_frequency_range = grants[index]['operationParam']['operationFrequencyRange']
+          # Get the cbsdIds whose grant is PAL
+          for ppa in config['ppaRecords']:
+                 if 'cbsdReferenceId' in ppa['ppaInfo'] and  response['cbsdId'] in ppa['ppaInfo']['cbsdReferenceId']:
+                   frequency_ranges_of_pals = [pal['channelAssignment']['primaryAssignment'] \
+                     for pal in config['palRecords'] if pal['palId'] in config['palRecords']]
+                   self.assertTrue(pals_of_ppa)                 
+                   low_freq = min([freq_range['lowFrequency'] for freq_range in frequency_ranges_of_pals])
+                   high_freq = max([freq_range['highFrequency'] for freq_range in frequency_ranges_of_pals])
+                   # check that the codnfig file contains consistent PAL&PPA data
+                   self.assertChannelsContainFrequencyRange( {'lowFrequency': low_freq, 'highFrequency':high_freq }, frequency_ranges_of_pals)
+                   # check if the PAL frequency range contains the grant frequency range
+                   if grant_frequency_range['lowFrequency'] >= low_freq and grant_frequency_range['highFrequency'] <= high_freq:
+                     cbsdIds_of_pal_grant.append(response['cbsdId'])
+                   # check that the grant in config file is not mixed of PAL and GAA channels
+                   elif grant_frequency_range['lowFrequency'] <= low_freq:
+                     self.assertLessEqual(grant_frequency_range['highFrequency'], low_freq)
+                   else:
+                     self.assertGreaterEquel(grant_frequency_range['lowFrequency'], high_freq)
       # send grant request with N1 grants
       del responses
       grant_responses = self._sas.Grant({'grantRequest': grants})['grantResponse']
       # check grant response
       self.assertEqual(len(grant_responses), len(config['grantRequests']))
       for grant_response in grant_responses:
-          self.assertEqual(grant_response['response']['responseCode'], 0)       
-      # inject PALs and N2 PPAs
-      ppa_ids = []       
-      for pal in config['palRecords']:
-          self._sas_admin.InjectPalDatabaseRecord(pal)
-                       
-      for ppa in config['ppaRecords']:
-          ppa_ids.append(self._sas_admin.InjectZoneData({'record': ppa}))          
+          self.assertEqual(grant_response['response']['responseCode'], 0)
+          # check grant channel type
+          if grant_response['cbsdId'] in cbsdIds_of_pal_grant:
+            self.assertEqual('PAL', grant_response['channelType'])
+          else:
+            self.assertEqual('GAA', grant_response['channelType'])       
       # inject N3 Esc sensor
       for esc_sensor in config['escSensorRecords']:
           self._sas_admin.InjectEscSensorDataRecord({'record': esc_sensor})
@@ -352,18 +390,24 @@ class FullActivityDumpTestcase(sas_testcase.SasTestCase):
         self.assertEqual(ppa_record['id'].split("/")[0], 'zone')
         self.assertEqual(ppa_record['id'].split("/")[1], 'ppa')
         self.assertEqual(ppa_record['id'].split("/")[2], self._sas._sas_admin_id)
-        del ppa_record['id']
-        # TODO : to update the following check according when PAL grants are supported in this test case
-        if 'cbsdReferenceId' in ppa_record['ppaInfo']:
-          self.assertFalse(ppa_record['ppaInfo']['cbsdReferenceId'])
-          del ppa_record['ppaInfo']['cbsdReferenceId']
+        del ppa_record['id'] 
         # verify that the injected ppas exist in the dump files
-        # TODO: add Check for GeoJSON Winding of ppa polygon  
+        # check GeoJson Winding of PPA record
+        utils.HasCorrectGeoJsonWinding(ppa_record['zone']['features'][0]['geometry'])     
         exist_in_dump = False
         for ppa in config['ppaRecords']:
           if 'id' in ppa:
-            del ppa['id']        
-          exist_in_dump = exist_in_dump or areTwoPpasEqual(ppa_record, ppa)
+            del ppa['id'] 
+            # we exclude the values of the CBSD reference Id, because they are not necessarily the same
+            ppa_cbsd_reference_ids = ppa['cbsdReferenceId']
+            ppa_record_cbsd_reference_ids = ppa_record['cbsdReferenceId']
+            del ppa['cbsdReferenceId']
+            del ppa_record['cbsdReferenceId']
+            exist_in_dump = exist_in_dump or areTwoPpasEqual(ppa_record, ppa)
+            # check that the record has the same length of the CBSD reference Ids
+            exist_in_dump &= len(ppa_cbsd_reference_ids) == len(ppa_record_cbsd_reference_ids)
+            ppa['cbsdReferenceId'] = ppa_cbsd_reference_ids
+            ppa_record['cbsdReferenceId'] = ppa_record_cbsd_reference_ids
           if exist_in_dump:
             break
         self.assertTrue(exist_in_dump)     


src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -229,6 +229,10 @@ def generate_FAD_1_default_config(self, filename):
                                                               pal_low_frequency,
                                                               pal_high_frequency,
                                                               device_a['userId'])
+
+      device_a['installationParam']['latitude'], device_a['installationParam'][
+        'longitude'] = getRandomLatLongInPolygon(ppa)
+      ppa['cbsdClusterList'] = [{'cbsdSerialNumber': device_a['cbsdSerialNumber'] , 'fccId': device_a['fccId']}]
```
kate-harrison: Please implement consistent with GRA.4.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -285,27 +289,61 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
       request = {'registrationRequest': config['registrationRequests']}
       responses = self._sas.Registration(request)['registrationResponse']
+
       # Check registration responses and get cbsd Id
       self.assertEqual(len(responses), len(config['registrationRequests']))
-      # TODO : to update the test so that it is possible to get a PAL Grant
+      cbsd_ids = []
+      for response in responses:
+        cbsd_ids.append(response['cbsdId'])
+      cbsdIds_of_pal_grant = []
```
kate-harrison: Please use `snake_case` consistently rather than mixing with `CamelCase`.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -285,27 +289,61 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
+              ppa['ppaInfo']['cbsdReferenceId'].append(cbsd_ids[index])
+              break
+        del ppa['cbsdClusterList']
+        ppa_ids.append(self._sas_admin.InjectZoneData({'record': ppa}))
       grants = config['grantRequests']
       for index, response in enumerate(responses):
           self.assertEqual(response['response']['responseCode'], 0)
           grants[index]['cbsdId'] = response['cbsdId']
+          grant_frequency_range = grants[index]['operationParam']['operationFrequencyRange']
+          # Get the cbsdIds whose grant is PAL
```
kate-harrison: Can't you just compare the `channelType` in the `grantResponse` with the `channelType` in the FAD? We have other tests to check that a PAL grant is issued correctly, so we don't need to double-check here.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -352,18 +390,24 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
+        # check GeoJson Winding of PPA record
+        utils.HasCorrectGeoJsonWinding(ppa_record['zone']['features'][0]['geometry'])     
         exist_in_dump = False
         for ppa in config['ppaRecords']:
           if 'id' in ppa:
-            del ppa['id']        
-          exist_in_dump = exist_in_dump or areTwoPpasEqual(ppa_record, ppa)
+            del ppa['id'] 
+            # we exclude the values of the CBSD reference Id, because they are not necessarily the same
+            ppa_cbsd_reference_ids = ppa['cbsdReferenceId']
```
kate-harrison: This should be un-indented once, right? Otherwise this code will not work if you have more than one PPA.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -352,18 +390,24 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
+            ppa_record_cbsd_reference_ids = ppa_record['cbsdReferenceId']
+            del ppa['cbsdReferenceId']
+            del ppa_record['cbsdReferenceId']
+            exist_in_dump = exist_in_dump or areTwoPpasEqual(ppa_record, ppa)
+            # check that the record has the same length of the CBSD reference Ids
+            exist_in_dump &= len(ppa_cbsd_reference_ids) == len(ppa_record_cbsd_reference_ids)
```
kate-harrison: You should be able to check the values of the CBSD reference IDs as well. They are deterministically generated based on the FCC ID and SN.

RedaHossino: in admin testing API doc there is a note for inject zone(PPA) : the cbsdReferenceId field in the PpaInformation object provided through this API will contain CBSD IDs, not the corresponding CBSD Reference IDs as defined in the SAS-SAS TS. Depending on the SAS's implementation, these may or may not have the same format.

is the note is still relevant to this issue? or should SAS UUT set the CBSD reference ID in the dump the same as the injected one with the PPA?

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
RedaHossino: thank you for you comments, I updated the code.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
RedaHossino: it is a good remark!
I kept the check for consistent PPA&PAL in the config file at the begning of the test, if you think it is not necessary, I can delete it.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
RedaHossino: thnx, fixed.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -273,7 +276,22 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
       # Very light checking of the config file.
       self.assertEqual(len(config['registrationRequests']),
                         len(config['grantRequests']))
-        
+      # check that the codnfig file contains consistent PAL&PPA data
```
kate-harrison: Small typo: config

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -273,7 +276,22 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
+                for pal in config['palRecords'] if pal['palId'] in ppa['ppaRecord']['ppaInfo']['palId']]
+              self.assertTrue(frequency_ranges_of_pals)
+              low_freq = min([freq_range['lowFrequency'] for freq_range in frequency_ranges_of_pals])
+              high_freq = max([freq_range['highFrequency'] for freq_range in frequency_ranges_of_pals])
+              self.assertChannelsContainFrequencyRange( {'lowFrequency': low_freq, 'highFrequency':high_freq }, frequency_ranges_of_pals)
+              # check that the grant in config file is not mixed of PAL and GAA channels
+              if low_freq <= grant_frequency_range['lowFrequency'] <= high_freq:
+                 self.assertLessEqual(grant_frequency_range['highFrequency'], high_freq)
+              if low_freq <= grant_frequency_range['highFrequency'] <= high_freq:
+                 self.assertGreaterEquel(grant_frequency_range['lowFrequency'], low_freq)
```
kate-harrison: If this code remains, there is a typo in `Equal`.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
kate-harrison: Thanks!

I like the idea of checking the config more carefully, but unfortunately I'm not sure that the checks are valid. For example, we could configure the test case to give a GAA Grant to a device which is on the PPA cluster list, and that would be a completely valid config (and in fact it might be a good test of the SAS).

I'm not sure what checks you *can* do, other than some obvious ones:
 - Grants are for valid channels
 - CBSD X doesn't belong to two different PPA cluster lists
 - Each PPA has a cluster list (?)

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
kate-harrison: That note is still accurate. I think what you need to do is create each PPA's cluster list *including the CBSD reference IDs*. Right now we have the index of each CBSD, so you just have to pull the right reg request, get the FCC ID and serial number, then use a helper function to generate the CBSD reference ID from the FCC ID, SN. Then you will have the same format that the SAS is expected to provide and I think your existing comparison function will work correctly.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -395,21 +384,22 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
         # check GeoJson Winding of PPA record
         utils.HasCorrectGeoJsonWinding(ppa_record['zone']['features'][0]['geometry'])     
         exist_in_dump = False
-        for ppa in config['ppaRecords']:
+        for ppaConf in config['ppaRecords']:
```
kate-harrison: Please use `snake_case` instead of `camelCase`.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -395,21 +384,22 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
-            del ppa['cbsdReferenceId']
-            del ppa_record['cbsdReferenceId']
-            exist_in_dump = exist_in_dump or areTwoPpasEqual(ppa_record, ppa)
-            # check that the record has the same length of the CBSD reference Ids
-            exist_in_dump &= len(ppa_cbsd_reference_ids) == len(ppa_record_cbsd_reference_ids)
-            ppa['cbsdReferenceId'] = ppa_cbsd_reference_ids
-            ppa_record['cbsdReferenceId'] = ppa_record_cbsd_reference_ids
-          if exist_in_dump:
-            break
+          # we exclude the values of the CBSD reference Id, because they are not necessarily the same
```
kate-harrison: See my other comment on this: I think it is actually possible to check these values here.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
RedaHossino: the check here is only to verify the following:
- the PAL channels of one PPA form  a frequencyRange correctly.
- if the CBSD is in the cluster list of a PPA then the grant  frequency range should be completely inside the frequency range of the PPA or completely outside of it, so it should not partially overlap with it.

don't you agree with that?

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -276,7 +276,7 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
       # Very light checking of the config file.
       self.assertEqual(len(config['registrationRequests']),
                         len(config['grantRequests']))
-      # check that the codnfig file contains consistent PAL&PPA data 
+      # check that the config file contains consistent PAL&PPA data
```
kate-harrison: Please see my other comment -- I don't think we can actually require that the CBSDs on the cluster list get a PAL grant: the CBSD could also ask for a GAA grant on a non-PAL channel and that would be valid.

RedaHossino: I replied to your previous comment, I will copy it here: 

the checks here are to verify the following:
- the PAL channels of one PPA form a frequencyRange correctly.
- if the CBSD is in the cluster list of a PPA then the grant frequency range should be completely inside the frequency range of the PPA or completely outside of it, so it should not partially overlap with it.

don't you agree with that?

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -269,7 +276,22 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
                         len(config['grantRequests']))
-        
+      # check that the config file contains consistent PAL&PPA data 
+      for index, grant in enumerate(config['grantRequests']):
+        grant_frequency_range = grant['operationParam']['operationFrequencyRange']
+        for ppa in config['ppaRecords']:
+            if index in ppa['ppaClusterList']:
+              frequency_ranges_of_pals = [{'frequencyRange' : {'lowFrequency' : pal['channelAssignment']['primaryAssignment']['lowFrequency'], 'highFrequency' : pal['channelAssignment']['primaryAssignment']['highFrequency']}} \
+                for pal in config['palRecords'] if pal['palId'] in ppa['ppaRecord']['ppaInfo']['palId']]
+              self.assertTrue(frequency_ranges_of_pals)
```
kate-harrison: If you're trying to do a check on the length, it's probably better to do that directly.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -269,7 +276,22 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
+              low_freq = min([freq_range['lowFrequency'] for freq_range in frequency_ranges_of_pals])
+              high_freq = max([freq_range['highFrequency'] for freq_range in frequency_ranges_of_pals])
+              self.assertChannelsContainFrequencyRange( {'lowFrequency': low_freq, 'highFrequency':high_freq }, frequency_ranges_of_pals)
```
kate-harrison: I think instead you can check directly that all lowFreqs are the same and all highFreqs are the same.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
kate-harrison: Sorry, I must have missed that comment somehow. I'll have to also admit now that I did not read the code super carefully because I assumed I knew what it was doing (sorry!).

I think the checks you've described are valid. I'd suggest using the `msg` param of the assert functions to tell the tester what's wrong with the config, which will serve the dual purpose of telling anyone who's reading the code what it's supposed to do (more quickly than reading the code).

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -285,9 +318,39 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
       request = {'registrationRequest': config['registrationRequests']}
       responses = self._sas.Registration(request)['registrationResponse']
+
       # Check registration responses and get cbsd Id
       self.assertEqual(len(responses), len(config['registrationRequests']))
-      # TODO : to update the test so that it is possible to get a PAL Grant
+      cbsd_ids = []
+      for index, response in enumerate(responses):
+        # in case that SAS reject registration request because the azimuth is not provided
+        if response['response']['responseCode'] != 0:
```
kate-harrison: Wouldn't this indicate an invalid config? In any case, I don't think we can make this change as it directly contradicts step 2 of the test spec.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -386,10 +442,20 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
+      for sas_th in config['sasTestHarnessConfigs'][1:]:
+        dump_message = self._sas.GetFullActivityDump(sas_th['serverCert'], sas_th['serverKey'])
+        # check that dump message is the same as the message retreived by the first SAS TH
+        self.assertEqual(response, dump_message)
+        # check that dump files are the same as the files retreived by the first SAS TH
+        for index, dump_file in enumerate(dump_message['files']):
+          if dump_file['recordType'] != 'coordination':                
+              downloaded_file = self._sas.DownloadFile(dump_file['url'],\
+                sas_th['serverCert'], sas_th['serverKey'])
+              self.assertEqual(hash_of_dump_file[index], hashlib.sha1(downloaded_file).hexdigest())
```
kate-harrison: I think it's risky to require that the SAS provide the files in exactly the same order each time. Maybe it's better to store a dictionary where the key is the URL used to retrieve the file and the value is the hash of the file (and add a note that a SAS which e.g. simply re-orders the data within the files may falsely fail this test case, just in case someone runs into trouble with that).

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
RedaHossino: yes, you are right, I updated the code, but then we are adding an additional constraints for the CBSD config file(to have always the azimuth) which is not necessary true for all the SASes, right?

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
RedaHossino: yes, that's true, I updated the code, thnx.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
kate-harrison: Hm, I guess the right behavior might be to give a warning in this case that the CBSD registration may be rejected, but continue to registration anyway. If we also add a message on line 337 (where we assert that registration was successful) saying something like "For a Cat A device, it is acceptable to reject a registration request which does not include self-consistent azimuth and beamwidth information. If this applies to the rejected registration request, please remove it from the config (or alter the antenna information) and retry."

kate-harrison: What do you think?

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -398,7 +398,7 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
               ppa_dump_data.extend(downloaded_file['recordData'])
           else:
               self.assertEqual('coordination', dump_file['recordType'])
-          hash_of_dump_file.append(hashlib.sha1(downloaded_file).hexdigest() if dump_file is not None else None)
+          hash_of_dump_file.append({'url':dump_file['url'], 'hashValue' : hashlib.sha1(downloaded_file).hexdigest()} if dump_file is not None else None)
```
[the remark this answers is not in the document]
RedaHossino: yes, that's better, I updated the code, thnx.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
RedaHossino: but we cann't allow some configs which don't work for all the "correct" SASes, and they depend on a choice in the SAS implemntation, I think we should allow only the configs which work for sure on all the correct SASes, don't you agree?

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -255,6 +255,7 @@ def generate_FAD_1_default_config(self, filename):
       sas_test_harness_1_config = {
         'sasTestHarnessName': 'SAS-TH-2',
         'hostName': 'localhost',
+        'url': 'localhost',
```
kate-harrison: I don't think this is part of the standard config.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -290,26 +291,33 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
...
                  self.assertLessEqual(grant_frequency_range['highFrequency'], high_freq, 'incorrect high frequency of the grant with index {0}, makes it GAA&PAL Mixed Grant'.format(index))
               if low_freq <= grant_frequency_range['highFrequency'] <= high_freq:
                  self.assertGreaterEqual(grant_frequency_range['lowFrequency'], low_freq, 'incorrect low frequency of the grant with index {0}, makes it a GAA&PAL Mixed Grant'.format(index))
       for index, device  in enumerate(config['registrationRequests']):
         # check azimuth in the CBSD config, if the beamwidth is not 0 or 360 and the azimuth is not provided, CBSD registration may be rejected
-        reg_conditional_installation_param = config['conditionalRegistrationData']['installationParam']
+        reg_conditional_device_data_list = [reg for reg in \
+              config['conditionalRegistrationData']['registrationData'] if reg['fccId'] == device['fccId'] and \
+              reg['cbsdSerialNumber'] == device['cbsdSerialNumber'] ]
+        if len(reg_conditional_device_data_list) == 1:
```
kate-harrison: Consider asserting that the length is <= 1 to catch config errors.

src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -398,7 +406,7 @@ def test_WINNF_FT_S_FAD_1(self, config_filename):
               ppa_dump_data.extend(downloaded_file['recordData'])
           else:
               self.assertEqual('coordination', dump_file['recordType'])
-          hash_of_dump_file.append({'url':dump_file['url'], 'hashValue' : hashlib.sha1(downloaded_file).hexdigest()} if dump_file is not None else None)
+          hash_of_dump_file[dump_file['url']] =  hashlib.sha1(downloaded_file).hexdigest() if dump_file is not None else None
```
kate-harrison: In what situations will `dump_file` be `None`? If it can happen here, then it can happen below (line 466) so your hash generation methodology should be the same in both places.

src/harness/sas_test_harness.py
```diff
@@ -369,7 +369,7 @@ def generateCbsdRecords(registration_requests, grant_requests_list):
                                    registration_request['cbsdSerialNumber'])
 
     internal_data = {
-        'id': cbsd_reference_id,
+        'id': 'cbsd/' + cbsd_reference_id,
```
kate-harrison: These changes make sense (thank you). Is this the only place where `generateCbsdReferenceId` is called?

PR #535, opened 2018-04-18
Sets antennaAzimuth, which is a required field.

piotrfilipiuk:
See winnforum_sas/schema/InstallationParamEsc.schema.json for details.


janetliuboston:
Thanks for the addition to match the SAS-SAS TS. Looks good.

PR #536, opened 2018-04-18
SSS.18: fix certificate injection

janetliuboston:
Fix certificate injection to be from config file for SSS.18 instead of default certificate. Otherwise, the SAS peer injected and the FAD request to that SAS peer could have different certificates.


janetliuboston:
Thank you!

PR #542, opened 2018-04-20
Clean legacy data

sbdt-google:
Remove obsolete data, or move them in 'legacy' folder.
Also for the obsolete scripts moved to a legacy folder.

May be removed definitely later


sbdt-google:
Thanks, submitted.

PR #544, opened 2018-04-20
Generate fake certs for peer SASs and DPs used in tests xPR

gilschieber:
Will also update FPR tests in PR#543.


jni2000:
we will review this.

PR #554, opened 2018-04-24
Clean certificate naming

norbertbarbosa:
Naming of certificate, especially the 'client' became confusing.
This PR update certificate naming:
- normalize naming
- remove 'client' name, as confusing after PR #521: now clearly identify 'device' vs 'domain_proxy'
- fix broken SCS test based on the 'client'
- update testcases to use sas.GetDefaultDomainProxySSLCertPath() directly
- transpose change of canceled PR #552
- fix missing SasReset() call in assertTlsHandshakeFailureOrHttp403()
- fix broken fake_sas.py


src/harness/certs/revoke_and_generate_crl.sh
```diff
@@ -56,18 +56,18 @@ function generate_crl_chain()
   openssl ca -gencrl -keyfile private/blacklisted_cbsd_ca.key -cert blacklisted_cbsd_ca.cert \
       -config ../../../cert/openssl.cnf -crldays 365 \
       -out crl/blacklisted_cbsd_ca.crl
-  echo -e "\n\n Generate CRL for blacklisted_sas_ca"
-  openssl ca -gencrl -keyfile private/blacklisted_sas_ca.key -cert blacklisted_sas_ca.cert \
+  echo -e "\n\n Generate CRL for sas_blacklisted_ca"
+  openssl ca -gencrl -keyfile private/sas_blacklisted_ca.key -cert sas_blacklisted_ca.cert \
```
assaneRed: "sas_blacklisted_ca" or "revoked_sas_ca"

src/harness/certs/revoke_and_generate_crl.sh
```diff
@@ -56,18 +56,18 @@ function generate_crl_chain()
   openssl ca -gencrl -keyfile private/blacklisted_cbsd_ca.key -cert blacklisted_cbsd_ca.cert \
```
assaneRed: change "blacklisted_cbsd_ca" or "revoked_cbsd_ca"

src/harness/certs/revoke_and_generate_crl.sh
```diff
@@ -56,18 +56,18 @@ function generate_crl_chain()
...
       -out crl/blacklisted_cbsd_ca.crl
-  echo -e "\n\n Generate CRL for blacklisted_sas_ca"
-  openssl ca -gencrl -keyfile private/blacklisted_sas_ca.key -cert blacklisted_sas_ca.cert \
+  echo -e "\n\n Generate CRL for sas_blacklisted_ca"
+  openssl ca -gencrl -keyfile private/sas_blacklisted_ca.key -cert sas_blacklisted_ca.cert \
       -config ../../../cert/openssl.cnf -crldays 365 \
-      -out crl/blacklisted_sas_ca.crl
+      -out crl/sas_blacklisted_ca.crl
   echo -e "\n\n Generate CRL for blacklisted_proxy_ca"
   openssl ca -gencrl -keyfile private/blacklisted_proxy_ca.key -cert blacklisted_proxy_ca.cert \
```
assaneRed: change "blacklisted_proxy_ca" or "revoked_proxy_ca"

src/harness/certs/generate_fake_certs.sh
```diff
@@ -139,18 +139,17 @@ openssl ca -cert sas_ca.cert -keyfile private/sas_ca.key -in sas_1.csr \
...
 openssl req -new -newkey rsa:2048 -nodes \
-    -reqexts oper_req -config ../../../cert/openssl.cnf \
-    -out client.csr -keyout client.key \
-    -subj "/C=US/O=Wireless Innovation Forum/OU=WInnForum Domain Proxy Certificate/CN=0123456789:0001"
-openssl ca -cert proxy_ca.cert -keyfile private/proxy_ca.key -in client.csr \
-    -out client.cert -outdir ./root \
-    -policy policy_anything -extensions oper_req_sign -config ../../../cert/openssl.cnf \
+    -reqexts cbsd_req -config ../../../cert/openssl.cnf \
+    -out device.csr -keyout device.key \
+    -subj "/C=US/O=Wireless Innovation Forum/OU=WInnForum CBSD Certificate/CN=test_fcc_id:test_serial_number"
```
assaneRed: I think you need to change CN=test_fcc_id:test_serial_number" to CN="test_fcc_id_a:test_serial_number_a"
or create new device with fcc_id="test_fcc_id" and serial_number="test_serial_number" and change all security testcases because all these tests use device_a

norbertbarbosa: This 'device.cert' was supposed to be the default generic certificate for cbsd, that is not only used for device_a, but potentially for any device.

I consider the CN part as purely informative: if we want to 'lock' a certificate to a given CBSD we have to use the OID FCCID and SERIAL and not be tied on a particular formating of the CN.
Because these OID are optional in the spec, it is possible to define this kind of 'generic' certificate.
Do you agree with this?

To go back to this 'device.cert', as it seems used only for SCS tests, I did the following:
    - remove the 'device.cert'
    - in sas.py, remove GetDefaultCbsdSSLCertPath() + Key 
    - in security_testcase.py: in doCbsdTestCipher(): clearly indicate that the default cert is 'device_a'  
    - update the generation of device_a and device_c by adding the OID FCCID and SERIAL

src/harness/certs/generate_fake_certs.sh
```diff
@@ -215,15 +205,15 @@ openssl ca -cert unrecognized_root_ca.cert -keyfile private/unrecognized_root_ca
...
+echo -e "\n\nGenerate 'device_corrupted' certificate/key"
+gen_corrupt_cert device_a.key device_a.cert device_corrupted.key device_corrupted.cert
 
 # Certificate for test case WINNF.FT.S.SCS.8 - Self-signed certificate presented during registration
 # Using the same CSR that was created for normal operation
-echo -e "\n\nGenerate 'self_signed_client' certificate/key"
-openssl x509 -signkey client.key -in client.csr \
-    -out self_signed_client.cert \
+echo -e "\n\nGenerate 'device_self_signed' certificate/key"
+openssl x509 -signkey device.key -in device.csr \
```
assaneRed: device_a or device?

cert/openssl.cnf
```diff
@@ -509,6 +509,30 @@ crlDistributionPoints=@crl_section
...
+subjectKeyIdentifier = hash
+authorityKeyIdentifier = keyid:always,issuer
+basicConstraints = CA:FALSE
+keyUsage = critical, digitalSignature, keyEncipherment, cRLSign
+extendedKeyUsage = clientAuth
+certificatePolicies=@cps,ROLE_CBSD
+crlDistributionPoints=@crl_section
+FCCID=ASN1:UTF8String:test_fcc_id_c
+SERIAL=ASN1:UTF8String:test_serial_number_c
+
```
assaneRed: I don't think you need to add these two req_sign (cbsd_req_device_a_sign and cbsd_req_device_c_sign), FCCID and SERIAL are optionals
Why you don't use cbsd_req_sign?

src/harness/certs/generate_fake_certs.sh
```diff
@@ -329,20 +319,20 @@ openssl ca -gencrl -keyfile private/cbsd_ca.key -cert cbsd_ca.cert \
...
+openssl ca -cert cbsd_ca.cert -keyfile private/cbsd_ca.key -in device_expired.csr \
+    -out device_expired.cert -outdir ./root \
     -policy policy_anything -extensions cbsd_req_sign -config ../../../cert/openssl.cnf \
     -batch -notext -create_serial -utf8 -startdate 20150214120000Z -enddate 20160214120000Z -md sha384
 
 # Certificate for test case WINNF.FT.S.SCS.15 - Certificate with inapplicable fields presented during registration
 echo -e "\n\nGenerate 'inapplicable certificate for WINNF.FT.S.SCS.15' certificate/key"
-openssl ca -cert cbsd_ca.cert -keyfile private/cbsd_ca.key -in client.csr \
-    -out client_inapplicable.cert -outdir ./root \
+openssl ca -cert cbsd_ca.cert -keyfile private/cbsd_ca.key -in device.csr \
```
assaneRed: change device.csr to device_a.csr

src/harness/certs/generate_fake_certs.sh
```diff
@@ -464,10 +454,10 @@ openssl ca -cert non_cbrs_root_signed_sas_ca.cert -keyfile private/non_cbrs_root
     -batch -notext -create_serial -utf8 -days 1185 -md sha384
 
 # Certificate for test case WINNF.FT.S.SSS.10 - Certificate of wrong type presented by SAS Test Harness 
-# Creating a wrong type certificate by reusing the client.csr and creating a client certificate.
+# Creating a wrong type certificate by reusing the device.csr and creating a client certificate.
 echo -e "\n\nGenerate wrong type certificate/key"
-openssl ca -cert cbsd_ca.cert -keyfile private/cbsd_ca.key -in client.csr \
-    -out wrong_type_sas.cert -outdir ./root \
+openssl ca -cert cbsd_ca.cert -keyfile private/cbsd_ca.key -in device.csr \
```
assaneRed: change device.csr to device_a csr

src/harness/security_testcase.py
```diff
@@ -269,6 +269,7 @@ def assertTlsHandshakeFailureOrHttp403(self, client_cert=None, client_key=None,
     try:
       self.assertTlsHandshakeFailure(client_cert, client_key, ciphers, ssl_method)
```
assaneRed: You have two choise:
1) change 
     client_cert = client_cert or sas.GetDefaultCbsdSSLCertPath()
     client_key = client_key or sas.GetDefaultCbsdSSLKeyPath()
in    def assertTlsHandshakeFailure(self, client_cert=None, client_key=None, ciphers=None, ssl_method=None):

to
     client_cert = client_cert or self.getCertFilename('device_a.cert')
     client_key = client_key or self.getCertFilename('device_a.key')
or keeping sas.GetDefaultCbsdSSLCertPath() and sas.GetDefaultCbsdSSLKeyPath()   and load device_a.cert and device_a.key for those function

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -116,8 +116,8 @@ def generate_SCS_8_default_config(self, filename):
     # Create the actual config for client cert/key path
 
     config = {
-      'clientCert': self.getCertFilename("self_signed_client.cert"),
-      'clientKey': self.getCertFilename("client.key")
+      'clientCert': self.getCertFilename("device_self_signed.cert"),
+      'clientKey': self.getCertFilename("device.key")
```
assaneRed: change device.key to device_a.key or don't remove the generation of device

src/harness/testcases/WINNF_FT_S_SCS_testcase.py
```diff
@@ -251,8 +251,8 @@ def generate_SCS_15_default_config(self, filename):
     # Create the actual config for client cert/key path
     
     config = {
-        'clientCert': self.getCertFilename("client_inapplicable.cert"),
-        'clientKey': self.getCertFilename("client.key")
+        'clientCert': self.getCertFilename("device_inapplicable.cert"),
+        'clientKey': self.getCertFilename("device.key")
```
assaneRed: change device.key to device_a.key or don't remove the generation of device

src/harness/sas.py
```diff
@@ -101,8 +103,8 @@ def _CbsdRequest(self, method_name, request, ssl_cert=None, ssl_key=None):
     return RequestPost('https://%s/%s/%s' % (self.cbsd_sas_active_base_url, self.cbsd_sas_version,
                                              method_name), request,
                        self._tls_config.WithClientCertificate(
-                           ssl_cert or GetDefaultCbsdSSLCertPath(),
-                           ssl_key or GetDefaultCbsdSSLKeyPath()))
+                           ssl_cert or GetDefaultDomainProxySSLCertPath(),
+                           ssl_key or GetDefaultDomainProxySSLKeyPath()))
```
assaneRed: GetDefaultDomainProxySSLCertPath() use domain_proxy(cert/key).
I think you need to define those two methods
def GetDefaultCbsdSSLCertPath():
  return os.path.join('certs', 'device_a.cert')

def GetDefaultCbsdSSLCertPath():
  return os.path.join('certs', 'device_a.key')

src/harness/sas.py
```diff
@@ -32,20 +32,22 @@ def GetTestingSas():
   return SasImpl(cbsd_sas_rsa_base_url, cbsd_sas_ec_base_url, sas_sas_rsa_base_url,\
     sas_sas_ec_base_url, cbsd_sas_version, sas_sas_version, sas_admin_id), SasAdminImpl(admin_api_base_url)  
 
-def GetDefaultCbsdSSLCertPath():
-  return os.path.join('certs', 'client.cert')
```
assaneRed: replace just client.cert to device_a.cert

src/harness/sas.py
```diff
@@ -32,20 +32,22 @@ def GetTestingSas():
...
 
+def GetDefaultDomainProxySSLCertPath():
+  return os.path.join('certs', 'domain_proxy.cert')
 
-def GetDefaultCbsdSSLKeyPath():
-  return os.path.join('certs', 'client.key')
```
assaneRed: replace just client.key to device_a.key

norbertbarbosa:
Ok, I was too quick when pushing the previous commit and something goes wrong with my rebase, so I revert it. Sorry for it.
The current state is:
- I fixed the revoke_and_generate_crl.sh naming to address the first 3 comments.
- Finally I think it is better to keep the generic 'device.cert' as you seems to agree also with it: I simply updated its CN to 'Generic device' to make it obvious.
- Regarding the OID FCCID and SERIAL: I think it is a good idea to have this 2 device_a and device_c certificates locked to these 2 device_a.json and device_c.json, as it implicitly test the requirement of a certificate tied to a specific device. I think also it was your intent when you wanted to change the CN value. I checked that it will not break any other testcases (they are used in GRA_4 and SIQ_7)

Please take another look

assaneRed:
Hello
In   **def doCbsdTestCipher(self, cipher, client_cert=None, client_key=None):**
You need to change
    client_cert = client_cert or **sas.GetDefaultCbsdSSLCertPath()**
    client_key = client_key or **sas.GetDefaultCbsdSSLKeyPath()**
to
    client_cert = client_cert or **sas.GetDefaultDomainProxySSLCertPath()**
    client_key = client_key or **sas.GetDefaultDomainProxySSLKeyPath()**

For register any device, you need to use "DomainProxy(cert/key)" and not "Cbsd(cert/key)"
if you use "device.cert" the SAS should check the CN.

It works in master because client(cert/key) was generate with proxy_ca

assaneRed:
the rest looks good

norbertbarbosa:
Hi assaneRed,
I don't agree with your last comment:
- the `doCbsdTestCipher` function is used with default parameter only in SCS tests. The SCS tests are supposed to test CBSD security and the SDS tests the DomainProxy security. So the SCS tests have to use a device certificate, and not a domain proxy certificate.
- we can use a generic 'device' certificate, ie. a certificate that not include the OIDs FCCID and SERIAL. These 2 fields are the only fields that the SAS should rely on. The CN are purely informative.

Perhaps we should start a discussion in the forum to clarify these aspect?

norbertbarbosa:
Ok, Kate tend to agree with you, that there is no 'generic' device certificate.
Please hold on until we internally clarify this aspect. Thanks

assaneRed:
Hello
For my understand CN is required by FCC and SERIAL  are optional

norbertbarbosa:
Sorry for the delay.
After internal clarification, we agree with your position:  the CBSD certificate is locked to a given device and we use the CN as it is the only mandatory field.

So I updated this PR with:
- sas.py: removed the GetDefaultCbsdSSLCertPath/Key as there is no default certificate for device.
- securityTestcase: doCbsdTestCipher(), assertTlsHandshakeFailure(), assertTlsHandshakeFailureOrHttp403(): the client_cert/client_key parameters are now mandatory, as they can rely on a default CBSD certificate.
- updated the SCS/SDS testcase accordly
- removed the 'device.cert' and replace it by 'device_a.cert' everywhere

PR #555, opened 2018-04-24
Remove adjacent channel power in interference calculation.

sbdt-google:
The R2-SGN-24 specification has been clarified recnetly regarding this
point to only use fundamental emission: see R2-SGN-24  section b.iii.


Thao-Nguyen:
Serge,
Looks good. I approved this PR.

sbdt-google:
Thanks. Submitted

PR #556, opened 2018-04-25
PropModel: fix for incidence angle using hzns instead of python

sbdt-google:
+ Fix ITM horizon angle to use the hzns() C++ routine.
As specified in the spec.
It has also the added advantage of no heavy calculation loop in
pure python: speed boost of 40% for all tests doing propagation
calculations.

### DIFF
--- a/src/harness/reference_models/propagation/ehata/ehata.py
+++ b/src/harness/reference_models/propagation/ehata/ehata.py
@@ -86,7 +86,7 @@ def CbsdEffectiveHeights(height_cbsd, its_elev):
   """
 
   npts = int(its_elev[0])
-  xi = its_elev[1] / 1000.   # step size of the profile points, in km
+  xi = its_elev[1] / 1000   # step size of the profile points, in km
   dist_km = npts * xi         # path distance, in km
   elev_cbsd = its_elev[2]
 
@@ -94,10 +94,10 @@ def CbsdEffectiveHeights(height_cbsd, its_elev):
     eff_height = height_cbsd
 
   else: # dist_km >= 3 km
-    i_start = 2 + int(3.0 / xi)
+    i_start = 2 + int(math.ceil(3.0 / xi))
     i_end = npts + 2
     if dist_km > 15:
-      i_end = 2 + int(15.0 / xi)
+      i_end = 2 + int(math.floor(15.0 / xi))
       dist_km = 15.0
 
     avg_height = sum(its_elev[i_start:i_end+1]) / float(i_end - i_start + 1)
--- a/src/harness/reference_models/propagation/ehata/ehata.py
+++ b/src/harness/reference_models/propagation/ehata/ehata.py
@@ -86,7 +86,7 @@ def CbsdEffectiveHeights(height_cbsd, its_elev):
   """
 
   npts = int(its_elev[0])
-  xi = its_elev[1] / 1000   # step size of the profile points, in km
+  xi = its_elev[1] / 1000.   # step size of the profile points, in km
   dist_km = npts * xi         # path distance, in km
   elev_cbsd = its_elev[2]
 
@@ -94,10 +94,10 @@ def CbsdEffectiveHeights(height_cbsd, its_elev):
     eff_height = height_cbsd
 
   else: # dist_km >= 3 km
-    i_start = 2 + int(math.ceil(3.0 / xi))
+    i_start = 2 + int(3.0 / xi)
     i_end = npts + 2
     if dist_km > 15:
-      i_end = 2 + int(math.floor(15.0 / xi))
+      i_end = 2 + int(15.0 / xi)
       dist_km = 15.0
 
     avg_height = sum(its_elev[i_start:i_end+1]) / float(i_end - i_start + 1)
--- a/src/harness/reference_models/propagation/ehata/its/PreprocessTerrainPath.cpp
+++ b/src/harness/reference_models/propagation/ehata/its/PreprocessTerrainPath.cpp
@@ -55,7 +55,7 @@ void FindAverageGroundHeight(float *pfl, InterValues *interValues)
     {
         // ******* WinnForum extension *******
         if (_WinnForum_Extensions) {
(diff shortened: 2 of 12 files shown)


jni2000:
Serge, Mo has done an initial review on your changes and here are comments from him
1. the ehata.py's division fix is a good
2. the rest in the following files are significant changes and "may" cause uncertainty and mismatch with what we have already done. the changes to itm are more worrisome also.

- 1- itm.py
- 2- itm_its_py.cpp
- 3-PreprocessTerrainPath.cpp
- 4-itm.h
- 5-itm_test.cpp
- 6-wf_itm.py
- 7-wf_itm_test.py
- 8-ehata.py

Frankly, can you please give more details on why these changes are necessary ? Also we are not sure what the effect of all these changes might be unless we can allocate time to test them rigorously and potentially do a lot of fixes to adapt. Do you have any comparison of the calculation results before and after the changes? Are the changes only mainly related the performance improvement ?

Thanks

src/harness/reference_models/propagation/ehata/ehata.py
```diff
@@ -86,18 +86,18 @@ def CbsdEffectiveHeights(height_cbsd, its_elev):
   """
 
   npts = int(its_elev[0])
-  xi = its_elev[1] / 1000   # step size of the profile points, in km
+  xi = its_elev[1] / 1000.   # step size of the profile points, in km
```
jni2000: This is good.

src/harness/reference_models/propagation/ehata/ehata.py
```diff
@@ -86,18 +86,18 @@ def CbsdEffectiveHeights(height_cbsd, its_elev):
...
   dist_km = npts * xi         # path distance, in km
   elev_cbsd = its_elev[2]
 
   if dist_km < 3.0:
     eff_height = height_cbsd
 
   else: # dist_km >= 3 km
-    i_start = 2 + int(math.ceil(3.0 / xi))
+    i_start = 2 + int(3.0 / xi)
```
arhannan: There's a possibility to start before 3 km

src/harness/reference_models/propagation/ehata/its/PreprocessTerrainPath.cpp
```diff
@@ -55,15 +55,15 @@ void FindAverageGroundHeight(float *pfl, InterValues *interValues)
     {
         // ******* WinnForum extension *******
         if (_WinnForum_Extensions) {
-          i_start = 2 + int(ceil(3.0 / xi));
+          i_start = 2 + int(3.0 / xi);
```
arhannan: Same as before, there's that possibility to start before 3 km

src/harness/reference_models/propagation/ehata/ehata.py
sbdt-google: Yes it will slightly start before 3km, which is the original NTIA code you can find in the eHata repository:  https://github.com/NTIA/ehata
I remember now that we changed this according to your request at the time, I think.

Seriously this will not significantly change the results, but if you insist we can keep the current implementation. However in that case, please send me a piece of code which fix the code as right now it is incorrect.

src/harness/reference_models/propagation/ehata/ehata.py
sbdt-google: Most likely this has no effect as its_elev[1] tend to be float, but just in case.

sbdt-google:
In response to your comment:

The change in ITM is in fact no-op: it does not change the results at all, as expressed by the unit test. The spec says that we shall use the hzns() routine to get the vertical incidence angle.
The data output from hzns (prop.the[0] and [1]) where not output by the model, so at the time we reimplemented the equivalent hzns() function in pure python, in order to minimize the change to the ITM C code.
The change here is simply now ouputing these 2 data down the stack as output, and reusing these. Although a bunch of file are touched to pass along the data from itm.cpp to the python wrapper, down to the wf_itm.py routine, the change is trivial and gives the same result as before. See the test file which test this against the old pure-python implementation.

You have nothing to do in your SAS implementation, this change affects only the reference model to be in line exactly with the spec requirement, and gives a performance boost of 40% to all heavy tests (DPA, IAP, ..)

Regarding the other change that Ariful has reviewed on the E-Hata effective height, I can agree to go back after we fix the issue.

Thanks.
Serge

arhannan:
Serge,

This issue we believe is due to floating point operations and rounding errors in Python. We can have a function that checks for rounding errors (just as we’re doing in GetDistanceInMeters)

```
// check for rounding errors to Ndigits
double CheckRoundErrors(double numberToCheck, int ndigits)
{
                if (fabs(numberToCheck - round(numberToCheck)) < pow(10, -ndigits))
                                numberToCheck = round(numberToCheck);

                return numberToCheck;
}

```
Then we can have for example
```
        if (_WinnForum_Extensions) {
          i_end = np + 2 - int(ceil(CheckRoundErrors(3.0 / xi, 8)));
        } else {
          i_end = np + 2 - int(CheckRoundErrors(3.0 / xi, 8));
        }
```

We can keep GetDistanceInMeters or simply use  d__km = CheckRoundErrors(pfl[0] * pfl[1], 5) / 1000.

sbdt-google:
Given the ITM horizon angle is controversial, I reverted the other fix (Effective Height computation),
which will be handled by a separate PR.

Discussion on the ITM horizon angle will continue by email and this PR will be stalled meanwhilel

sbdt-google:
Thanks all for the review.

arhannan:
Hi Serge,

Some compiler have an issue with M_PI

> its/itm.cpp(883): error C2065: 'M_PI': undeclared identifier
> its/itm.cpp(884): error C2065: 'M_PI': undeclared identifier

Can we define M_PI?
```
#ifndef M_PI
	#define M_PI 3.14159265358979323846
#endif
```

Thanks

sbdt-google:
Ariful,

Please check the new PR to fix this:
https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/648

PR #558, opened 2018-04-26
fix sas version variable name

janetliuboston:
The following testcases need updated sas cbsd version variable names:

RLQ.4
REG.10
REG.13
HBT.3

self._sas._sas_version should be self._sas.cbsd_sas_version

Also, REG.10, REG.13, and RLQ.4 set the CBSD SAS version to a higher version, but do not put the version back. Added the logic to put the version numbers back to the original config version.


janetliuboston:
You're right, Kate. The sas.GetTestingSas() initializes the version from the config file before the start of each test case, so this is not needed. I'll remove it from all places so they will be uniform. Thanks!

kate-harrison:
Great, thanks! Please submit whenever you're comfortable.

PR #473, opened 2018-03-13
IAP reference model

jni2000:
Please start review the IAP and aggregate interference to non-DPA reference models implementation. 

Directory structure is as below :

•	iap: Provides APIs to compute post IAP aggregate interference margin from all the grants in the neighborhood of protection entity
•	interference : Provides utility APIs to compute interference from a single grant, finding grants in the neighborhood of protection entities and extracting grants from CBSD object. These utility APIs are used by aggregate_interference and iap reference models

PR #559, opened 2018-04-26
new IAP PR to avoid the access problem of the old PR #473

jni2000:
Looks like when review comments accumulate to a certain number, the PR will be not accessible. We have to create this new PR to replace the original PR #473. Please, reviewers, if the original PR #473 is not accessible, use this instead.


jni2000:
thanks to all reviewers, we tackled this big job!

PR #562, opened 2018-04-27
Change 'TriggerDailyActivitiesImmediately...' to fail immediately  

harinineelu:
Change 'TriggerDailyActivitiesImmediatelyAndWaitUntilComplete' to have the test case fail immediately if:

  - the status check returns HTTP 500
  - if any other error occurs 20 times in a row


harinineelu:
Thanks Javi, Merged #562

PR #566, opened 2018-04-30
Generic database class

TaranFurnell:
Adds a generic database class for use with the FDB testcases. Supports:
- https with custom CA (or http if the https flag is false)
- Basic authentication (only required for FSS database)
- Will serve content with the specified url->filepath mappings


src/harness/database.py
```diff
@@ -0,0 +1,159 @@
...
+        self.wfile.write('no auth header received.\n')
+        self.wfile.write('Ignore 404 response below.\n\n')
+        return ""
+      elif self.headers['Authorization'] != self.server.authorization:
+        logging.info('Authorization Failed. Incorrect username:password.')
+        self.send_response(401)
+        self.end_headers()
+        self.wfile.write('Incorrect username:password.\n')
+        self.wfile.write('Ignore 404 response below.\n\n')
+        return ""
```
TaranFurnell: To preempt a reasonably likely comment. Yes, this is a bit of a hacky way to send a 401 response (as it will also return the text for a 404 in the message body).

I do not think this is a problem given that:
- Authentication is only used for one database.
- The tests expect the authentication to work anyway, this code path is only for debugging purposes.
- The amount of work required to get this functioning exactly the same as a real server is significantly more than we would get out of it doing so.
- This server is only for testing.

TaranFurnell:
Great! One thing that is probably worth doing is if I change the authentication from being a string with the authentication information to just a simple boolean (and bake in the expected username and password). There's no need for the username and password to be configurable for testing purposes (it is just another way for the test to fail without a good reason), and ITS will be sandboxing each SASes testing so I don't believe it is a security issue (all anyone would get is some random test data anyway).

kate-harrison:
James, would you be OK if we merged now and fixed any issues later? I think it will be easier for everyone to integrate this into the FDB and WDB test cases if it's officially in master.

jni2000:
Ok, Taran and Kate, please go ahead and merge. Thanks.

PR #569, opened 2018-05-01
Reference model general cleanup - Part 1

sbdt-google:
This is one step of various cleanup to bring the ref models in a better coherent state.

In this step:
  - reuse the common objects for GrantInfo, etc..
  - migrate some objects in the common/data.py  (FssInfo and EscInfo)
  - misc API cleanup by removing unused parameters
  - misc cosmetic cleanup

Note: preliminary - will continue on the iap and pre_iap directory (note that this last one was left broken last week)

### DIFF
--- a/src/harness/reference_models/interference/aggregate_interference.py
+++ b/src/harness/reference_models/interference/aggregate_interference.py
@@ -35,12 +35,12 @@
 """
 
 import numpy as np
-from reference_models.interference import interference as interf
-from reference_models.geo import utils
-import multiprocessing
-from multiprocessing import Pool
 from functools import partial
 
+from reference_models.common import mpool
+from reference_models.common import data
+from reference_models.interference import interference as interf
+from reference_models.geo import utils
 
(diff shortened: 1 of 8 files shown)


src/harness/reference_models/interference/aggregate_interference.py
```diff
@@ -35,11 +35,16 @@
...
+from reference_models.common import data
 from reference_models.interference import interference as interf
 from reference_models.geo import utils
-import multiprocessing
-from multiprocessing import Pool
-from functools import partial
+
+# The grid resolution for area based protection entities.
+GWPZ_GRID_RES_ARCSEC = 10
+PPA_GRID_RES_ARCSEC = 10
```
nasrinivasa: Shouldn't this be 2 arc sec for area protection?

sbdt-google: Yes sorry, I was temporarily changing them to run regression test which do not last too long. Just fixed. this

sbdt-google:
Thanks for the review.

PR #570, opened 2018-05-01
E-Hata: Fix effective height issue

sbdt-google:
Also:
 - make the python version exactly equal to the C version. This will insure consistency and avoid people having to implement two almost identical  code in order to replicate exactly the ref model.
 - fix old unit test not working caused by fixes in begin of year

Originally part of PR #556, but has been splitted here. See #556 however for initial discussion on this issue.

### Notes: 
Effective height problem apparent at 3km, for example by running code:
```
from reference_models.propagation import wf_hybrid
from reference_models.geo import vincenty
rx_lat, rx_lon, _ = vincenty.GeodesicPoints(40, -80, 3, 0)
loss, _, internal = wf_hybrid.CalcHybridPropagationLoss(40, -80, 10,
rx_lat, rx_lon, 1.5, False, 0.5, 3625, 'SUBURBAN')
print "Loss:", loss
```
### Fix
Ariful initially recommended playing with floating point results (see #556) to get back integer values when close.
But I guess the fundamental issue of effective height is that sometimes we are in a condition >= 3km, but we do not have any points beyond 3km to use for the average height ==> produce a Nan. 
Best robustness is to always insure at least one point (the last) is used when beyond the 3km case.


sbdt-google:
I respectfully disagree.
The issue on this specific case is in fact that we use the GetDistanceInMeters() to change something like 2.999999 to 3km, which was introduced for other reasons related to non continuity of the eHata with discontinuities localized in whole distance. And which could create different results depending on the compiler (as reported by FW).
Because we get 3km although it is really 2.99999km , we go through the 3km path case, and this one recompute thing disregarding the dist_km=3, and redoing calculation of things based on the original values. This lead to the i_start bigger that i_end.
The proposed change fixes this specific case completely, and is generally robust against all cases of i_start bigger than i_end (although I dont think there are other ones possible), leading to zero division.

My thinking:
In software we always have a float precision problem, and it is not solvable in general. Playing around with changing the values can only bring you so far, but you will always hit a wall where your epsilon is not enough. Basically a robust software should not have discontinuities like we see with this version of eHata, where a change of 1e-7 on something will lead to several dB difference of path loss.
Tackling this fundamental issue is a bigger project we dont have time or resources to do right now.

Right now and given the small time frame left, we can only fix evident issue like the one reported here.
More deep fixes will have to be done as part of SAS v2 in the future.

arhannan:
When dist_km = 3km, and xi = its_elev[1] / 1000 = .03. 
3.0/xi = 100.
npts  = 100
ceil(3.0/xi) = 101 because 3.0/xi > 100. In a case we saw 1e-11 
Thus i_start  = 103
i_end = npts + 2 = 102 thus i_start > i_end

The check solves a specific problem, i_start > i_end, it doen't solve why we have that problem. There's still the issue of int. int(2.99999) will give 2.

kate-harrison:
Can we please finish and submit this PR by the end of today?

sbdt-google:
Hi Ariful,

I have done a quick pass, basically I reverted all proposed changes.
And removed the use of the GetDistanceInMeters() in that function which is not necessary. 
This fixes the problem without having to put a patch on a patch.

Please have a look.

MoGhorbanzadeh:
Deletion of GetDistanceInMeters causes hybrid model's discontinuity for a large number of points under stress testing due to the compiler issues as Serge discussed above.

sbdt-google:
We have not seen the effect of discontinuity that you mention, caused by this change. 
Can you share your stress test script, for example you can put it in a new directory reference_models/tools/stress_test ? And give information on the various compilers your are referring to ?

Looking at this specific code, you will see that only one variable has been changed (d__km), and it is only used to switch the case for effective height at the switch points 3km and 15km.

So there is only 2 instance where the code would possibly give a significant different result than before, when the distance is around exactly 3km or around exactly 15km. I don't see how this can affect a large number of points. 
Only the PPA creation has a notion of computing a path loss at an exact distance (every 200m up to 40km), and all other use of the propagation are between 2 points which can be considered random lat/lng, so it should be very rare that they fall exactly at 3km or at exactly 15km.

Please note also that the PAT tests has 0.1% failure tolerance to deal with exactly such numerical issues.

Issue #519, opened 2018-04-10
Small problems with SCS/SDS/SSS.17-19


kate-harrison:
Hi James and all,

 

Here are follow up comments I sent out before.

 

[1]Some testcases doesn’t call the SasReset() before injections/registrations, reset is needed.

 

    Comment:

    All the testcases need to reset SAS UUT to the baseline state. This is required in the test spec.

  However, many of SxS don’t send admin/reset to the SAS UUT.

      Please add self.SasReset() for each testcase.

 

        For example, some testcases call assertRegistered() or assertRegisteredAndGranted() without admin/reset. They don’t have admin/reset. Only doCbsdTestCipher have admin/reset which is used in SxS.1~5.

 

[2]The code calls generate_short_lived_certs.sh but it gives error

 

security_testcase.py line277

276    # Build short lived certificate command

277    command = "cd {0} && ./generate_short_lived_certs.sh {1} {2} {3}".format(cert_path, client_type,

278                                                                             cert_name, str(cert_duration_minutes))

279    # Create the short lived certificate

280    command_exit_status = subprocess.call(command, shell=True)

 

  (1)I think command needs to include ‘bash’

 

      Comment:

line 280:  command_exit_status = subprocess.call(command, shell=True)

         this doesn’t run.

 

         If I give ‘bash’ like below, it runs.

         command = "cd {0} && bash ./generate_short_lived_certs.sh {1} {2}…

command_exit_status = subprocess.call(command, shell=True)

 

         I suggest to add ‘bash’ like above.

 

   (2)file generate_short_lived_certs.sh includes ‘\r’.

 

      Comment:

“generate_short_lived_certs.sh” runs itself, but when called from subprocess(), it gives error.

       If I remove ‘\r’ in the file, it runs.

 

       simple_crl_server.py is the same situation when it calls revoke_and_geerate_crl.sh.

       I suggest to remove ‘\r’ from the .sh files.

 

 

Best Regards,

Takashi

jni2000:
Thanks, Kate, we will address this.

kate-harrison:
Has this issue been resolved or is there more work that needs to be done?

jni2000:
we will take care of this.

jni2000:
issue has been addressed and fixed in PR #573

PR #573, opened 2018-05-02
fix issue #519

jni2000:
This PR addresses the following issues as described in issue #519.

[1]Some   testcases doesn’t call the SasReset() before injections/registrations, reset   is needed.     Comment:     All the testcases need to reset SAS UUT to the baseline state. This is   required in the test spec.     However, many of SxS don’t send admin/reset to the SAS UUT.            Please add self.SasReset() for each   testcase.         For example, some testcases call   assertRegistered() or assertRegisteredAndGranted() without admin/reset. They   don’t have admin/reset. Only doCbsdTestCipher have admin/reset which is used   in SxS.1~5. | WINNF_FT_S_SCS_testcase.py     WINNF_FT_S_SDS_testcase.py | Done | We have added   sas.Reset() call  to the test cases   SCS/SDS 17,18 & 19 as pre condition.
-- | -- | -- | --
[2]The code calls generate_short_lived_certs.sh but it gives   error     security_testcase.py line277     276 # Build short lived certificate command     277 command = "cd {0} && ./generate_short_lived_certs.sh {1}   {2} {3}".format(cert_path, client_type,     278 cert_name, str(cert_duration_minutes))     279 # Create the short lived certificate     280 command_exit_status = subprocess.call(command, shell=True)     (1)I think command needs to include ‘bash’            Comment:     line 280: command_exit_status = subprocess.call(command, shell=True)          this doesn’t run.          If I give ‘bash’ like below, it   runs.          command = "cd {0}   && bash ./generate_short_lived_certs.sh {1} {2}…     command_exit_status = subprocess.call(command, shell=True)               I suggest to add ‘bash’ like   above. | 1.security_testcase.py     2.simple_crl_server.py | Done | Prepended   the bash accordingly to the line , where exist a call to the following   scripts     1.revoke_and_generate_crl.sh     2.generate_short_lived_certs.sh     Changes would look like below     revoke_cert_command = "cd {0} && bash ./revoke_and_generate_crl.sh " \                             "-r   {1}".format(getCertsDirectoryAbsolutePath(),                                                cert_name)
(2)file generate_short_lived_certs.sh includes ‘\r’.            Comment:     “generate_short_lived_certs.sh” runs itself, but when called from   subprocess(), it gives error.             If I remove ‘\r’ in the file, it   runs.                       simple_crl_server.py is the same   situation when it calls revoke_and_geerate_crl.sh.             I suggest to remove ‘\r’ from the   .sh files. | 1.revoke_and_generate_crl.sh     2.generate_short_lived_certs.sh | Done | Removed   '\r' from the both scripts , Basically I did dos2unix for these files.

### DIFF
--- a/src/harness/security_testcase.py
+++ b/src/harness/security_testcase.py
@@ -305,7 +305,7 @@ class SecurityTestCase(sas_testcase.SasTestCase):
     cert_path = os.path.join(harness_dir, 'certs')
 
     # Build short lived certificate command
-    command = "cd {0} && ./generate_short_lived_certs.sh {1} {2} {3}".format(cert_path, client_type,
+    command = "cd {0} && bash ./generate_short_lived_certs.sh {1} {2} {3}".format(cert_path, client_type,
                                                                              cert_name, str(cert_duration_minutes))
     # Create the short lived certificate
     command_exit_status = subprocess.call(command, shell=True)
--- a/src/harness/simple_crl_server.py
+++ b/src/harness/simple_crl_server.py
@@ -177,7 +177,7 @@ def revokeCertificate():
   # Retrives the all certificates from certs directory.
   cert_name = getCertificateNameToBlacklist()
   logging.info('Certificate selected to blacklist is:%s', cert_name)
-  revoke_cert_command = "cd {0} && ./revoke_and_generate_crl.sh " \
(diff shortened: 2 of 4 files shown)


jni2000:
Kate and Paul

Takashi has approved this fix, please take a look. Thanks.

PR #575, opened 2018-05-02
Add maximum batch size support

TaranFurnell:
- Adds configurable maximum batch size for domain proxy

### DIFF
--- a/src/harness/fake_sas.py
+++ b/src/harness/fake_sas.py
@@ -90,6 +90,7 @@ class FakeSas(sas_interface.SasInterface):
   """
 
   def __init__(self):
+    self.maximum_batch_size = 100
     pass
 
   def Registration(self, request, ssl_cert=None, ssl_key=None):
--- a/src/harness/sas.py
+++ b/src/harness/sas.py
@@ -30,9 +30,15 @@ def GetTestingSas():
   sas_sas_version = config_parser.get('SasConfig', 'SasSasVersion')
   sas_admin_id = config_parser.get('SasConfig', 'AdminId')
   maximum_batch_size = config_parser.get('SasConfig', 'MaximumBatchSize')
-  return SasImpl(cbsd_sas_rsa_base_url, cbsd_sas_ec_base_url, sas_sas_rsa_base_url,\
-      sas_sas_ec_base_url, cbsd_sas_version, sas_sas_version, sas_admin_id, maximum_batch_size),
-    SasAdminImpl(admin_api_base_url)
+  return SasImpl(
+      cbsd_sas_rsa_base_url,
+      cbsd_sas_ec_base_url,
+      sas_sas_rsa_base_url,
+      sas_sas_ec_base_url,
+      cbsd_sas_version,
+      sas_sas_version,
+      sas_admin_id,
+      maximum_batch_size), SasAdminImpl(admin_api_base_url)
 
 
 def GetDefaultDomainProxySSLCertPath():
--- a/src/harness/sas.py
+++ b/src/harness/sas.py
@@ -73,7 +73,7 @@ class SasImpl(sas_interface.SasInterface):
     self.sas_sas_version = sas_sas_version
     self._tls_config = TlsConfig()
     self._sas_admin_id = sas_admin_id
-    self.maximum_batch_size = maximum_batch_size
+    self.maximum_batch_size = int(maximum_batch_size)
 
   def Registration(self, request, ssl_cert=None, ssl_key=None):
     return self._CbsdRequest('registration', request, ssl_cert, ssl_key)
--- a/src/harness/sas.py
+++ b/src/harness/sas.py
@@ -29,8 +29,10 @@ def GetTestingSas():
   cbsd_sas_version = config_parser.get('SasConfig', 'CbsdSasVersion')
   sas_sas_version = config_parser.get('SasConfig', 'SasSasVersion')
   sas_admin_id = config_parser.get('SasConfig', 'AdminId')
+  maximum_batch_size = config_parser.get('SasConfig', 'MaximumBatchSize')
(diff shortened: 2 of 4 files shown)


src/harness/fake_sas.py
```diff
@@ -90,6 +90,7 @@ class FakeSas(sas_interface.SasInterface):
   """
 
   def __init__(self):
+    self.maximum_batch_size = 100
```
jni2000: This is maximum number of grant requests in one DP grant request message, right ? If so, can we use a more descriptive variable name to reflect this meaning ?

kate-harrison:
Javi, I didn't think of it until Taran pointed it out just now, but I think you should be able to see it in action with the IPR test case. Would you mind trying it out there so that we can get this checked in ASAP?

jcereced:
Sounds good, I'll move over to IPR and try it there.....I was getting side tracked with some of the MCP issues we are running into.  If I don't get to it tonite I'll do it tomorrow.

thanks

kate-harrison:
Thank you! Any reviewing help you can give in the next few days will be greatly appreciated!

src/harness/sas.py
```diff
@@ -29,8 +29,10 @@ def GetTestingSas():
   cbsd_sas_version = config_parser.get('SasConfig', 'CbsdSasVersion')
   sas_sas_version = config_parser.get('SasConfig', 'SasSasVersion')
   sas_admin_id = config_parser.get('SasConfig', 'AdminId')
+  maximum_batch_size = config_parser.get('SasConfig', 'MaximumBatchSize')
   return SasImpl(cbsd_sas_rsa_base_url, cbsd_sas_ec_base_url, sas_sas_rsa_base_url,\
-    sas_sas_ec_base_url, cbsd_sas_version, sas_sas_version, sas_admin_id), SasAdminImpl(admin_api_base_url)  
+      sas_sas_ec_base_url, cbsd_sas_version, sas_sas_version, sas_admin_id, maximum_batch_size),
+    SasAdminImpl(admin_api_base_url)
```
jcereced: not sure if this is really an issue or just the way it's  displaying in here.  When I copied and try to run it I got an indentation problem on line 35

jcereced:
I'm having an issue getting this to run and I've already run into it earlier when doing something else.  There was a change earlier this week in the TH and it's requiring either new certificates or some adjustment that I don't have yet so when I try to run some of the "newer" code I get a curl error.


E       CurlError: could not load PEM client certificate, OpenSSL error error:02001002:system library:fopen:No such file or directory, (no key found, wrong pass phrase, or wrong file format?)

request_handler.py:134: CurlError

do you guys know what I need to do to fix it?  I won't be able to get someone to help me until Monday..thanks

TaranFurnell:
Have you tried regenerating your certificates? The naming conventions and
code behind that has changed somewhat.

TaranFurnell:
Ok, I've fixed all but the last issue (which I can't reproduce). Incase it's a config error I've added a cast that forces the batch size to be an integer, which is the only reason I can think you would be getting the error. Can you try again with the updated code?

If you still have the error can you let me know the testcase you're using and where the `_grantRequestWithMaximumBatchSize` function that fails gets called from (there are 2 places).

jcereced:
Ok tested the changes and it's working great!!! but run into new issue:

I'm using a config that has 3 grants (I'll paste it in the next comment)

It works great with M=1, M=3,

but when M=2

util.py:76:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
testcases/WINNF_FT_S_IPR_testcase.py:181: in test_WINNF_FT_S_IPR_1
    conditional_registration_data=config['conditionalRegistrationDataN2'])
test_harness_objects.py:177: in registerCbsdsAndRequestGrants
    self.testcase.assertEqual(len(grant_responses), len(grant_requests))
E   AssertionError: 2 != 3

It seems to only send two


and when M=100 it looks like it doesn't send any:


util.py:76:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
testcases/WINNF_FT_S_IPR_testcase.py:181: in test_WINNF_FT_S_IPR_1
    conditional_registration_data=config['conditionalRegistrationDataN2'])
test_harness_objects.py:177: in registerCbsdsAndRequestGrants
    self.testcase.assertEqual(len(grant_responses), len(grant_requests))
E   AssertionError: 0 != 3




Here is the loop:

    for iteration in range(0, int(math.ceil(len(grant_requests)/int(maximum_batch_size)))):


for M=1   it is 0, 3/1    so it's   0,3
for M=3  it is 0, 3/3      so it's  0,1

I don't know what math.ceil does (I'll read on it) but i'm guessing the range is not being calculated properly if the len(gran_request)  is not divisible by M....


I'll play a bit more with it

TaranFurnell:
That loop doesn't match the PR, it's supposed to be:
`for iteration in range(0, int(math.ceil(len(grant_requests)/float(maximum_batch_size)))):`
Notice the float cast, this is needed to force floating point division, otherwise the ceiling does nothing.

jcereced:
that's what happens when you decide to vi files by hand to save time.....................doh...my bad...retrying

jcereced:
Oh yeah...looks perfect now....sorry for screwing it up.  Looks good!!!!!

PR #581, opened 2018-05-03
Set default cert/key/ca cert

TaranFurnell:


src/harness/database.py
```diff
@@ -48,6 +48,10 @@
 __PASSWORD = 'password'
 __AUTHORIZATION_STRING = 'Basic ' + base64.b64encode(USERNAME+':'+PASSWORD)
```
jni2000: Line   number 49 will prevent code from executing since USERNAME, PASSWORD  are not defined.          The Defined variables start with __USERNAME & __PASSWORD

TaranFurnell:
Fixed please review.

jni2000:
Thanks, Taran. There is another one

 def getBaseUrl():
    return self.base_url

This will prevent code from executing since def getBaseUrl(): is class function not having the parameter self. Function should be defined like : def getBaseUrl(self):

I can't add this comment inline with the codes and thus put into message and you might not notice that. Please take a look. We are very close.

PR #583, opened 2018-05-04
DPA: Cleanup indentation of move_list module.

sbdt-google:
To be in line with other test harness code. No functionality change.
Just to avoid issues in the future with messing up the indentation.


Thao-Nguyen:
I like the way it looks now, nice and clean! Thanks so much for fixing the indentations.

sbdt-google:
Thanks !

PR #586, opened 2018-05-04
assaneRed_Fix_short_lived_certificate

assaneRed:
add correct CN when generate short lived certificates

### DIFF
--- a/src/harness/security_testcase.py
+++ b/src/harness/security_testcase.py
@@ -261,7 +261,6 @@ class SecurityTestCase(sas_testcase.SasTestCase):
     try:
       self.assertTlsHandshakeFailure(client_cert, client_key, ciphers, ssl_method)
     except AssertionError as e:
-      self.SasReset()
       try:
         if is_sas:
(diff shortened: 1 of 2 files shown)


assaneRed:
Hello Kate,
Can you, please, take a look tothis PR?
It's just to fix a small bug when generate short lived certificates

Best regard,
Assane

src/harness/security_testcase.py
```diff
@@ -261,7 +261,6 @@ def assertTlsHandshakeFailureOrHttp403(self, client_cert, client_key, ciphers=No
     try:
       self.assertTlsHandshakeFailure(client_cert, client_key, ciphers, ssl_method)
     except AssertionError as e:
-      self.SasReset()
```
jni2000: why are we deleting this line ?

assaneRed: because "SasReset ()" is already called by the SSS_xx.testcase.py that use "assertTlsHandshakeFailureOrHttp403" so we do not need to reset sas again

PR #587, opened 2018-05-04
Fw mcp integration

jni2000:
initial MCP integration, please review


jni2000:
something is not right, will create a new one.

PR #590, opened 2018-05-04
QPR.5

harinineelu:


src/harness/testcases/WINNF_FT_S_QPR_testcase.py
```diff
@@ -491,3 +494,342 @@ def test_WINNF_FT_S_QPR_8(self, config_filename):
 
     # CHECK: Total EIRP of all approved grants is <= 49.15 dBm
     self.assertLessEqual(total_eirp_dbm, 49.15)
+
```
jcereced: tiniest of issues...this code is added at the end of the file so other QPR TCs with a higher number 6,7,8 will be above it.  Please move so it'll be above QPR6.  I run it and the config generated fine and CBSDs registered/granted etc....looking through the specifics of the config now

src/harness/testcases/WINNF_FT_S_QPR_testcase.py
```diff
@@ -491,3 +494,342 @@ def test_WINNF_FT_S_QPR_8(self, config_filename):
...
+        }
+    }
+    # Grant Requests - N2
+    grant1N2 = {
+        'operationParam': {
+            'maxEirp': 19,
+            'operationFrequencyRange': {
+                'lowFrequency': 3650000000,
+                'highFrequency': 3660000000
+            }
```
jcereced: This grant will not be successful since grant1N1 already asked for the same frequency so when the CBSD tries to get a grant at the same frequency it'll get a 401 from the SAS (CBSD already has that grant).  Was this on purpose?

src/harness/testcases/WINNF_FT_S_QPR_testcase.py
```diff
@@ -491,3 +494,342 @@ def test_WINNF_FT_S_QPR_8(self, config_filename):
...
+    # Check grant response 1
+    self.assertEqual(len(grant_responses1), len(grant_request_n1))
+
+    # Step 3: For CBSDs successfully registered in Step 1, Send grant request 2
+    request2 = {'grantRequest': grant_request_n2}
+    grant_responses2 = self._sas.Grant(request2)['grantResponse']
+    # Check grant response 2
+    self.assertEqual(len(grant_responses2), len(grant_request_n2))
+
+    for i, device in enumerate(successful_reg_requests):
```
nasrinivasa: Since we are not doing any distance checks, we will be checking for all CBSDs that have been registered and have  approved grants. This is ok for small number of CBSDs but for large numbers, will this be an issue?

kate-harrison: I think we expect only small numbers of CBSDs in this test case, so I'm not too worried. If you see an obvious way to make it more efficient, please let us know, but I don't expect it to be a problem.

src/harness/testcases/WINNF_FT_S_QPR_testcase.py
harinineelu: Thanks for checking, it was a mistake. Fixed it.

src/harness/testcases/WINNF_FT_S_QPR_testcase.py
harinineelu: Done :)

harinineelu:
Thanks for taking a look. 
Javi, I've fixed both. Could you please take a look. Thanks!

harinineelu:
Merged #590

PR #596, opened 2018-05-07
Border: Explicitely using median ITM pathloss

sbdt-google:
No change in fact since median is the default (and it was not specified before). 
But better to make it explicit.

This is related to change proposal: WINNF-18-I-00107
so far approved by everyone.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_BPR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_BPR_testcase.py
@@ -291,6 +291,7 @@ class BorderProtectionTestcase(sas_testcase.SasTestCase):
           closest_point_lat,
           closest_point_lon,
           closest_point_height,
+          reliability=0.5,
           cbsd_indoor=is_cbsd_indoor,
           freq_mhz=freq_mhz)
       pl = propagation.db_loss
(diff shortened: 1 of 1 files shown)


src/harness/testcases/WINNF_FT_S_BPR_testcase.py
```diff
@@ -291,6 +291,7 @@ def test_WINNF_FT_S_BPR_2(self, config_filename):
           closest_point_lat,
           closest_point_lon,
           closest_point_height,
+          reliability=0.5,
```
janetliuboston: Do these args match

propagation = wf_itm.CalcItmPropagationLoss(
          cbsd_lat,
          cbsd_lon,
          cbsd_height,
          closest_point_lat,
          closest_point_lon,
          closest_point_height,
          reliability=0.5,
          cbsd_indoor=is_cbsd_indoor,
          freq_mhz=freq_mhz)

def CalcItmPropagationLoss(lat_cbsd, lon_cbsd, height_cbsd,
                           lat_rx, lon_rx, height_rx,
                           cbsd_indoor=False,
                           reliability=0.5,
                           freq_mhz=3625.,
                           its_elev=None,
                           is_height_cbsd_amsl=False):

sbdt-google: It does not make a difference because we pass them as keyword arguments.
In fact, in Python, when an keyword argument is specified in the function prototype with a default value, it is best practice to always call the routine with keywords arguments (reliability=0.5, etc..).
Hence these arguments order can be rearranged later if you wish, without changing the client code.
So you dont have to worry about the order.

janetliuboston: Thanks :)

PR #599, opened 2018-05-07
Fix WDB test cases

harinineelu:


src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -92,13 +94,29 @@ def generate_WDB_1_default_config(self, filename):
 
     conditionals = [conditionals_b]
 
-    # Create the actual config file
+    pal_db_relative_file_path = os.path.join('testcases', 'testdata', 'pal_db',
+                                             'pal_db.json')
+
+    # Create the PAL database file.
+    pal_db_contents = pal_records_a
+    pal_db_contents.extend(pal_records_b)
```
jni2000: This will simply concatenate two PAL record into the file. Just a little concern that the final output is no longer in a valid JSON format. Making this an array of PAL records will solve this problem. However, the current solution works for the test case. Please think about this and see if we could have any future impact using non-JSON format with JSON contents.

kate-harrison: Why do you say that? I think it will write the PAL DB as a list of PAL records. Here's an experiment I did to verify (1) that `json.dumps()` works as expected and that (2) `extend` works as expected.

```
In [1]: import json

In [2]: json.dumps([{1:2}, {3:4}])
Out[2]: '[{"1": 2}, {"3": 4}]'

In [3]: a = [1]

In [4]: a.extend([2])

In [5]: a
Out[5]: [1, 2]
```

If there's an error, I'd like to correct it -- but I can't see one right now.

jni2000: Yes, you are right. I misunderstand the "pal_record_b_0". It worked. I take my comment back. Thanks.

harinineelu:
Merged #599

PR #601, opened 2018-05-08
using the standard GWPZ data format

jni2000:
Fix the IAP GWPZ format error


jni2000:
Kate and Serge, here are the fixes for IAP reference model using the standard GWPZ data format. Please review.

PR #604, opened 2018-05-08
DPA: Support for DPA absolute check interference

sbdt-google:
When SAS UUT is alone, we may want to test it against the absolute
threshold (and not the reference model result). A new option in
CheckInterference supports this. By default use the old mode.
To activate call `CheckInterference()` with
`do_abs_check_single_uut = True`

Also fix a bug on CbsdGrantInfo (maxEirp as a tuple) introduced
in previous commit.


sbdt-google:
Thanks for the quick review.

PR #605, opened 2018-05-08
Log Common errors in configurations

harinineelu:
Using common error strings


src/harness/common_strings.py
```diff
@@ -0,0 +1,18 @@
+"""Strings which are repeated throughout the test code.
+"""
+
+EXPECTED_SUCCESSFUL_REGISTRATION = (
+    'Registration was expected to succeed and failure to register almost '
+    'certainly indicates a configuration error. Please attempt registration of'
+    ' the same device using REG.11 and verify that it succeeds.'
```
janetliuboston: Would using the 'REG.11' in this message be misleading since this same message is used for different testcases?

src/harness/common_strings.py
```diff
@@ -0,0 +1,18 @@
...
+)
+
+EXPECTED_SUCCESSFUL_REGISTRATION_AND_GRANT = (
+    'The Registration and Grant requests were expected to succeed and failure'
+    ' almost certainly indicates a configuration error. Please attempt the '
+    'same requests in GRA.17 and verify that they succeed.')
```
janetliuboston: Similar comment about 'GRA.17' here.

harinineelu:
Merged #605

PR #607, opened 2018-05-08
Resolves issues in FDB.3 and database

TaranFurnell:


src/harness/database.py
```diff
@@ -73,10 +73,9 @@ def __init__(self,
     if https:
       self.server.socket = ssl.wrap_socket(
           self.server.socket,
-          certfile=__SSL_CERT,
-          keyfile=__SSL_KEY,
-          ca_certs=__SSL_CA_CERT_FILE,
-          ssl_version=ssl.PROTOCOL_TLSv1_2,
```
jni2000: why are we taking "ssl_version=ssl.PROTOCOL_TLSv1_2," out ?

TaranFurnell: TLS1.2 is only required for SASes, databases can just use the default.

jni2000: Thanks.

PR #611, opened 2018-05-08
fix run time error: list indices must be integers not str

janetliuboston:
This commit fixes run time error : “list indices must be integers, not str”

Noticed that ‘gwblRecords’ are not present in protected_entities and we are invoking pre-iap filtering with the same, so ensure 'gwblRecords' and 'fssRecords' are available.


janetliuboston:
Agree, submitted. Thanks, Kate!

kate-harrison:
Or maybe the easier fix is to add them as empty lists if they don't exist? Otherwise I think the logic is too complicated.

janetliuboston:
Sounds good.

TaranFurnell:
This at the top should work.

```python
for key in ['gwblRecords', 'fssRecords', 'ppaRecords', 'palRecords', 'gwpzRecords']:
  if key not in protected_entities:
    protected_entities[key] = []
```

janetliuboston:
Thanks, Taran. I'll use your shorter version. :)

janetliuboston:
Updated.

PR #612, opened 2018-05-08
Tiny fixes: Var names and import logging

harinineelu:


harinineelu:
Thanks Janet :) Merged #612

PR #614, opened 2018-05-08
Changes to reg and grant request - format issue.

harinineelu:
Same as #613, but for QPR.7


harinineelu:
Thanks Janet. Merged #614

PR #615, opened 2018-05-08
Zones and DPA update

sbdt-google:
- Update database with latest Exclusion zone file
- Update database with latest Coastal (ESC) and Portal DPA KML. These files will be updated when fully finalized as part of ongoing discussions
- Update the zones.py drivers to read all those data, including the special DPA parameters
- Update the BuildDpa routines used by the actual tests (IPR, MCP) to read proper DPA (either portal or coastal-ESC) and:
   + initialize all special DPA parameters
   + manage correctly channels according to spec (Always ON OOB DPA for coastal ESC only keep the 3540-3550)
   + update the example to exercise the new code (proper unit test will be added later)


src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -135,27 +137,40 @@ def __init__(self, protected_points,
                beamwidth=3,
                azimuth_range=(0, 360),
                freq_ranges_mhz=[(3550, 3650)],
-               catb_neighbor_dist=ml.CAT_B_NBRHD_DIST_DEFAULT):
+               catb_neighbor_dist=ml.CAT_B_NBRHD_DIST_DEFAULT,
```
Thao-Nguyen: Similar to the previous comment, "catb_neighbor_dist" should be set to 200 (km), just like you did in zones._LoadDpaZones().

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -135,27 +137,40 @@ def __init__(self, protected_points,
...
     """Initialize the DPA attributes."""
     self.protected_points = protected_points
     self.threshold = threshold
     self.radar_height = radar_height
     self.azimuth_range = azimuth_range
     self.beamwidth = beamwidth
-    self.channels = GetDpaProtectedChannels(freq_ranges_mhz)
     self.catb_neighbor_dist = catb_neighbor_dist
+    self.dpa_type = dpa_type
+    self.channels = None
```
Thao-Nguyen: Could you please explain why we need to set "self.channels = None" before calling ResetFreqRange()?

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -135,27 +137,40 @@ def __init__(self, protected_points,
...
     Args:
       freq_ranges_mhz: The protection frequencies (MHz) as a list of tuple
         (freq_min_mhz, freq_max_mhz) of the DPA protected frequency ranges.
     """
-    channels = GetDpaProtectedChannels(freq_ranges_mhz)
+    channels = GetDpaProtectedChannels(freq_ranges_mhz,
+                                       is_portal_dpa=(self.dpa_type=='portal'))
     if channels != self.channels:
       self.ResetLists()
     self.channels = channels
```
Thao-Nguyen: Could you please explain why "channels" is compared with "self.channels", and if True, self.ResetLists() is called? And then self.ResetLists() is called again (line 154)?

Thao-Nguyen: Also, the description of ResetLists() should be """Reset move list and neighbor list."""

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -396,7 +415,18 @@ def GetDpaProtectedChannels(freq_ranges_mhz):
...
+    # For ESC DPA, channels below 3550 are always ON, and only the highest one dominates.
+    idx_above_3550 = 0
+    for chan in channels:
+      if chan[0] >= 3550:
+        break
+      idx_above_3550 +=1
+    highest_channel_idx_below_3550 = max(0, idx_above_3550 - 1)
+    channels = channels[highest_channel_idx_below_3550:]
+
+  return channels
```
Thao-Nguyen: Clever!

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -495,26 +525,27 @@ def _DefaultProtectionPoints(name,
...
 
   Args:
-    num_pts_front_border: number of points in the front border
-    num_pts_back_border: number of points in the back border
-    num_pts_front_zone: number of points in the front zone
-    num_pts_back_zone: number of points in the back zone
-    front_us_border_buffer_km: buffering of US border for delimiting front/back.
+    dpa_geometry: The DPA geometry as a |shapely.Polygon or Point|.
+    num_pts_front_border: Number of points in the front border.
+    num_pts_back_border: Number of points in the back border.
```
Thao-Nguyen: It might be helpful to note that "num_pts_front_border" and "num_pts_back_border" >2. I actually tried 1 and got an error message "ZeroDivisionError: float division by zero" on line 553, in SampleLine(). :)

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -495,26 +525,27 @@ def _DefaultProtectionPoints(name,
...
-  dpa_zone = zones.GetDpaZones()[name]
   # Case of DPA points
-  if isinstance(dpa_zone, sgeo.Point):
-    return [ProtectionPoint(longitude=dpa_zone.x,
-                            latitude=dpa_zone.y)]
+  if isinstance(dpa_geometry, sgeo.Point):
+    return [ProtectionPoint(longitude=dpa_geometry.x, latitude=dpa_geometry.y)]
 
+  # Case of Polygon
+  us_border = zones.GetUsBorder()
```
Thao-Nguyen: Have you tried viewing "usborder.kmz" in Google Earth? Polygon 4 looks odd to me (big circle). And I can not see CONUS border.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -495,26 +525,27 @@ def _DefaultProtectionPoints(name,
...
   us_border_ext = us_border.buffer(front_us_border_buffer_km / 111.)
```
Thao-Nguyen: "front_us_border_buffer_km / 111." will convert the buffer to degrees?

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -495,26 +525,27 @@ def _DefaultProtectionPoints(name,
...
-  front_border = dpa_zone.exterior.intersection(us_border_ext)
-  back_border = dpa_zone.exterior.difference(us_border_ext)
-  front_zone = dpa_zone.intersection(us_border_ext)
-  back_zone = dpa_zone.difference(us_border_ext)
+  front_border = dpa_geometry.exterior.intersection(us_border_ext)
+  back_border = dpa_geometry.exterior.difference(us_border_ext)
+  front_zone = dpa_geometry.intersection(us_border_ext)
+  back_zone = dpa_geometry.difference(us_border_ext)
```
Thao-Nguyen: Interesting way to get protection points! Has this approach been discussed/documented somewhere?

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -562,13 +593,23 @@ def BuildDpa(dpa_name, protection_points_method=None):
     IOError: if the provided file cannot be found.
     ValueError: if the provided file is not a valid GeoJSON MultiPoint geometry.
   """
```
Thao-Nguyen: Are you going to generate the geojson files containing the protected points for each DPA?

src/harness/reference_models/dpa/dpa_mgr_example.py
```diff
@@ -136,3 +136,17 @@
...
   print 'Check Interference @%.2fdB margin: %s' % (margin_db, 'OK' if check else 'FAIL')
+
+  # Simulate the BuildDpa feature
+  print '-- Test Build DPAs --'
+  dpa_pt = dpa_mgr.BuildDpa('Pensacola')
+  print 'Pensacola:'
+
+  print dpa_pt
+  dpa_zone = dpa_mgr.BuildDpa('East3',
+                              protection_points_method='default(4,2,1,1)')
```
Thao-Nguyen: I was expecting to get 8 protection points (4 on the front border, 2 on the back border, 1 inside the front zone, and 1 inside the back zone). However, the results show only 7 protection points (see below). Did I do anything wrong?
Dpa(protected_points=[ProtectionPoint(latitude=30.2777761712794, longitude=-86.8723702411151), ProtectionPoint(latitude=28.936338148926364, longitude=-89.25247288586112), ProtectionPoint(latitude=30.13606756959702, longitude=-88.64110991511458), ProtectionPoint(latitude=29.80435257686979, longitude=-86.89639795810538), ProtectionPoint(latitude=28.776692879972796, longitude=-89.07071605705865), ProtectionPoint(latitude=30.14701158854293, longitude=-87.70039734848852), ProtectionPoint(latitude=28.450159209933283, longitude=-87.53895141517933)], ...)

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -562,13 +593,23 @@ def BuildDpa(dpa_name, protection_points_method=None):
...
+  try:
+    dpa_zone = zones.GetCoastalDpaZones()[dpa_name]
```
Thao-Nguyen: Not important. But in the "Coastal_DPAs.kml", shouldn't <Data name="portalOrg"> be <Data name="coastalOrg">?

src/harness/reference_models/geo/zones.py
```diff
@@ -181,42 +276,135 @@ def GetCoastalProtectionZone():
...
+  """Gets Coastal DPA zones.
+
+  Coastal DPA zones are Dynamic Protection Area monitored through the use of ESC sensors.
 
-  DPA zones a Dynamic Protection Area protected through the use of ESC sensors.
+  Returns:
+    A dict of DPA struct keyed by their names, each one holding following attributes:
+      geometry: A |shapely.Polygon  or Point| defining the DPA.
+      protectionCritDbmPer10MHz: The protection threshold (dBm/10MHz).
+      refHeightMeters: The radar antenna height.
```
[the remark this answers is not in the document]
sbdt-google: Done

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -83,6 +83,8 @@ class Dpa(object):
     azimuth_range: The radar azimuth range (degrees) as a tuple of
       (min_azimuth, max_azimuth) relative to true north.
     catb_neighbor_dist: The CatB neighbording distance (km).
+    dpa_type: The DPA category, either 'esc' or 'portal'.
+
     grants: The list of registered grants.
```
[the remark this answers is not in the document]
sbdt-google: I remove from here. I put underscore because the variable is now "private" (pseudo private, you can always access it in python), simply because it is unsafe to set it directly from some client code (and it may be tempting to do so).
You should instead always pass through some of the function SetGrantsFromFad() or SetGrantsFromList(). This is because when we change the grants we need to do some class state changing things (ResetLists and has_th_grants setup).

I passed the `has_th_grants`  as private, because client code does not need it most likely. It is more of an internal thing to avoid recalculating the state each time.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -135,27 +137,40 @@ def __init__(self, protected_points,
                beamwidth=3,
```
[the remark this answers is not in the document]
sbdt-google: No specific reason except that it feels that this number could change more easily than the radar height. And possibly someone may want to get this value in his code, so it is defined.
At the end, all those are defined by the KML.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: Yes, 111 would be the equatorial distance for one degree.
Have a look at the new code, where I use the US centroid of 45 degrees instead.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: Not exactly as is, it is my "poor man" approach to the thing, but it has roots in everyone approach of having points on the border, especially front border, and less point in side/back border, plus some points in the middle.

Please check the latest code on previous commit that I updated for:
 - handling correctly inland zones with no back zone/border (because fully included in the usborder)
 - handling correctly multipolygon (there is one in alaska)
 - having a minimum step size, to avoid creating too many points for small DPA like ports or inland stuff.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: Not really (as it is the role of ITS/NTIA/NIST to do so) ??
Of course from the generate points with my simple method, it is easy to create a json file with:
```
with open(filename, 'w') as fd:
   json.dump(sgeo.MultiPoint([pt for pt in dpa.protected_points]))
```

src/harness/reference_models/dpa/dpa_mgr_example.py
sbdt-google: 2 things:
 - the line sampling routine was missing the last end of the line - I fixed, so probably the reason for the missing point.
 - but in general this routine does not insure perfect number is output as requested, especially for the zone part. Look at it: we take the area then deduce the step, then grid the polygon. This is quite approximate, but sufficient for our purpose.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: I dont think so. portalOrg is only meaningful for the Portal-DPA, and not for the Coastal-DPA KML.
The fact that it is there in the Coastal is just for convenience by Nick I guess. The software will ignore it on the Coastal DPA KML.
From Andy Clegg proposed document describing the format:

> For P-DPAs, the name of an informing incumbent portal organization to which the DPA belongs. Multiple organizations can be listed. This establishes the authorization for portal users under the listed organization(s) to be allowed to “book” conference rooms (channels) belonging to this building (DPA). If no organizations are listed, only portal administrators will be able to activate any channels in this DPA.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: The code in zones is temporary while we waited for having the actual distance provided in the KML file, or some other mean. I received this morning a sheet, so I will include that for now, and will update when finally decided what is the final source.

Note that there a still a few details to be figured out (do we still use HAAT and 300km ,etc..) - we'll know better when the DPA WG1 requirement is finalized with latest development.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: No real need. It was just to have all the actual attributes clearly showed in the __init__, so a reader can see exactly what are the attributes. It is bad design to spread attributes (especially public ones) all over the place. As a reader of the code I want to check in the __init__ and see exactly what is the state variable of my class.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: We need to reset if the state of the DPA change, ie grants are changed or channel are changed.
I agree that all this is not fully hermetic and could be improved. If I change the radar height, the list will become invalid as well, but currently we dont manage that, as we do not expect that someone would do such thing.

The channel list change may happen however the way the tests are written, I am not fully sure anymore. And the grants too.
Perhaps I should revisit all this to be a pure calculation class without state keeping - just returning the lists as standalone objects. But I guess it is a bit late now to do massive architectural change.

sbdt-google: Done for the description.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: thanks!

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: Ok I will put a check to handle this case.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: Yes it is somewhat a bit broken for Earth due to a known issue on the generation script, which creates something slightly invalid in the geometry. It is ok however in shapely. 

I have a low priority item to fix the script and file.

sbdt-google:
Hi Thao,

I have processed your comments and made a few other improvements for inland DPA and multi-geometry DPAs.
I still have a small change to do for integration of the neighboring distance provided this morning.
Will ping you when that last set of changes are done. Meanwhile let me know if there are other things
you see or would like to have.

Serge

Thao-Nguyen:
Hi Serge,
Sounds good! Thanks so much!
Thao

sbdt-google:
Hi Thao,

I have made a last set of changes to:
 - integrate the temp list for neighbor distance.
 - improve the management of properties to extract:
    * separation of portal and coastal properties
    * definition of converter and default value in single place there
    * various improvements

Can you have a quick look, and let me know what is missing for getting your approval ?
Thanks,
Serge

data/ntia/Coastal_DPAs_CatBNeighborhoods.csv
```diff
@@ -0,0 +1,41 @@
...
+West10,Monterey,192
+West9,SanFrancisco,144
+West8,Manchester,208
+West7,DelgadaCanyon,160
+West6,Eureka,160
+West5,PortOrford,176
+West4,Eugene,176
+West3,Portland,160
+West2,Aberdeen,192
+West1,LaPush,176
```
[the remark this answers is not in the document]
sbdt-google: In fact this is one of the open question which needs to be solved at the spec level, when updating the requirement to handle the variable catB distance.
I personally think that the new way of setting distance should render the high HAAT condition useless, but I guess this will be decided in WG1.
Meanwhile I have decided to keep the old behavior, and we will do a special PR to update this point. Do you agree?

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -135,27 +136,42 @@ def __init__(self, protected_points,
...
+            'catb_neighbor_dist=%.1f, monitor_type=%r)' % (
+                self.protected_points, self.threshold, self.radar_height,
+                self.beamwidth, self.azimuth_range, self.channels,
+                self.catb_neighbor_dist, self.monitor_type))
+
   def ResetFreqRange(self, freq_ranges_mhz):
     """Reset the frequency ranges of the DPA.
 
+    If the range have changed, the move and keep lists are also reset.
+
```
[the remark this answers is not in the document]
sbdt-google: Changed the description in the 3 places.
Yes we keep the neighbor list and recompute the keep list as needed from these 2 list.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -481,55 +513,100 @@ def _CalcTestPointInterfDiff(point,
...
-       for pt in SampleLine(back_border, num_pts_back_border)] +
-      [ProtectionPoint(longitude=pt[0], latitude=pt[1])
-       for pt in utils.GridPolygon(front_zone, step_front_dpa_arcsec)] +
-      [ProtectionPoint(longitude=pt[0], latitude=pt[1])
-       for pt in utils.GridPolygon(back_zone, step_back_dpa_arcsec)])
+    min_step = min_step_arcsec / 3600.
+    if step < min_step:
+      num_points = round(line.length / min_step) + 1
+      step = line.length / (num_points-1)
+    return [line.interpolate(dist) for dist in np.arange(0, line.length+step/2., step)]
```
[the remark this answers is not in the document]
sbdt-google: Yes you are right. It will create a duplicate if one of the line loop around (as is the case for . It will also create some duplicate if we have the back line not empty (like for coastal DPA), as when we sample the back line we will restart from the last point of the front line.

So I am changing back the condition to not include the last point of a line, since it will be captured in both cases anyway. And changing back the condition:
```
  step=line.length / (num_points-1)
into:
  step = line.length / num_points
```
Hence we will have the correct number of requested points.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -481,55 +513,100 @@ def _CalcTestPointInterfDiff(point,
...
+
+  def CvtKmToArcSec(dist_km, ref_geo):
+    return dist_km / (111 * np.cos(ref_geo.centroid.y * np.pi / 180.)) * 3600.
```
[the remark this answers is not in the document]
sbdt-google: I update the formula to be more explicit. 
I guess 111 is the easy number to remember for the approcimate distance in equator of 1 degree of latitude/longitude).

src/harness/reference_models/dpa/dpa_mgr_example.py
```diff
@@ -136,3 +136,28 @@
...
+  # Simulate the BuildDpa feature for various types
+  print '\n-- Test Build DPAs --'
+  dpa_pt = dpa_mgr.BuildDpa('Pensacola')
+  print '**Pensacola: %dpts' % len(dpa_pt.protected_points)
+  print dpa_pt
+
+  dpa_zone = dpa_mgr.BuildDpa('East3',
+                              protection_points_method='default(4,2,1,1)')
+  print '**East3: %dpts' % len(dpa_zone.protected_points)
+  print dpa_zone
```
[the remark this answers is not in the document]
sbdt-google: I corrected a few things (see previous comment) + support for 0 point.
We still get 9 points on east3, but without the duplicate point.
The 9th point caused by the approximate front zone grid sampling.

src/harness/reference_models/dpa/dpa_mgr_example.py
```diff
@@ -136,3 +136,28 @@
...
+
+  dpa_portal = dpa_mgr.BuildDpa('BATH')
+  print '**BATH: %dpts' % len(dpa_portal.protected_points)
+  print dpa_portal
+
+  dpa_inland = dpa_mgr.BuildDpa('Alameda',
+                                protection_points_method='default(1000,2000,1,1,40,0.5,2)')
+  print '**Alameda: %dpts' % len(dpa_inland.protected_points)
+  print dpa_inland
```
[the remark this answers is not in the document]
sbdt-google: Should be fixed as well with previous fix.

src/harness/reference_models/dpa/dpa_mgr_example.py
```diff
@@ -136,3 +136,28 @@
...
+
+  dpa_multigeo = dpa_mgr.BuildDpa('Alaska9',
+                                protection_points_method='default(10,5,2,1,40,0.5,2)')
+  print '**Alaska9: %dpts' % len(dpa_multigeo.protected_points)
+  print dpa_multigeo
```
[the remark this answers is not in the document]
sbdt-google: I havent looked at them specifically. Can it be that the part2 is cut by the extended border in a bit of strange way, so this explains the oddness.
Also please note again that the separation is in the lat/lon space, and it does not corresponds to the actual distance (in km) when going  in the horizontal vs vertical.

Again this method is just a simple method to have something setup by default on defaut testset, or simulations. The actual protected points to use should best be derived by an offline procedure and saved into a geojson as  MultiPoint (especially for coastal zones).

src/harness/reference_models/geo/zones_test.py
```diff
@@ -33,15 +33,37 @@ def test_read_protection(self):
...
+    z = zones.GetCoastalDpaZones()
+    for name, zone in z.items():
+      self.assertTrue(zone.geometry.is_valid)
+    self.assertAlmostEqual(z['East5'].geometry.area, 6, 1)
+    self.assertEqual(z['East5'].freqRangeMhz, [(3500, 3650)])
+    self.assertEqual(z['East5'].protectionCritDbmPer10MHz, -144)
+    self.assertEqual(z['East5'].refHeightMeters, 50)
+    self.assertEqual(z['East5'].minAzimuthDeg, 0)
+    self.assertEqual(z['East5'].maxAzimuthDeg, 360)
+    self.assertEqual(z['East5'].catbNeighborDist, 200)
```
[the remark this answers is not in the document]
sbdt-google: My mistake. Done

Thao-Nguyen:
Thanks, Serge! Will do.

Thao-Nguyen:
Hi Serge,
There are no major issues on my side, so I approve this PR as you suggested. 
But please consider/fix these minor things before you submit:
- Many DPAs, e.g., Alaska2, Alaska9, etc., in the latest e_DPAs.kml now have the same name "Part 1". I'm not sure why, but it caused "dpa_mgr_example.py" to fail (ValueError: DPA Alaska9 not found in DPA database). 
- In dpa_mgr_example.py, line 730: change "freqRangeMhz" to "freqRangeMHz".
- In zones_test.py, line 50: change "freqRangeMhz" to "freqRangeMHz" and change [(3500, 3650)] to [(3550, 3650)].

sbdt-google:
Thanks. Submitting now

Thao-Nguyen:
Great! Thank you!

PR #617, opened 2018-05-09
fix GWPZ and PPA data records in testdata

jni2000:
Fix test data issues for MCP
ppa_record_x.json: added missing PPA info element.
gwpz_record_0.json: added "landCategory"

### DIFF
--- a/src/harness/reference_models/common/data.py
+++ b/src/harness/reference_models/common/data.py
@@ -265,13 +265,16 @@ def getAuthorizedGrantsFromDomainProxies(domain_proxies):
   Returns:
     A list of |CbsdGrantInfo| for each authorized grant in the given Domain Proxies.
   """
-  grants = []
+  grant_records = []
+  cbsd_records = []
   for domain_proxy in domain_proxies:
     for cbsd in domain_proxy.getCbsdsWithAtLeastOneAuthorizedGrant():
+      reg_record = cbsd.getRegistrationRequest()
       for grant in cbsd.getAuthorizedGrants():
-        grants.append(
-            constructCbsdGrantInfo(
-                cbsd.getRegistrationRequest(),
-                grant.getGrantRequest()))
-
-  return grants
+        grant_record = grant.getGrantRequest()
+        grant_records.append(grant_record)
+      cbsd_records.append({
+                   'registration':reg_record,
+                   'grants': grant_records
+                   })
+  return cbsd_records
--- a/src/harness/reference_models/common/data.py
+++ b/src/harness/reference_models/common/data.py
@@ -265,16 +265,13 @@ def getAuthorizedGrantsFromDomainProxies(domain_proxies):
   Returns:
     A list of |CbsdGrantInfo| for each authorized grant in the given Domain Proxies.
   """
-  grant_records = []
-  cbsd_records = []
+  grants = []
   for domain_proxy in domain_proxies:
     for cbsd in domain_proxy.getCbsdsWithAtLeastOneAuthorizedGrant():
-      reg_record = cbsd.getRegistrationRequest()
       for grant in cbsd.getAuthorizedGrants():
-        grant_record = grant.getGrantRequest()
-        grant_records.append(grant_record)
-      cbsd_records.append({
-                   'registration':reg_record,
-                   'grants': grant_records
-                   })
-  return cbsd_records
+        grants.append(
+            constructCbsdGrantInfo(
+                cbsd.getRegistrationRequest(),
+                grant.getGrantRequest()))
+
+  return grants
--- a/src/harness/testcases/WINNF_FT_S_MCP_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_MCP_testcase.py
@@ -53,7 +53,7 @@ class McpXprCommonTestcase(sas_testcase.SasTestCase):
     self.test_type = test_type
     self.sas_test_harness_objects = []
     self.domain_proxy_objects = []
-    self.protected_entity_records = {}
+    self.protected_entity_records = []
     self.num_peer_sases = len(config['sasTestHarnessConfigs'])
 
     for domain_proxy in config['domainProxyConfigs']:
@@ -116,12 +116,12 @@ class McpXprCommonTestcase(sas_testcase.SasTestCase):
       iteration_content: A dictionary with multiple key-value pairs that contain iteration data
     """
     # Step 5 : Inject IAP protected entities into UUT
-    for key in iteration_content['protectedEntities']:
+    for key in self.iteration_content['protectedEntities']:
       if not key in self.protected_entity_records:
         self.protected_entity_records[key] = iteration_content['protectedEntities'][key]
       else:
         self.protected_entity_records[key].extend(iteration_content['protectedEntities'][key])
-
+    self.protected_entity_records.extend(iteration_content['protectedEntities'])
     if 'fssRecords' in self.protected_entity_records:
       for fss_record in self.protected_entity_records['fssRecords']:
         try:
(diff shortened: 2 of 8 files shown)


src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -617,6 +616,11 @@ def generate_MCP_1_default_config(self, filename):
                                                                  pal_low_frequency,
                                                                  pal_high_frequency,
                                                                  'test_user_1')
+
+    # Add a CBSD to be part of the PPA
+    cbsd_reference_id1 = generateCbsdReferenceId('test_fcc_id_x', 'test_serial_number_x')
```
kate-harrison: What is this for?

src/harness/reference_models/common/data.py
```diff
@@ -265,13 +265,16 @@ def getAuthorizedGrantsFromDomainProxies(domain_proxies):
...
     for cbsd in domain_proxy.getCbsdsWithAtLeastOneAuthorizedGrant():
+      reg_record = cbsd.getRegistrationRequest()
       for grant in cbsd.getAuthorizedGrants():
-        grants.append(
-            constructCbsdGrantInfo(
-                cbsd.getRegistrationRequest(),
-                grant.getGrantRequest()))
-
-  return grants
+        grant_record = grant.getGrantRequest()
```
kate-harrison: Will this work correctly? I assume that you're trying to construct the `GrantData` object from SAS-SAS, and it has a slightly different format than the `GrantRequest` object. Where is the output of this function used?

It also looks like the output type of this function has changed; will the code that uses this function still work?

src/harness/reference_models/common/data.py
```diff
@@ -265,13 +265,16 @@ def getAuthorizedGrantsFromDomainProxies(domain_proxies):
...
+        grant_records.append(grant_record)
+      cbsd_records.append({
+                   'registration':reg_record,
+                   'grants': grant_records
+                   })
+  return cbsd_records
```
TaranFurnell: This isn't the fix we want as it will break other usage. I think the issue is a bit deeper than it initially looks. I'll have a look into it a bit more thoroughly and discuss what the expected behavior is supposed to be with Kate. (Should have a PR up by the end of the day, hopefully sooner). In the meantime I'd be happy to approve the `WINNF_FT_S_MCP_testcase.py` changes as a separate PR (Minus the section Kate commented on).

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
jni2000: The ppa record in testdata folder does not have ‘cbsdReferenceId’ field, here we are creating the field and adding dummy reference id to it. 
This code is added for testing IAP and Aggregate models. If we are going to add the field in input ppa record, we can ignore the change.
Please let us know your thoughts.

src/harness/reference_models/common/data.py
jni2000: Taran has modified Aggregate interference code in PR621, so this changes are not required now, Will remove.

src/harness/reference_models/common/data.py
jni2000: Thanks, restored this back to "master" copy and wait for your fix.

jni2000:
Taran and Kate

We noticed that you have addressed some of the fixes in PR #621. We restored ours back. Here we have the following remaining changes, please review
1. fixes in "src/harness/reference_models/pre_iap_filtering/pre_iap_util.py", please check if these are still valid.
2. added the missing "PPA info" fields into "testdata/ppa_record_x.json"

Changes in "src/harness/reference_models/common/data.py" and "src/harness/testcases/WINNF_FT_S_MCP_testcase.py" are taken out and we will wait for your changes in PR #621.

Please take a look. Thanks.

TaranFurnell:
I left a comment on PR #621 but repeating it here because I noticed you approved it just before I pushed the change. I've added the preIAP change to PR #621, that will make this PR purely about fixing the test data (which Kate knows far more about than me).

jni2000:
All python code fixes have been covered in PR #621 and thus all are reverted back to master copies. This PR will only fix test data issues for MCP
1. ppa_record_x.json: added missing PPA info element.
2. gwpz_record_0.json: added "landCategory"

jni2000:
Kate, I think Taran has incorporated the correct portions of our fixes into PR #612 and has already merged in. In this PR, we will only focus on fixing the test data by adding the missing fields. Please can you help review this ? Thanks.

src/harness/testcases/testdata/ppa_record_0.json
```diff
@@ -4,6 +4,10 @@
   "usage": "PPA",
   "terminated": false,
   "ppaInfo": {
+    "palId": ["test_fcc_id_pal_1"],
```
kate-harrison: These four fields are filled in the make...Consistent() function, so there's no need to add them here. We should add the region to ppa_record_3.json but otherwise these changes can be reverted.

kate-harrison: Slight modification: I agree that adding an *empty list* for the CBSD reference IDs makes sense. Thank you!

jni2000: do we want to make it empty or with a single fake id? I think it make sense to have at least on cbad id as otherwise the PPA is not valid.

kate-harrison: A fake ID would confuse the SAS. Empty is better, and is allowed in our definition of the testing API.

jni2000: changed cbsdReferenceId to empty list

jni2000:
Thanks, for the fast review :), Can you please also approve on behalf of Taran ?

kate-harrison:
James, I just realized that actually we shouldn't have changed `gwpz_record_0.json` since that's what's injected into the SAS UUT. I think you meant to change `reference_models/interference/test_data/gwpz_0.json` (and similar files). If you agree, can you please create a small PR to fix this? Thanks!

jni2000:
Kate,

First of all, I checked all the gwpz record data in /reference_models/interference/test_data, they are all up to date, but just inserted the “landCategory” before “zone”. So we are fine with them.

I checked the change we made in this gwpz record and compared it with the gwpz_ut.json in /reference_models/interference/test_data (I assume this is a file to be injected into UUT as well) and they are similar, both of them have “landCategory” file, but in different places (before or after “zone”), so I do not think we need to revert the change in src/harness/testdata/gwpz_record_0.json. Also, even if we have that “landCategory“ filed that we do not have before, SAS UUT shall able to consume it.

Let me know if you think otherwise.

James Ni
Director of SAS System
978.996.7480

kate-harrison:
(1) Sorry, you're right, the files in `reference_models/interference/test_data/` appear to be up to date. I was looking at an older version which had not yet been updated. No change is needed here.

(2) If you look at the usage for gwpz_record_0.json: (a) it is used in testcases like SIQ.2, in which it is injected directly into the SAS; and (b) it is used to generate the GPR configs. In (a), we do NOT want to include the land category field since that is not a valid input to the SAS UUT. In (b), we specifically add the land category to the field right after it's injected to the SAS, so we don't want to include the land category (in gwpz_record_0.json and in the default MCP/GPR config) here either. Basically the MCP code will add the land category to gwpz_record_0.json at the right time, so we don't need to include it in the file itself.

jni2000:
ok, will do a fix for 2 as you suggested.

kate-harrison:
Thank you!

jni2000:
fix is in PR #644, please review.

PR #618, opened 2018-05-09
Fix unit tests & benchmark on latest code state

sbdt-google:
Unit testing:
- Make the unit tests work from top level script :  ./run_all_tests.sh
- Moving around the supporting test-only routines
- Fix issue with previous IAP moving of the 'inter-sas-duplicate-grant' on some unit tests import
- misc minor cleanup

Multiprocessing & geo caches:
 - improve the mpool library
 - provide a configuration mechanism for top level scripts related to multiprocess and geo cache.


sbdt-google:
Thanks!

PR #624, opened 2018-05-10
fix injectDatabaseUrl errors

jni2000:
fix injectDatabaseUrl errors found CommScope.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_WDB_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_WDB_testcase.py
@@ -305,7 +305,8 @@ class WinnforumDatabaseUpdateTestcase(sas_testcase.SasTestCase):
     cpi_database.setFilesToServe(files)
 
     # Inject the CPI database URL into the SAS UUT.
-    self._sas_admin.InjectDatabaseUrl({'type': 'CPI', 'url': config['cpiDatabaseConfig']['indexUrl']})
+    self._sas_admin.InjectDatabaseUrl({'type': 'CPI', 'url': cpi_database.getBaseUrl()+
+                                       config['cpiDatabaseConfig']['indexUrl']})
 
     # Step 3: Trigger daily activities.
     self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete()
--- a/src/harness/testcases/WINNF_FT_S_WDB_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_WDB_testcase.py
@@ -231,7 +231,8 @@ class WinnforumDatabaseUpdateTestcase(sas_testcase.SasTestCase):
             'status': 'ACTIVE',
             'publicKeyIdentifierFile': cpi_device_d_public_key_relative_file_path
         }],
-        'indexUrl': '/rest/cpi/v1/index.csv')
+        'indexUrl': '/rest/cpi/v1/index.csv',
+        'indexPath': os.path.join('testcases', 'testdata', 'cpi_db', 'index.csv')
     }
 
     # Create the actual configuration.
@@ -285,8 +286,8 @@ class WinnforumDatabaseUpdateTestcase(sas_testcase.SasTestCase):
 
     # Create the main CPI database file.
     cpis_keys = ['cpiId', 'status', 'publicKeyIdentifier']
-    ensureFileDirectoryExists(config['cpiDatabaseConfig']['indexUrl'])
-    with open(config['cpiDatabaseConfig']['indexUrl'], 'w+') as output_file:
+    ensureFileDirectoryExists(config['cpiDatabaseConfig']['indexPath'])
+    with open(config['cpiDatabaseConfig']['indexPath'], 'w+') as output_file:
       csv_writer = csv.writer(output_file, delimiter=',')
       for cpi in config['cpiDatabaseConfig']['cpis']:
         pki_url = cpi_database.getBaseUrl(
--- a/src/harness/testcases/WINNF_FT_S_WDB_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_WDB_testcase.py
@@ -160,8 +160,8 @@ class WinnforumDatabaseUpdateTestcase(sas_testcase.SasTestCase):
                                 config['palDatabaseConfig']['filePath'])
 
     # Inject the PAL database URL into the SAS UUT.
-    self._sas_admin.InjectDatabaseUrl(pal_database.getBaseUrl()+
-                                      config['palDatabaseConfig']['fileUrl'])
+    self._sas_admin.InjectDatabaseUrl({'type': 'PAL', 'url': pal_database.getBaseUrl()+
+                                       config['palDatabaseConfig']['fileUrl']})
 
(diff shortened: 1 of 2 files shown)


src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -305,7 +305,7 @@ def test_WINNF_FT_S_WDB_2(self, config_filename):
     cpi_database.setFilesToServe(files)
 
     # Inject the CPI database URL into the SAS UUT.
-    self._sas_admin.InjectDatabaseUrl(config['cpiDatabaseConfig']['indexUrl'])
+    self._sas_admin.InjectDatabaseUrl({'type': 'CPI', 'url': config['cpiDatabaseConfig']['indexUrl']})
```
kate-harrison: I think we need one more fix here so that we are sure to inject the full URL, not just the relative path, and then you will have addressed all of Javi's bugs :)

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -160,8 +160,8 @@ def test_WINNF_FT_S_WDB_1(self, config_filename):
                                 config['palDatabaseConfig']['filePath'])
 
     # Inject the PAL database URL into the SAS UUT.
-    self._sas_admin.InjectDatabaseUrl(pal_database.getBaseUrl()+
-                                      config['palDatabaseConfig']['fileUrl'])
+    self._sas_admin.InjectDatabaseUrl({'type': 'PAL', 'url': pal_database.getBaseUrl()+
+                                       config['palDatabaseConfig']['fileUrl']})
```
kate-harrison: Looks like we need to append the file name to this URL as well (thanks Javi!).

jni2000: The fileUrl already included the file name in it.

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
jni2000: fixed.

jni2000:
Kate and Javi

Fixed the CPI database full path issue. The PAL database url path already included the file name. Please can you do a double check before merge ?

Thanks

kate-harrison:
James, I think you just need to change the 'filePath' in the PAL DB config to point to a file rather than a folder. So something like "/pal_db.json" would work.

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -161,7 +161,7 @@ def test_WINNF_FT_S_WDB_1(self, config_filename):
 
     # Inject the PAL database URL into the SAS UUT.
     self._sas_admin.InjectDatabaseUrl({'type': 'PAL', 'url': pal_database.getBaseUrl()+
-                                       config['palDatabaseConfig']['fileUrl']})
+                                       config['palDatabaseConfig']['filePath']})
```
kate-harrison: I think you need to change the value of 'fileUrl', not this line. I understand 'filePath' to be the local path to the file, not the URL that the DB server will be expecting. So please revert this change and update 'fileUrl' to be 'pal_db.json'.

jni2000: yes, you are right.

jni2000:
Please take a look of both WDB.1 and .2 injected URL now, they shall be like

{ "type": "PAL", "url": "https://10.250.12.169:8003/rest/pal/v1/pal_db.json"}
{ "type": "CPI", "url": "https://10.250.12.169:8003/rest/cpi/v1/index.csv"}

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
jni2000: I was wrong, now fixed.

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -231,8 +231,7 @@ def generate_WDB_2_default_config(self, filename):
             'status': 'ACTIVE',
             'publicKeyIdentifierFile': cpi_device_d_public_key_relative_file_path
         }],
-        'indexUrl':
-            os.path.join('testcases', 'testdata', 'cpi_db', 'index.csv')
+        'indexUrl': '/rest/cpi/v1/index.csv')
```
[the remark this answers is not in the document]
jni2000: great catch. Please take a look of the fix.

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -285,8 +286,8 @@ def test_WINNF_FT_S_WDB_2(self, config_filename):
 
     # Create the main CPI database file.
     cpis_keys = ['cpiId', 'status', 'publicKeyIdentifier']
-    ensureFileDirectoryExists(config['cpiDatabaseConfig']['indexUrl'])
-    with open(config['cpiDatabaseConfig']['indexUrl'], 'w+') as output_file:
+    ensureFileDirectoryExists(config['cpiDatabaseConfig']['indexPath'])
+    with open(config['cpiDatabaseConfig']['indexPath'], 'w+') as output_file:
```
kate-harrison: You need to change line 301 as well, then this looks great.

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -298,8 +298,8 @@ def test_WINNF_FT_S_WDB_2(self, config_filename):
     files = {('/' + cpi['publicKeyIdentifierFile']):
              cpi['publicKeyIdentifierFile']
              for cpi in config['cpiDatabaseConfig']['cpis']}
-    files['/' + config['cpiDatabaseConfig']['indexUrl']] = config[
-        'cpiDatabaseConfig']['indexUrl']
+    files['/' + config['cpiDatabaseConfig']['indexPath']] = config[
```
kate-harrison: The first should be 'indexUrl' and the second should be 'indexPath', as described in database.py.

jni2000: yes, should be good now. please take a look.

src/harness/testcases/WINNF_FT_S_WDB_testcase.py
```diff
@@ -298,7 +298,7 @@ def test_WINNF_FT_S_WDB_2(self, config_filename):
     files = {('/' + cpi['publicKeyIdentifierFile']):
              cpi['publicKeyIdentifierFile']
              for cpi in config['cpiDatabaseConfig']['cpis']}
-    files['/' + config['cpiDatabaseConfig']['indexPath']] = config[
+    files['/' + config['cpiDatabaseConfig']['indexUrl']] = config[
```
kate-harrison: Sorry I didn't catch it earlier, but the leading "/" should be removed from this line since it is already present in the index URL.

jni2000: yes, you are right. thanks.

PR #627, opened 2018-05-10
fix the missing 'initialFad' in TH config of GPR. EPR, FPR and PPR

jni2000:
This PR is for fixing the issue found by Takashi as described below.

Kate, per you analysis, I agree with your fix suggestions except that I do not think we need to remove the "if" on line 86 of MCP but just add the required 'initialFad' to the GPR.2 test_harness config file. Please take a look of the fix and let me know if think otherwise.

> [Kate] Ahh,
 good point! I guess we should remove the "if" on line 86 of MCP (since line 87 is actually required for successful execution if there are any SAS THs) and set the 'initialFad' to be the same as the 'iterationData' FAD in the GPR_2 config (simulating a SAS TH which had CBSDs already loaded at the beginning of the test and no change the next day).
 

> [Takashi] For GPR_2, there are two sas_test_harness_[0,1]_config defined. And there is no 'initialFad' set in the config in line 301~317.
In that case, in step3 of MCP, no initialFad is written. (Fad Records are written IF there is initialFad, however in GPR_2 case, there is no initialFad.)
84         # Initialize content of test harness.
85         if 'initialFad' in test_harness:
86           sas_test_harness_object.writeFadRecords(test_harness['initialFad'])
Next time, writeFadRecords will be done in Step13. Until then, there is no FAD file. There is no empty FAD, either.
The test harness tries to get FAD file in Step 6,7 (before Step 13).
174       # Pull FAD from SAS Test Harnesses and fad objects are created for each SAS Test Harnesses
175       self.test_harness_fads = []
176       for test_harness in self.sas_test_harness_objects:
177         self.test_harness_fads.append(getFullActivityDumpSasTestHarness(
178                                           test_harness.getSasTestHarnessInterface()))
Then it gives error here because of no FAD files.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_EPR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_EPR_testcase.py
@@ -322,6 +322,14 @@ class EscProtectionTestcase(McpXprCommonTestcase):
                                           [[grant_request_2], [grant_request_3]]
     )
 
+    # Generate SAS Test Harnesses dump records
+    dump_records_sas_test_harness_0 = {
+        'cbsdRecords': cbsd_fad_records_sas_test_harness_0
+    }
+    dump_records_sas_test_harness_1 = {
+        'cbsdRecords': cbsd_fad_records_sas_test_harness_1
+    }
+
     # SAS Test Harnesses configuration
     sas_test_harness_0_config = {
         'sasTestHarnessName': 'SAS-TH-1',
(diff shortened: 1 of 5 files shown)


jni2000:
Yes, that is what I think too. I will work on other xPR fixes all together and let you review before merge.

jni2000:
Additional similar fixes done for EPR, FPR and PPR test cases. Looks like no need to do for BRP, IPR and QPR test cases. Please review.

jni2000:
Do not know why github did not pull the newest fixes from the branch to this PR. Created a new one #631 to replace this one. All changes are now in #631. This PR is closed and the comment history shall be kept for record.

PR #630, opened 2018-05-11
Update openssl config of SAS cert

RedaHossino:
Update openssl config of SAS cert to use only one config with both server auth and client auth

## [jni2000] Note: Add the following context from Takashi's comment email to explain the background of this modification

> Hi, Reda, and all
Thanks Reda, I have studied some more. 

>BTW
 I don’t know if there is a way to generate a certificate with both(server and client auth), I tried that before and I couldn’t make it work on windows :/
Yes I’m also thinking the same idea(below Option-2 solution). 
We haven’t try this, but if it can be done, that’s more easier to implement. 
I think the current MCP code can be used with small/no changes. 
We may need to change the openssl.conf (and generate_fake_certs.sh)

Below is thought in our team what is the problem and possible approach. 

 (SAS-TH) ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::: (SAS-UUT)
  Client ............................................................................ Server
(ClientCert_ut)    ------ RequestFAD_ut -------->(ServerCert_ut)
for client auth    <----- ResponseFAD_ut ------- for server auth
  Server ........................................................................... Client
(ServerCert_th)  <----- RequestFAD_th -------- (ClientCert_th)
for server auth  ------- ResponseFAD_th-------> for client auth

Observation-1
certs/server.cert is generated for the purpose of serverAuth. We can confirm it by doing below. 

$ openssl x509 -text -in certs/server.cert
 
             X509v3 Key Usage: critical
                 Digital Signature, CRL Sign
             X509v3 Extended Key Usage:
                 TLS Web Server Authentication

This comes from settings in openssl.conf  (& certs/generate_fake_certs.sh, -regexts) 
   in certs/generate_fake_certs.sh, “sas_req” is set in “–regexts” field. 
   in openssl.conf,       extendedKeyUsage = serverAuth  is set in [ sas_req ] 

Observation-2
  SAS-TH uses the certs/server.cert certificates as ClientCert_ut during  Request FAD_ut.
  i.e. Current code : ServerCert_th = ClientCert_ut = certs/server.cert

From these two observations, 
SAS-TH is using certs/server.cert whose extendedKeyUsage=serverAuth for the purpose of clientAuth. 

Problem 
According to the TLS rule, extendedkeyUsage field is mandatory. We should not violate the usage written in the certificate file. 
SAS-TH is violating the TLS rule of the usage of the certificate. 
--I found some of the server implementation can pass this, however, it is not permitted to use it in this way as a manner. 

Possible solution for SAS-TH to use the correct ClientCert_ut file during  Request FAD_ut. 
Option-1  
   SAS-TH will use the certs/sas.cert certificates as ClientCert_ut. 
Option-2  
   Add clientAuth in the extendedKeyUsage field when generating certs/server.cert. 
   It is permitted to have both serverAuth and clientAuth in the extendedKeyUsage field in the certificate certs/servrer.cert.  
Then SAS-TH can use the same file as ClientCert_ut and ServerCert_th. 

generate_fake_certs.sh 
  98 # Generate fake server certificate/key.$
  99 echo -e "\n\nGenerate 'server' certificate/key"$
100 openssl req -new -newkey rsa:2048 -nodes \$
101 -reqexts sas_req -config ../../../cert/openssl.cnf \$
102 -out server.csr -keyout server.key \$
103 -subj "//C=US/O=Wireless Innovation Forum/OU=WInnForum SAS Provider Certificate/CN=localhost"$
104 openssl ca -cert sas_ca.cert -keyfile private/sas_ca.key -in server.csr \$
105 -out server.cert -outdir ./root \$
106 -policy policy_anything -extensions sas_req_sign -config ../../../cert/openssl.cnf \$
107 -batch -notext -create_serial -utf8 -days 1185 -md sha384$
 

openssl.conf
[ sas_req ] (openssl.conf  line317)  (server.cert generation is reffering here) 
subjectKeyIdentifier = hash
basicConstraints = CA:FALSE
keyUsage = critical, digitalSignature, cRLSign
extendedKeyUsage = serverAuth 
certificatePolicies=@cps,ROLE_SAS,FRN
crlDistributionPoints=@crl_section

[ sas_client_mode_req ] (openssl.conf  line345)  (FYI, reference for sas.cert generation)
subjectKeyIdentifier = hash
basicConstraints = CA:FALSE
keyUsage = critical, digitalSignature, keyEncipherment, cRLSign
extendedKeyUsage = clientAuth
certificatePolicies=@cps,ROLE_SAS,FRN
crlDistributionPoints=@crl_section

Best Regards,
Takashi

### DIFF
--- a/cert/openssl.cnf
+++ b/cert/openssl.cnf
@@ -319,7 +319,7 @@ policyIdentifier = 2.16.840.1.114412.2.1
 subjectKeyIdentifier = hash
 basicConstraints = CA:FALSE
 keyUsage = critical, digitalSignature, cRLSign
-extendedKeyUsage = serverAuth
+extendedKeyUsage = serverAuth,clientAuth
 certificatePolicies=@cps,ROLE_SAS,FRN
 crlDistributionPoints=@crl_section
 [ cps ]
@@ -333,7 +333,7 @@ subjectKeyIdentifier = hash
 authorityKeyIdentifier = keyid:always,issuer
 basicConstraints = CA:FALSE
 keyUsage = critical, digitalSignature, cRLSign
-extendedKeyUsage = serverAuth
+extendedKeyUsage = serverAuth,clientAuth
 certificatePolicies=@cps,ROLE_SAS,FRN
 crlDistributionPoints=@crl_section
 [ cps ]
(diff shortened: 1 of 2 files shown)


jni2000:
Paul and Nobert, I have no comments on the modifications. Please can you also help to verify the changes, especially to make sure that the modification did not violate our original intention or limit the usages. Thanks.

Issue #632, opened 2018-05-11
Error in calling getRandomLatLongInPolygon(...) in GPR test cases


jni2000:
Takashi reported this issue as described below

https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/93adfd017e70bdeddf8408a76001b4855231df33/src/harness/testcases/WINNF_FT_S_GPR_testcase.py

[1]  getRandomLatLongInPolygon()  in line 192  in generate_GPR_2_default_config 

   Current: (e.g. Line 192) 
    device_3['installationParam']['longitude'] = getRandomLatLongInPolygon(gwpz_record_1['zone'])

I think the type mismatches. 

Expected: I’m not sure but [‘zone’] is not needed here?
device_3['installationParam']['longitude'] = getRandomLatLongInPolygon(gwpz_record_1)

kate-harrison:
I'm closing this, as I believe this issue is now fixed.

PR #633, opened 2018-05-11
fix the getRandomLatLongInPolygon call input parameter issue

jni2000:
fix the getRandomLatLongInPolygon call input parameter issue. The getRandomLatLongInPolygon shall take a higher level of object above "zone" element.


jni2000:
Kate and Takashi

Please review the fix to the issue #632 reported by Takashi.

Thanks

PR #634, opened 2018-05-11
Add extensive checking of the MCP and xPR configs.

harinineelu:
... and fix the default configs where necessary.


kate-harrison:
We executed the code, which is actually how we found the various configuration errors :) The conditionals which are removed should all be specified in other parts of the config, so we are just removing a redundant definition which was not used in the MCP code anyway. Thanks!

PR #642, opened 2018-05-14
Fix mpool to support mono-cpus.

sbdt-google:
The code was not supporting mono-cpu, andthis would lead to an error.


sbdt-google:
Thanks, submitting.

Issue #639, opened 2018-05-14
SIQ.3 checks on C2 and C4 are not correct


jni2000:
In SIQ.3, 

1. CBSD C2 is out of the range of every protected entity and thus the queried frequency range of 3550-3700 shall be all available. The current checks failed when respond with 3550-3700.
2. CBSD C4 is only within the 150km range of the FSS and thus the frequency range of 3550-3650 shall be all available and returned in response. The current checks for C4 failed when responded with 3550-3650.

kate-harrison:
I agree, this appears to be a bug in the code. The [check](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/WINNF_FT_S_SIQ_testcase.py#L241) uses the function `assertChannelsContainFrequencyRange()`, which based on [this test](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/helper_unit_testcase.py#L54) will fail the assertion if the list of channels is *larger* than the frequency range that's supposed to be covered. I suggest we create a copy of this function which removes that constraint and use it for C2 and C4 in SIQ.3.

jni2000:
Ok, will fix this. Thanks.

kate-harrison:
This is fixed in #649; closing.

PR #649, opened 2018-05-15
fix SIQ.3 C2 and C4 checking

jni2000:
This is fox fixing issue #639. Basically 
1. for C2, since CBSD C2 is out of every single protected entities and thus the full frequency range from 3550-3700 shall be available to be responded in spectrum inquiry response.
2. for C4, CBSD C4 is only in range of the FSS site with a GWBL, so we only need to exclude the GWBL-FSS frequency range and thus frequency range 3550-3650 shall be available to be responded in spectrum inquiry response.

The original codes will fail the tests when SAS respond single frequency range of 3550-3700 and 3550-3650 respectively.

Please review.

### DIFF
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -298,10 +298,10 @@ class SasTestCase(unittest.TestCase):
         reverse=False)
     for index, channel in enumerate(channels):
       if index == 0:
-        self.assertLessEqual(channel['frequencyRange']['lowFrequency'],
+        self.assertGreaterEqual(channel['frequencyRange']['lowFrequency'],
                          frequency_range['lowFrequency'])
       else:
-        self.assertLessEqual(
+        self.assertGreaterEqual(
             channel['frequencyRange']['lowFrequency'],
             channels[index - 1]['frequencyRange']['highFrequency'])
 
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -282,14 +282,21 @@ class SasTestCase(unittest.TestCase):
     self.assertEqual(channels[0]['frequencyRange']['highFrequency'],
                      frequency_range['highFrequency'])
 
-  def assertChannelsWithinFrequencyRange(self, channels, frequency_range):
-    """Checks if the frequency range of all channels are within 
-       a given frequency range.
+  def assertChannelsOverlapFrequencyRange(self, channels, frequency_range,
+                                          constrain_low=False, constrain_high=False):
+    """Checks if the frequency range of all channels overlap
+       the given frequency range.
 
     Args:
       channels: A list of dictionaries containing frequencyRange,
         which is a dictionary containing lowFrequency and highFrequency.
       frequency_range: A dictionary containing lowFrequency and highFrequency.
+      constrain_low: boolean indicating if the low edge frequency of the channel
+        at the lowest frequency end need to be constrained on the low frequency
+        of the frequency range.
+      constrain_high: boolean indicating if the high edge frequency of the channel
+        at the highest frequency end need to be constrained on the high frequency
+        of the frequency range.
     """
     channels.sort(
         key=
@@ -298,8 +305,12 @@ class SasTestCase(unittest.TestCase):
         reverse=False)
     for index, channel in enumerate(channels):
       if index == 0:
-        self.assertGreaterEqual(channel['frequencyRange']['lowFrequency'],
-                         frequency_range['lowFrequency'])
+        if constrain_low:
+          self.assertEqual(channel['frequencyRange']['lowFrequency'],
+                           frequency_range['lowFrequency'])
+        else:
+          self.assertGreaterEqual(channel['frequencyRange']['lowFrequency'],
+                                  frequency_range['lowFrequency'])
       else:
         self.assertGreaterEqual(
             channel['frequencyRange']['lowFrequency'],
@@ -307,8 +318,12 @@ class SasTestCase(unittest.TestCase):
 
     channels.sort(
         key=lambda ch: (ch['frequencyRange']['highFrequency']), reverse=True)
-    self.assertLessEqual(channels[0]['frequencyRange']['highFrequency'],
-                     frequency_range['highFrequency'])
+    if constrain_high:
+      self.assertEqual(channels[0]['frequencyRange']['highFrequency'],
+                       frequency_range['highFrequency'])
+    else:
+      self.assertLessEqual(channels[0]['frequencyRange']['highFrequency'],
+                           frequency_range['highFrequency'])
 
   def assertChannelIncludedInFrequencyRanges(self, channel, frequency_ranges):
     """Checks if the channel lies within the list of frequency ranges.
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -282,6 +282,34 @@ class SasTestCase(unittest.TestCase):
     self.assertEqual(channels[0]['frequencyRange']['highFrequency'],
                      frequency_range['highFrequency'])
 
+  def assertChannelsWithinFrequencyRange(self, channels, frequency_range):
+    """Checks if the frequency range of all channels is within 
+       a given input frequency range.
+
+    Args:
+      channels: A list of dictionaries containing frequencyRange,
+        which is a dictionary containing lowFrequency and highFrequency.
+      frequency_range: A dictionary containing lowFrequency and highFrequency.
+    """
+    channels.sort(
+        key=
+        lambda ch: (ch['frequencyRange']['lowFrequency'],
+                    ch['frequencyRange']['highFrequency']),
+        reverse=False)
+    for index, channel in enumerate(channels):
+      if index == 0:
+        self.assertLessEqual(channel['frequencyRange']['lowFrequency'],
+                         frequency_range['lowFrequency'])
+      else:
+        self.assertLessEqual(
+            channel['frequencyRange']['lowFrequency'],
+            channels[index - 1]['frequencyRange']['highFrequency'])
+
+    channels.sort(
+        key=lambda ch: (ch['frequencyRange']['highFrequency']), reverse=True)
+    self.assertLessEqual(channels[0]['frequencyRange']['highFrequency'],
+                     frequency_range['highFrequency'])
+
   def assertChannelIncludedInFrequencyRanges(self, channel, frequency_ranges):
     """Checks if the channel lies within the list of frequency ranges.
 
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -309,10 +309,10 @@ class SasTestCase(unittest.TestCase):
           self.assertEqual(channel['frequencyRange']['lowFrequency'],
                            frequency_range['lowFrequency'])
         else:
-          self.assertGreaterEqual(channel['frequencyRange']['lowFrequency'],
-                                  frequency_range['lowFrequency'])
+          self.assertLessEqual(channel['frequencyRange']['lowFrequency'],
+                               frequency_range['lowFrequency'])
       else:
-        self.assertGreaterEqual(
+        self.assertLessEqual(
             channel['frequencyRange']['lowFrequency'],
             channels[index - 1]['frequencyRange']['highFrequency'])
 
@@ -322,8 +322,8 @@ class SasTestCase(unittest.TestCase):
       self.assertEqual(channels[0]['frequencyRange']['highFrequency'],
                        frequency_range['highFrequency'])
     else:
-      self.assertLessEqual(channels[0]['frequencyRange']['highFrequency'],
-                           frequency_range['highFrequency'])
+      self.assertGreaterEqual(channels[0]['frequencyRange']['highFrequency'],
+                              frequency_range['highFrequency'])
 
   def assertChannelIncludedInFrequencyRanges(self, channel, frequency_ranges):
     """Checks if the channel lies within the list of frequency ranges.
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -304,6 +304,8 @@ class SasTestCase(unittest.TestCase):
                     ch['frequencyRange']['highFrequency']),
         reverse=False)
     for index, channel in enumerate(channels):
+      self.assertGreater(channel['frequencyRange']['highFrequency'],
+                         frequency_range['lowFrequency'])
       if index == 0:
         if constrain_low:
           self.assertEqual(channel['frequencyRange']['lowFrequency'],
@@ -316,6 +318,9 @@ class SasTestCase(unittest.TestCase):
             channel['frequencyRange']['lowFrequency'],
             channels[index - 1]['frequencyRange']['highFrequency'])
 
+    for index, channel in enumerate(channels):
+      self.assertLess(channel['frequencyRange']['lowFrequency'],
+                      frequency_range['highFrequency'])
     channels.sort(
         key=lambda ch: (ch['frequencyRange']['highFrequency']), reverse=True)
     if constrain_high:
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -283,8 +283,8 @@ class SasTestCase(unittest.TestCase):
                      frequency_range['highFrequency'])
 
   def assertChannelsWithinFrequencyRange(self, channels, frequency_range):
-    """Checks if the frequency range of all channels is within 
-       a given input frequency range.
+    """Checks if the frequency range of all channels are within 
+       a given frequency range.
 
     Args:
       channels: A list of dictionaries containing frequencyRange,
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -291,12 +291,12 @@ class SasTestCase(unittest.TestCase):
       channels: A list of dictionaries containing frequencyRange,
         which is a dictionary containing lowFrequency and highFrequency.
       frequency_range: A dictionary containing lowFrequency and highFrequency.
-      constrain_low: boolean indicating if the low edge frequency of the channel
-        at the lowest frequency end need to be constrained on the low frequency
-        of the frequency range.
-      constrain_high: boolean indicating if the high edge frequency of the channel
-        at the highest frequency end need to be constrained on the high frequency
-        of the frequency range.
+      constrain_low: A Boolean flag indicating if the lower frequency edge of the
+        channel with the lowest frequency segment need to be constrained to be 
+        equal to the lower frequency edge of the frequency range.
+      constrain_high: A Boolean flag indicating if the upper frequency edge of the
+        channel with the highest frequency segment need to be constrained to be 
+        equal to the upper frequency edge of the frequency range.
     """
     channels.sort(
         key=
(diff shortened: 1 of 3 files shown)


src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -92,9 +92,14 @@ def test_WINNF_FT_S_SIQ_3(self):
 
     # Inject GWBL station operating at range 3650 - 3700 MHz within 150 km of
     # FSS (approx. 100 km from FSS location set above).
-    gwbl = json.load(
+    gwbl_data = json.load(
         open(os.path.join('testcases', 'testdata', 'gwbl_record_0.json')))
-    self._sas_admin.InjectWisp(gwbl)
+    self._sas_admin.InjectWisp(gwbl_data)
+    gwbl = gwbl_data['record']
+    gwbl_low_frequency = gwbl['deploymentParam'][0]['operationParam'][
```
kate-harrison: The GWBL frequency is irrelevant since we protect the top 50 MHz (3650-3700) regardless of the GWBL frequency. Please revert this and the related changes.

kate-harrison: From the SAS Admins agreement: "If the CBSD is within 150 km of an FSS station operating below 3700MHz, and a GWBL station is also within 150 km of the FSS station (using any portion of the channel), SAS shall declare the 3650-3700 MHz as unavailable. (This applies to the Grant procedure as well.)"

jni2000:
The fix here is the simplest one without needs of now functions. What would be the function you suggested in issue #639 look like ?

jcereced:
hey adding myself as reviewer.  We were talking about this and something odd on the check for C1 (first CBSD).  It says freq should be up to 3650 for GAA, why not 3700?  there are no incumbents near it.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -238,8 +246,7 @@ def test_WINNF_FT_S_SIQ_3(self):
             for channel in channels_pal[0]))
 
     # Checks for cbsd 2 (index 1)
-    self.assertChannelsContainFrequencyRange(channels_low[1], freq_range_low)
-    self.assertChannelsContainFrequencyRange(channels_high[1], freq_range_high)
+    self.assertChannelsContainFrequencyRange(all_channels[1], freq_range_all)
```
kate-harrison: This would require the SAS to include the PAL channel (F) as available for GAA use, which is not required by the test spec.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -262,8 +269,7 @@ def test_WINNF_FT_S_SIQ_3(self):
     self.assertEqual(len(channels_excluded[2]), 0)
 
     # Checks for cbsd 4 (index 3)
-    self.assertChannelsContainFrequencyRange(channels_low[3], freq_range_low)
-    self.assertChannelsContainFrequencyRange(channels_high[3], freq_range_high)
+    self.assertChannelsContainFrequencyRange(all_channels[3], freq_range_all_but_fss_gwbl)
```
kate-harrison: Same problem here.

kate-harrison:
Javi: it should be OK if the SAS says 3650-3700 is available for C1, it's just not mandatory that it do so. I guess the fix for C1 would be similar to the fixes for C2, C4.

jcereced:
Thanks Kate.  I'll check the fix with it's ready to see how it deals with C1 and if anything else is needed.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Yes, the issue is that if SAS's response included F as available for GAA, which is correct, the test code shall pass instead of fail.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Same answer.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: This will allow SAS to include F but not mandated as any subsection of frequency within the allowed range will pass.

jni2000:
For C1, we just need to add an additional "self.assertChannelsContainFrequencyRange(channels[1], [3650-3700]), it shall fix the problem.

jni2000:
Kate and Javi

Can you please take a look of my response and see if we have any more questions/issues on this ? We can also address the C1 issue if a separate PR is not needed.

James

kate-harrison:
James, I agree that the SAS should be allowed to include F as available for GAA. My problem with your proposed fix is that it *requires* the SAS to do this, which is not something that the test spec requires.

jni2000:
Kate, I understand your concern. However, I believe "self.assertChannelsContainFrequencyRange(....)" function will pass if "all_channel[3]" are within the allowed frequency range. And this "all_channel[3]" is extracted from the SAS response. If SAS included F, F will be in the "all_channel[3]" array and will be checked. If F is not included, it will not be part of "all_channel[3]" array and thus will not be checked. So the logic will not fail if SAS did not include F in response. Correct me if I am wrong.

kate-harrison:
As far as I can tell, the call `self.assertChannelsContainFrequencyRange(all_channels[1], freq_range_all)` on line 249 requires that the SAS's response include all channels 3550-3700.

As [this unit test](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/helper_unit_testcase.py#L72) shows, the function will require that `all_channels[1]` includes all frequencies which are in `freq_range_all` and `freq_range_all` is 3550-3700, so this line requires that the SAS response include all channels.

The problem is that you can't include F in the second arg, otherwise it will require SAS UUT to include F in its response (which is not valid according to the test spec). But, as you pointed out in your initial report of the problem, if F is included in the first arg (SAS response), the function will also fail -- but it shouldn't. Hence my suggestion that we create a copy of this function which is less strict and just requires that the first arg cover *at least* all of the freqs in the second arg (the current function requires that the first arg cover *exactly* the freqs in the second arg).

jni2000:
Fixed now as suggested. Please review.

jni2000:
please can we move forward on reviewing this PR ?

arhannan:
Hi James,

we are stilling seeing some issues.

consider this example

```
      channels = [{'frequencyRange' : {'lowFrequency': 3550, 'highFrequency': 3620}},
                   {'frequencyRange' : {'lowFrequency': 3620, 'highFrequency': 3650}}]
       
      frequency_range = { 'lowFrequency': 3550, 'highFrequency': 3620}     
      self.assertChannelsContainFrequencyRange(channels, frequency_range)
```
This fails because of `self.assertEqual(channels[0]['frequencyRange']['highFrequency'],
                     frequency_range['highFrequency'])`

> AssertionError: 3550 != 3620

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -238,8 +246,7 @@ def test_WINNF_FT_S_SIQ_3(self):
             for channel in channels_pal[0]))
 
     # Checks for cbsd 2 (index 1)
-    self.assertChannelsContainFrequencyRange(channels_low[1], freq_range_low)
-    self.assertChannelsContainFrequencyRange(channels_high[1], freq_range_high)
+    self.assertChannelsWithinFrequencyRange(all_channels[1], freq_range_all)
```
kate-harrison: This does not match the test spec, since a SAS which responds with zero available channels  for C2 would pass, although the test spec requires that they give *at least* [3550,F] and [F,3650] as available.

jni2000: Yes, agreed. This is also what I was thinking while running. So what I am thnking
1. for satisfying the "at least" requirement, we shall make sure the required frequency ranges are returned
2. on top of 1, we shall allow SAS to return more than the "at least" requirement as long as the returned frequency ranges are valid.

Do you agree ?

The challenge is then how to handle the complicity in the test code logic to satisfy the above especially for 2 when there are so many combinations. My current thinking is to check if any one of the following is true:
1. SAS returned [3550,F] and [F,3650] 
2. SAS returned [3550,F+f(low)1] and [F(high)-f2,3650], where F(low)+f1 <= F(high)-f2
3. SAS returned [3550,F+f(low)1] and [F(high)-f2, f3], where f3 < 3700 for C2 or 3650 for C4
4. SAS returned [3550, 3700-f4] for C2 and [3550-3650] for C4, where 3700-f4 >= 3650

There are a lot other combinations we may have to skip. let me know your opinion. Also for C1 and C3 shall we also allow extending beyond "at least" requirements ?

kate-harrison:
I agree with your first points 1 and 2 (by "valid" I assume you mean "consistent with the protected entity information").

I don't think the code has to be as complicated as you're thinking, and I really wouldn't want to make assumptions about what the SAS UUT will do. Why isn't the fix as simple as changing the implementation of `assertChannelsContainFrequencyRange()` to be more permissive, as suggested in #639? It seems to be a simple fix which gives maximum flexibility to the SAS UUT. Is there a logical error in this proposed solution? Or is the implementation much more complex than it seems?

jni2000:
The problem we are seeing is, for example, if SAS responses only a single frequency range of [3550-3700] for C2 and [3550-3650] for C4, it is valid how will a more permissive assertChannelsContainFrequencyRange() allow this ?

kate-harrison:
Let's look at C2 specifically for a moment. If your SAS gives "3550-3700" as a single available channel, then `channels_low[1]` on [line 242](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/WINNF_FT_S_SIQ_testcase.py#L242) will contain the single "3550-3700" channel. `freq_range_low` is equal to 3550-3620. Right now, this line errors because 3550-3700 is larger than 3550-3620. I'm suggesting to make the function NOT give an error when the first arg completely contains (and exceeds) the second arg. Shouldn't this fix your problem?

jni2000:
I see the direction of your suggestion. However, channel_low[] generated by 

```
channels_low.append(filterChannelsByFrequencyRange(channels,
                                                         freq_range_low))
```
will not contain "3550-3700" when freq_range_low is equal to 3550-3620.

kate-harrison:
The change I suggested in #665 would address this problem. You can add it to this PR if you want to test.

jni2000:
We will need to change the low_frequency as below to the one follows
```
    freq_range_low = {
        'lowFrequency': 3550000000,
        'highFrequency': pal_low_frequency,
   }
```
```
    freq_range_low = {
        'lowFrequency': 3550000000,
        'highFrequency': 3700000000,
   }
```
And this will affect the rest of the logic.

kate-harrison:
Why do you think we would need to make that change?

jni2000:
ok, found it, I was still having the old version of "filterChannelsByFrequencyRange". The new version even with the current ">=" and "<=" shall work. Thanks, I'll rebase and make the fix,

jni2000:
Put in a fix, a little completed but works. Please review.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -239,12 +243,15 @@ def test_WINNF_FT_S_SIQ_3(self):
             for channel in channels_pal[0]))
 
     # Checks for cbsd 2 (index 1)
-    self.assertChannelsContainFrequencyRange(channels_low[1], freq_range_low)
-    self.assertChannelsContainFrequencyRange(channels_high[1], freq_range_high)
+    self.assertChannelsOverlapFrequencyRange(channels_low[1], freq_range_low, True, False)
+    self.assertChannelsOverlapFrequencyRange(channels_high[1], freq_range_high, False, False)
+    for channel in channels[1]:
```
kate-harrison: I think you mean to use `all_channels[1]` instead.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -239,12 +243,15 @@ def test_WINNF_FT_S_SIQ_3(self):
...
+      self.assertLessEqual(channel['frequencyRange']['highFrequency'],
+                           freq_range_all['highFrequency'])
```
kate-harrison: Just use 3700000000 directly since `freq_range_all` is otherwise unused.

src/harness/sas_testcase.py
```diff
@@ -282,6 +282,49 @@ def assertChannelsContainFrequencyRange(self, channels, frequency_range):
     self.assertEqual(channels[0]['frequencyRange']['highFrequency'],
                      frequency_range['highFrequency'])
 
+  def assertChannelsOverlapFrequencyRange(self, channels, frequency_range,
```
kate-harrison: Please add some unit tests, similar to what was done for the other function, so that we can be sure this works as intended? I think there are a number of bugs below and they would be quickly caught if we had some unit tests.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -208,6 +208,10 @@ def test_WINNF_FT_S_SIQ_3(self):
         'lowFrequency': 3650000000,
```
kate-harrison: The changes in this file all seem logically correct, assuming that the new function is correctly implemented.

src/harness/sas_testcase.py
```diff
@@ -282,6 +282,49 @@ def assertChannelsContainFrequencyRange(self, channels, frequency_range):
...
+        lambda ch: (ch['frequencyRange']['lowFrequency'],
+                    ch['frequencyRange']['highFrequency']),
+        reverse=False)
+    for index, channel in enumerate(channels):
+      if index == 0:
+        if constrain_low:
+          self.assertEqual(channel['frequencyRange']['lowFrequency'],
+                           frequency_range['lowFrequency'])
+        else:
+          self.assertGreaterEqual(channel['frequencyRange']['lowFrequency'],
```
kate-harrison: I think this is backwards. You want to make sure that the channel's lower freq is less than or equal to the freq range's lower freq (to ensure that it's completely covered). If you had freq range = 3600-3650 and channel = 3550-3650, you would fail here because 3550 < 3600.

src/harness/sas_testcase.py
```diff
@@ -282,6 +282,49 @@ def assertChannelsContainFrequencyRange(self, channels, frequency_range):
...
+                                  frequency_range['lowFrequency'])
+      else:
+        self.assertGreaterEqual(
```
kate-harrison: Why are you reversing the direction of the comparison? Is there a bug in `assertChannelsContainFrequencyRange()`?

jni2000: Thanks for pointing out, it was an leftover mistake. Fixed.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -239,12 +243,15 @@ def test_WINNF_FT_S_SIQ_3(self):
             for channel in channels_pal[0]))
 
     # Checks for cbsd 2 (index 1)
-    self.assertChannelsContainFrequencyRange(channels_low[1], freq_range_low)
-    self.assertChannelsContainFrequencyRange(channels_high[1], freq_range_high)
+    self.assertChannelsOverlapFrequencyRange(channels_low[1], freq_range_low, True, False)
```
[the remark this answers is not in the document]
jni2000: fixed as per suggested for all the places.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: yes, thanks.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: changed

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -239,12 +243,15 @@ def test_WINNF_FT_S_SIQ_3(self):
             for channel in channels_pal[0]))
```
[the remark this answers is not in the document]
jni2000: Done as per suggested.

src/harness/sas_testcase.py
jni2000: will incorporate unit test in.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: Thanks.

jni2000:
Kate, thanks for the review and the catches. Fixed all of them, please review. Will incorporate unit test for the new helper function.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -231,7 +231,12 @@ def test_WINNF_FT_S_SIQ_3(self):
 
     # Checks for cbsd 1 (index 0)
     self.assertChannelsContainFrequencyRange(channels_low[0], freq_range_low)
-    self.assertChannelsContainFrequencyRange(channels_high[0], freq_range_high)
+    constrain_low = True
+    constrain_high = False
+    self.assertChannelsOverlapFrequencyRange(channels_high[0], freq_range_high, constrain_low, constrain_high)
```
kate-harrison: You can just do:

```
self.assertChannelsOverlapFrequencyRange(channels_high[0], freq_range_high, constrain_low=True, constrain_high=False)
```

This is how you use keyword args in Python, and it makes the code more compact and readable.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -208,6 +208,10 @@ def test_WINNF_FT_S_SIQ_3(self):
         'lowFrequency': 3650000000,
         'highFrequency': 3700000000,
     }
+    freq_range_all = {
```
kate-harrison: Please remove this since it's no longer needed.

src/harness/sas_testcase.py
kate-harrison: Thanks! To save some time and effort, I'll wait to see the unit tests before reviewing this function further.

To save *yourself* some time and effort, I think you can mostly use the original unit tests and edit as needed.

Then to de-duplicate the code a bit, we can use
```
assertChannelsOverlapFrequencyRange(..., constrain_low=True, constrain_high=True)
```
instead of
```
assertChannelsContainFrequencyRange(...)
```
in SIQ_3 and remove assertChannelsContainFrequencyRange().

kate-harrison:
James, when do you think you can have an update to this code ready for review?

jni2000:
Unit test cases are added.

src/harness/sas_testcase.py
jni2000: Test cases added and passed all.

jni2000:
Kate, we have added and passed all required test cases. Please can you review ?

arhannan:
Changes look good and SIQ3 is passing now. I will await Kate's review.

PR #650, opened 2018-05-16
Update DPA Ids to ensure it is near CBSDs

sushantparkhi:
Updated DPA Ids of GRA.1, HBT.12, IPR.1, and IPR.2 using KML File in #615

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_IPR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_IPR_testcase.py
@@ -115,7 +115,7 @@ class FederalIncumbentProtectionTestcase(sas_testcase.SasTestCase):
     }
 
     dpa_1 = {
-        'dpaId': 'East4',
+        'dpaId': 'Pascagoula',
         'frequencyRange': {
             'lowFrequency': 3620000000,
             'highFrequency': 3630000000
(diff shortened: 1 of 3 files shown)


src/harness/testcases/WINNF_FT_S_IPR_testcase.py
```diff
@@ -296,7 +296,7 @@ def generate_IPR_2_default_config(self, filename):
     }
 
     dpa_1 = {
-        'dpaId': 'East4',
+        'dpaId': 'East3',
```
kate-harrison: Please double-check that these DPAs obey the constraints of the test case (I think dpa_2 should be the same dpaId as dpa_1 but a different channel, and dpa_3 needs to be an adjacent DPA).

sushantparkhi: Yes, I double checked that dpa_1 is East3, dpa_2 is East3 and dpa_3 is East21(It is weird that East21 is adjacent to East3 in the file!)

kate-harrison: Great! I've asked one of my colleagues to double-check the finer details since, as you can tell, I was doing only a cursory inspection :)

sbdt-google: The east dpa ids have been shuffled when separation of the at-sea DPA with their port one has been done for the 3 dpa in the east. This explains the strange numbering we have now.

sushantparkhi:
Thanks!

Issue #651, opened 2018-05-16
head-room margin per SAS is not accounted for in the calculation in calculatePostIapAggregateInterference


jni2000:
> Navin
with regards to IAP reference model, when computing the allowed reference aggregate in calculatePostIapAggregateInterference(), the head-room margin per SAS is not accounted for in the calculation.

> Kate: I agree, it seems that the headroom is not being added back in. The current equation in line 643 adds the leftover and the aggregate interference caused by CBSDs managed by SAS UUT, but it omits the difference between `q_p` and the original threshold (i.e. the headroom). I guess we need to add another arg to "calculatePostIapAggregateInterference()" to denote the original threshold, and then this value can be used on line 643. If you agree, can you please submit the fix?

>jni2000
We have thoroughly reviewed the codes and agreed that we need to put in a fix.

kate-harrison:
Closing as a duplicate of #637.

PR #653, opened 2018-05-16
fix issue 651 - IAP calculation headroom margin issue

jni2000:
We have reviewed the IAP calculation and agreed that there is an issue as Navin pointed out. We found the root cause of the issue was related to the input parameter to calculatePostIapAggregateInterference(). We concluded that there is no need to account for  HM_p_s as per the below equation. The really issue was that the input to calculatePostIapAggregateInterference() API was wrong. Q_p – Mg_p (esc_iap_threshold value) was passed as input. It shall be corrected to pass Q_p (esc_thresh_q value). 
```

    Qp        Ap
 10---- - 10 ----                          
    10        10                ASASp s
{------------------}    + 10 -----------
         Nsas                   10

```
Please note, interf.IN_BAND_INSERTION_LOSS for ESC protection threshold was accommodated for channel frequencies greater than 3650MHz only. For rest of the pass band insertion loss was not considered. This is corrected by accommodating insertion loss at the time of esc_iap_threshold calculation. 
 
#651

### DIFF
--- a/src/harness/reference_models/iap/iap.py
+++ b/src/harness/reference_models/iap/iap.py
@@ -305,7 +305,7 @@ def performIapForEsc(protected_entity, sas_uut_fad_object, sas_th_fad_objects):
   # a pre-defined Pre-IAP headroom (Mg) at each protection threshold(Q)
   # Final IAP protection threshold for ESC is considered after subtracting 
   # insertion loss
-  esc_iap_threshold = interf.dbToLinear(esc_thresh_q - MARGIN_ESC_DB - interf.IN_BAND_INSERTION_LOSS)
+  esc_iap_threshold = interf.dbToLinear(esc_thresh_q - MARGIN_ESC_DB + interf.IN_BAND_INSERTION_LOSS)
 
   grant_objects = data.getGrantObjectsFromFAD(sas_uut_fad_object, sas_th_fad_objects)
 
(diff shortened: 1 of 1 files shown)


src/harness/reference_models/iap/iap.py
```diff
@@ -303,7 +303,9 @@ def performIapForEsc(protected_entity, sas_uut_fad_object, sas_th_fad_objects):
                    + 10*np.log10(IAPBW_HZ / ESC_RBW_HZ)
   # Actual protection threshold used for IAP - calculated by applying
   # a pre-defined Pre-IAP headroom (Mg) at each protection threshold(Q)
-  esc_iap_threshold = interf.dbToLinear(esc_thresh_q - MARGIN_ESC_DB)
+  # Final IAP protection threshold for ESC is considered after subtracting 
+  # insertion loss
+  esc_iap_threshold = interf.dbToLinear(esc_thresh_q - MARGIN_ESC_DB - interf.IN_BAND_INSERTION_LOSS)
```
nasrinivasa: Shouldn't you be adding the insertion loss to the threshold value as the threshold has to be increased by that value since we are doing IAP at the filter input?

jni2000: thanks, you are right. Fixed now.

Issue #654, opened 2018-05-16
filterChannelsByFrequencyRange not returning correct results


arhannan:
filterChannelsByFrequencyRange seems not be be returning the  correct channels.
```
Example 1
channels = [{'frequencyRange' : {'lowFrequency': 3545, 'highFrequency': 3560}},
            {'frequencyRange' : {'lowFrequency': 3560, 'highFrequency': 3570}}, 
            {'frequencyRange' : {'lowFrequency': 3570, 'highFrequency': 3575}}]

frequency_range = { 'lowFrequency': 3550, 'highFrequency': 3575}
print filterChannelsByFrequencyRange(channels, frequency_range)            

Example 2
channels = [{'frequencyRange' : {'lowFrequency': 3550, 'highFrequency': 3560}},
            {'frequencyRange' : {'lowFrequency': 3560, 'highFrequency': 3570}}, 
            {'frequencyRange' : {'lowFrequency': 3570, 'highFrequency': 3580}}]

frequency_range = { 'lowFrequency': 3565, 'highFrequency': 3585}
print filterChannelsByFrequencyRange(channels, frequency_range)
``` 
The result is 
Example 1 result
> [{'frequencyRange': {'lowFrequency': 3560, 'highFrequency': 3570}}, {'frequencyRange': {'lowFrequency': 3570, 'highFrequency': 3575}}]
Example 2 result
> [{'frequencyRange': {'lowFrequency': 3570, 'highFrequency': 3580}}]

The right results should be
Correct example 1 result
> [{'frequencyRange': {'lowFrequency': 3545, 'highFrequency': 3560}}, {'frequencyRange': {'lowFrequency': 3560, 'highFrequency': 3570}}, {'frequencyRange': {'lowFrequency': 3570, 'highFrequency': 3575}}]

Correct example 2 result
> [{'frequencyRange': {'lowFrequency': 3560, 'highFrequency': 3570}}, {'frequencyRange': {'lowFrequency': 3570, 'highFrequency': 3580}}]

PR #655, opened 2018-05-16
bug fix filterChannelsByFrequencyRange

arhannan:
filterChannelsByFrequencyRange was not returning all channel in the input frequency range

### DIFF
--- a/src/harness/util.py
+++ b/src/harness/util.py
@@ -405,9 +405,9 @@ def filterChannelsByFrequencyRange(channels, freq_range):
   return [
       channel for channel in channels
       if
-      freq_range['lowFrequency'] <= channel['frequencyRange']['lowFrequency']
+      channel['frequencyRange']['lowFrequency'] <= freq_range['highFrequency'] 
       and
-      channel['frequencyRange']['highFrequency'] <= freq_range['highFrequency']
+      channel['frequencyRange']['highFrequency'] >= freq_range['lowFrequency']
   ]
 
 def _orderAttributes(obj):


src/harness/util.py
```diff
@@ -405,9 +405,9 @@ def filterChannelsByFrequencyRange(channels, freq_range):
   return [
```
kate-harrison: Can you please update the docstring to read "Returns channels which partially or fully overlap with 'freq_range'." ? Then the intention of this function is clearer.

jni2000: Kate answered my question. Thanks. Approved after updating the docstring.

gilschieber: I don't understand where was the problem with this function. Could you give a clear example?
In my opinion, the new version does not give the expected result (and the SIQ3 test now fails...)

This function was a helper function only used by SIQ_3 testcase, in order to facilitate the different checks used in this testcase. 
Maybe the name of the function is not well chosen, and we could modify it to something like returnChannelsWithinFrequencyRange 

The objective of the function is that, given a list of channels (5 or 10 MHz), and given a frequency range (f_min, f_max), returns the list of channels that are included in the frequency range.

Let's take a simple example where the initial input channel list contains only 1 channel: 
3620000000 - 3630000000
Let' s take the frequency range as:
1) 3630000000 - 3650000000

The expected result is [None] since the channel is not within the frequency range.
With the new version, the result is the initial channel 3620000000 - 3630000000.
Clearly not the expected result.

By the way, this test is now failing after this change.

arhannan: see [Issue 654](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues/654).
We will make a change to address the bug you have raised.

PR #657, opened 2018-05-16
Fix syntax errors in PAT testcase

harinineelu:


harinineelu:
Thanks both of you! Merged #657

PR #659, opened 2018-05-16
Fix PreIAP and Aggregate Interference use of deploymentParam

TaranFurnell:


kate-harrison:
Yes, these require an update similar to the one you did for other test data files in order to be consistent. However, since we don't use the land category in the pre-IAP filtering, the change won't affect the actual code or unit tests and the changes in this PR are sufficient to get the code running. Perhaps you can make the update in the next PR so that we keep all the test data consistent?

PR #662, opened 2018-05-17
IAP: bug fixes and cleanup

sbdt-google:
Several bug fixes for IAP
General cleanup of the IAP module (aggregate_interference cleanup will follow):
  - ESC data extraction done on single place in common/data.py (so will be reused in aggregate_interference and will avoid code duplication)
  - FSS parameter extraction was wrong for tilt. Remove duplication by using the common/data.py routine already used elsewhere
  - building the final post-IAP dictionary was wrong: the longitude was constantly erased and only one was kept per latitude (note that this expains some issues see in KeyError exception running MCP)
  - general cleanup to use the common naming for variables, and improve the readability of the main IAP loop

PR #664, opened 2018-05-17
Aggregate interference: Cleanup and bug fix

sbdt-google:
- General cleanup of doc, 
- Reuse of the common/data.py routines: getFssInfo and getEscInfo  to avoid code duplication. 
  This fixes some issues for FSS (tilt and AMSL). This code assume that the getEscInfo() routine has been 
  added in common/data.py (see PR #662).
- Move the aggr_interference- only routine from interference.py module to aggregate_interference.py module.


sbdt-google:
Thanks, will wait for the other PR as the getEscInfo() routine has been added on that other PR.

sbdt-google:
Thanks submitted now

PR #665, opened 2018-05-17
filterChannelsByFrequencyRange fix

arhannan:
Modified logic to use intersection

### DIFF
--- a/src/harness/util.py
+++ b/src/harness/util.py
@@ -405,9 +405,9 @@ def filterChannelsByFrequencyRange(channels, freq_range):
   return [
       channel for channel in channels
       if
-      channel['frequencyRange']['lowFrequency'] <= freq_range['highFrequency'] 
-      and
-      channel['frequencyRange']['highFrequency'] >= freq_range['lowFrequency']
+      (bool(set(range(channel['frequencyRange']['lowFrequency'], 
+                channel['frequencyRange']['highFrequency'])).intersection(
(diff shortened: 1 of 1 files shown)


src/harness/util.py
```diff
@@ -405,9 +405,9 @@ def filterChannelsByFrequencyRange(channels, freq_range):
   return [
       channel for channel in channels
       if
-      channel['frequencyRange']['lowFrequency'] <= freq_range['highFrequency'] 
-      and
-      channel['frequencyRange']['highFrequency'] >= freq_range['lowFrequency']
+      (bool(set(range(channel['frequencyRange']['lowFrequency'],
```
kate-harrison: Can you explain what problem you're trying to solve? As far as I can tell, the logic is the same (confirmed if all the unit tests still pass) but the new code is much less efficient.

arhannan: Yes it is less efficient. See @gilschieber's comment [here](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/655#pullrequestreview-121118827)

kate-harrison: Good point, I agree with his comment. But if you change the ">=" on line 410 to be just ">" -- and make no other changes -- I think it resolves this problem and the problem reported in #654. Do you agree?

arhannan:
Thanks James and Kate. Updated

PR #670, opened 2018-05-18
Cert generation script for SAS UUTs

harinineelu:
Script to help ITS with cert generation for SAS UUTs


harinineelu:
Thanks James and Assane. Merged #670

PR #671, opened 2018-05-18
Fix for AMSL not taken into account (PPA creation, BPR, QPR and PAT)

sbdt-google:
Fixed for : 
  - PPA creation
  - test cases BPR, QPR and PAT

Also fixed a small bug in PPA creation: beamwidth is optional for CatA so support this.

Finally minor cosmetic cleanup on indentation not right.


jcereced:
We reviewed the code and it looks good.  I would like to also try to run it, if I can't by EOD i'll go ahead and approve based on the review

jcereced:
Sorry won't be able to run it today....will definitely run it once it's in Master so let's put it in.  Aproved!

PR #681, opened 2018-05-21
Assert high frequency for IAP FSS co-channel

RedaHossino:


src/harness/reference_models/iap/iap.py
```diff
@@ -523,8 +523,8 @@ def performIapForFssCochannel(fss_record, sas_uut_fad_object, sas_th_fad_objects
   fss_low_freq, fss_high_freq = fss_freq_range
 
   # Get channels for co-channel CBSDs
-  fss_high_freq = min(fss_high_freq, interf.CBRS_HIGH_FREQ_HZ)
-  protection_channels = interf.getProtectedChannels(fss_low_freq, fss_high_freq)
+  assert(fss_high_freq >= interf.CBRS_HIGH_FREQ_HZ)
```
kate-harrison: Consider adding a message to the assert (I think it's the third arg?) to help the tester understand what went wrong if this assert fails.

src/harness/reference_models/iap/iap.py
```diff
@@ -523,8 +523,8 @@ def performIapForFssCochannel(fss_record, sas_uut_fad_object, sas_th_fad_objects
   fss_low_freq, fss_high_freq = fss_freq_range
 
   # Get channels for co-channel CBSDs
-  fss_high_freq = min(fss_high_freq, interf.CBRS_HIGH_FREQ_HZ)
-  protection_channels = interf.getProtectedChannels(fss_low_freq, fss_high_freq)
+  assert fss_high_freq >= interf.CBRS_HIGH_FREQ_HZ, 'FSS high frequency should not be less than CBRS high frequency(Hz)'
```
sbdt-google: Are we sure that we want to use "assert" here, and not a traditional raise exception ?
assert can be completely disabled when launching your script with python -O or other option (optimize).
Why not= instead:
```
 if fss_high_freq >= interf.CBRS_HIGH_FREQ_HZ:
    raise ValueError('Fss High frequency should not be less than CBRS high frequency(Hz)')
```

src/harness/reference_models/iap/iap.py
sbdt-google: Are we sure that we want to use "assert" here, and not a traditional raise exception ?
assert can be completely disabled when launching your script with python -O or other optimize options. Best not to rely on assert to do data validation, this is more a debugging tool like in other languages.
 
Why not instead:
```
 if fss_high_freq >= interf.CBRS_HIGH_FREQ_HZ:
    raise ValueError('Fss High frequency should not be less than CBRS high frequency(Hz)')
```
By the way, there are 3 or 4 places in the test harness where we do such 'assert' and that we may want to change:
   - in interference.py
   - in util.py
   - in sas_test_harness.py
   - in test_harness_objects.py

Issue #693, opened 2018-05-22
WG4 change to SIQ.3: SIQ response for C1 is slightly changed


kate-harrison:
Instead of 
> upper edge of F to 3650 MHz

we now have

> upper edge of F to 3700 MHz

for C1 in SIQ.3. This should just be a matter of changing this line

>     self.assertChannelsOverlapFrequencyRange(channels_high[0], freq_range_high, constrain_low=True, constrain_high=False)

to

>     self.assertChannelsOverlapFrequencyRange(channels_high[0], freq_range_high_all, constrain_low=True, constrain_high=True)

and creating `freq_range_high_all` which spans `pal_high_frequency` to 3700 MHz.

harinineelu:
Fixed.

PR #694, opened 2018-05-22
Fix for Issue #693 (Fix C1 in SIQ.3)

harinineelu:


src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
```diff
@@ -222,12 +226,16 @@ def test_WINNF_FT_S_SIQ_3(self):
...
+          filterChannelsByFrequencyRange(channels, freq_range_excluded))
 
     # Checks for cbsd 1 (index 0)
     self.assertChannelsContainFrequencyRange(channels_low[0], freq_range_low)
-    self.assertChannelsOverlapFrequencyRange(channels_high[0], freq_range_high, constrain_low=True, constrain_high=False)
+    self.assertChannelsOverlapFrequencyRange(
+        channels_high[0],
+        freq_range_high_all,
+        constrain_low=True,
+        constrain_high=True)
```
jni2000: The original logic was to not constrain the high frequency edge of the channel to 3650 and thus will allow to go over it. The outer checking on all channels to not exceed 3700 will guarantee it is bounded. The modified logic will fail if the channel range is [3630 - F_high], when F_high < 3700, e.g. 3680, 3690 etc. Please double check. Can you give an example that the original logic will fail ?

kate-harrison: The logic was changed in today's WG4 meeting; please see issue #693 to understand the differences. Kambiz from Federated should also be aware of them if you want to double-check.

jni2000:
Javi

Can you please run a test with the highest channel frequency range [3630 - 3680] ? I think it is also valid that SAS does not always respond with [3630 - 3700].

Thanks.

src/harness/testcases/WINNF_FT_S_SIQ_testcase.py
jni2000: I did not get this update. Do you mean that we changed the test spec to set the check to [F_high, 3700] ? If so, I have no concern with this fix.

harinineelu:
Thanks Javi and James. Merged #694

Issue #685, opened 2018-05-21
ESC Record Antenna pattern values not as per spec


nasrinivasa:
https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/ed8642f2154d192300afda39c8c5f76cf13f6c37/src/harness/reference_models/iap/test_data/esc_0.json#L10

The antenna gain values should be referenced with respect to the bore-site but the current record is not doing so. This needs to be addressed.

kate-harrison:
I guess your point is that the high-gain points should centered around 0 degrees rather than around ~180 degrees?

James, can your team take a look at this? I think they were the original authors of this code.

jni2000:
ok, we will take a look of this one.

kate-harrison:
This was fixed in #696 so I'm closing the issue now.

PR #696, opened 2018-05-22
fix the azimuthRadiationPattern setting error (issue #685)

jni2000:
Also please check if we need to do the same fix for 

src/harness/testcases/testdata/esc_sensor_record_0.json

### DIFF
--- a/src/harness/reference_models/iap/test_data/esc_0.json
+++ b/src/harness/reference_models/iap/test_data/esc_0.json
@@ -1,1453 +1,1453 @@
-{  
-   "id":"esc_sensor/admin_0/esc_sensor_0",
-   "installationParam":{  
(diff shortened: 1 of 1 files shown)


jni2000:
Please review the fix for esc_record_0.json. Also, "src/harness/testcases/testdata/esc_sensor_record_0.json" is not in line with the spec. Do we need to make change to it too ?

jni2000:
This fix is for issue #685

nasrinivasa:
"src/harness/testcases/testdata/esc_sensor_record_0.json" seems to have an omni pattern, the antenna gain is the same for all angles. Not sure if it was intentional.

nasrinivasa:
I agree with you on that Kate.

Issue #688, opened 2018-05-21
ESC mask loss not taken into accounty in agg_intereference


sbdt-google:
For 3650-3680 MHz, the ESC mask loss taken into account in the IAP but not in the agg_interference check.

sbdt-google:
See proposal at #697

kate-harrison:
Fixed in the PR mentioned above, which will be submitted soon.

PR #697, opened 2018-05-22
Fix MCP by computing ESC interference post-filter

sbdt-google:
See also #688
The MCP fails because in some case (no peer SAS), the interference is tested against the fixed ESC threshold (-109dBm), without taking into account the rolloff filter.

Currently the ESC has a special processing compared to other entities (such as FSS blocking), in the sense that all computation are done at filter input by:
 - changing in the IAP loop the threshold by the reverse of filter rolloff
 - not taking into account the filter in the aggregate interference calculation

This should work in theory, although it introduces a special case in the code. It makes the MCP fails
however since we have the need to compare it to the absolute threshold when peer SAS.

There was some discussion one month ago (between Yi, Sam, and Navin mainly
in thread: 'ESC sensor protection discussion') about the processing
required for ESC in IAP and aggregate interference.

To me it seems more straightforward to simply compute the interference
using the ESC mask, ie at the filter output, as expected and done for example in FSS blocking case.

In this PR, I introduce the EscMaskLoss in the interference calculation for ESC, which then propogate
to both IAP and aggregate interference. This avoid putting a hack to compute the offset to apply to the threshold winthin the MCP testcase code. It also simplifies the code and make it faster.


kate-harrison:
Navin, can you please formally approve this PR so that it can be submitted? Thanks!

sbdt-google:
Yes please have a look to the code.
I am not sure why Federated took initially the path of doing the ESC calculation pre-filter, while all other are post filter (FSS cochannel and FSS blocking for example). This approach worked for a time, until we hit the MCP implementation. See explanation above in the PR description.

I discussed with Yi this change who said that he is fine with it if it makes things more consistent (everything computed post-filter, including the ESC) and fixes the issues seen in MCP test.

jni2000:
Ok, please can you ask Yi to communicate with Masoud on this and make sure that he is okay with it. We need to make sure this is an WG.4 agreement. I will approve it, but do please make sure Masoud and the rest of WG.4 are aware of this.

jni2000:
Sam and I just had a discussion on this and recalled the historical agreement we made with Google (Yi). The current implementation is what we agreed on and implemented in IAP. The current implementation in MCP is the part to change to adapt to what we did in IAP instead of the other way around as suggested here. We definitely need have a discussion on this between Yi and Masoud/Sam to settle this down. Please understand our concerns. Thanks.

nasrinivasa:
Theoretically, the results should not change whether you do your computation at the input of the filter or at the output, but from a consistency point of view, doing it at the output of the filter makes sense as we don't have to do anything special for ESC. So i am fine with the change to make the ESC interference calculation to be consistent with the other implementation

PR #667, opened 2018-05-17
fix FSS mask loss claculation issue

jni2000:
This PR is for fixing the issue as described in #652.

PR #702, opened 2018-05-23
FSS mask: alternative proposal to fix FSS mask loss.

sbdt-google:
PR #667 is a proposal to fix the FSS mask loss, which still has bugs according to me.
Because of disagreement on the remarks made on #667 , here is my proposed PR to fix the FSS Mask loss. In addition of fixing the reported bugs, it should also be much faster to run.

A unit test file is also added. Note that all the tests performed here fails in the other proposed solution.


sbdt-google:
Submitting now

nasrinivasa:
Serge, The code looks good (know you have already merged it)

PR #709, opened 2018-05-24
Fix propagation segmentation fault when Tx & Rx location identical. 

sbdt-google:
Fix crash if Tx and Rx location are the same.

Also disable the return of the "internals" information which are not used anywhere as part of test harness.
This contains big arrays and prevent efficient caching of the propagation results.


sbdt-google:
Thanks.

PR #713, opened 2018-05-24
Tiny update of PCR tests to use feature collection

RedaHossino:


src/harness/testcases/WINNF_FT_S_PCR_testcase.py
```diff
@@ -598,12 +598,12 @@ def test_WINNF_FT_S_PCR_3(self, config_filename):
 
     # Prepares the PAL Ids to trigger ppa creation request.
     pal_ids = [record['palId'] for record in pcr_1_test_config['palRecords']]
-
+    user_claimed_ppa_contour_feature_collection = utils.InsureFeatureCollection(config['userClaimedPpaContour'], as_dict=True)
```
kate-harrison: Why not put this in the config generation function? Then the configs will have a consistent format as well.

RedaHossino: I thought that it is simpler if we have only the geometry object on the config file, but you are right about the format of config files as well.
so I updated the code.

Issue #705, opened 2018-05-23
signal.alarm not cleaned up on test error


TaranFurnell:
Code using signal.alarm does not clean up the alarm on test failure. This can affect other tests if run in the same execution. Recommended approach: Change all signal alarms to do all time-based checks the same as in PR #690

kate-harrison:
James, are you able to address this? I think it should be fairly straightforward given that Taran has provided an example in #690.

jni2000:
ok, will work on this

kate-harrison:
Fixed in #715 which will be submitted soon.

PR #715, opened 2018-05-24
fix issue #705 - signal alarm not cleaned up on test error

jni2000:


src/harness/sas_testcase.py
```diff
@@ -228,13 +226,8 @@ def TriggerDailyActivitiesImmediatelyAndWaitUntilComplete(self):
     - If the status is not changed within 2 hours it will throw an exception.
     """
     self._sas_admin.TriggerDailyActivitiesImmediately()
-    signal.signal(signal.SIGALRM,
-                  lambda signum, frame:
-                  (_ for _ in ()).throw(
-                      Exception('Daily Activity Check Timeout')))
-
     # Timeout after 2 hours if it's not completed
```
kate-harrison: Hm, I think we meant to change this to 24 hours (the meeting minutes from April 24 are not very clear but I think that was the general agreement). Would you mind making that modification now?

jni2000: ok

jni2000:
thanks for the quick review even before I roll out of the page :)

PR #718, opened 2018-05-24
Fix for getFqdnLocalhost()/getUnusedPort()

norbertbarbosa:
Following discussion about getFqdnLocalhost()/getUnusedPort() in [Sas_test_dev]

Here the proposed solution, assuming Dorch comment about the ITS system architecture:
We assume that ITS will use a different machine for each SAS provider,
directly connected to internet to run the TH.
This machine has a firewall that open only a limited contiguous number of ports.

The previous solution of using 'random available' port sdon't fit this configuration.
Also it can produce problem with generated config file where the saved port became
unavailable.

The new proposed solution:
- we use a configured range of port, that must be keep available only for the TH
- in each webserver implementation (database.py, sas_test_harness.py), we:
  - always use '' when creating the server socket to listen to all interface
  - in the destructor, we ensure that we release the allocated port, to allow
    reusing port accross testcase
- configuration is done in a new test.cfg
- we correctly manage the generation of default config file, which occur during the import

### DIFF
--- a/src/harness/database.py
+++ b/src/harness/database.py
@@ -71,7 +71,7 @@ class DatabaseServer(threading.Thread):
     self.port = port
     self.base_url = 'http' + ('s' if https else '') + '://' + self.address
     self.setDaemon(True)
-    self.server = DatabaseHTTPServer(('localhost', port), DatabaseHandler, name, authorization)
+    self.server = DatabaseHTTPServer(('', port), DatabaseHandler, name, authorization)
     if https:
       self.server.socket = ssl.wrap_socket(
           self.server.socket,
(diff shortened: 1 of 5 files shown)


jcereced:
James, are you saying you need to test it or that you need someone else to test it?

jni2000:
We do not have time to do the test, if someone else tested, it is good to go. Logic looks good.

norbertbarbosa:
Note I just finished some intensive tests on our side...
I did another commit to fix a potential error when using min_port=-1, due to the prone error conversion as int.

I will wait one hour more, and if no comment I will submit it.
Thanks

PR #719, opened 2018-05-24
Minor fix issue caused by recent addition of logging.

sbdt-google:
Was loggin "region" instead of "region_type". This cause the code to fail.
 
Also fixed the doc using region_type everywhere for consistency.


sbdt-google:
Thanks!

PR #730, opened 2018-05-26
Update of DPA and Exclusion Zone

sbdt-google:
- E-DPA: update of some of the missing distances. Basically, the difference is now:
  Alameda:128km
  Pensacola: 304km
  Webster Field: 288km
  Pascagoula: 288 km

- Update of the exclusion zone KML with Nellis Air Force Base.
- Fix DPA reading to not acept invalid geometry.
  This will provide automatic issue detection in future updated KML.
   For now the Coastal correct the geometries until Hawaii6 is fixed by NTIA.


sbdt-google:
Thanks

PR #731, opened 2018-05-26
Remove multiprocess Manager

sbdt-google:
+ Remove the use of the mp Manager, and using the pool mechanism instead.
Some profiling of the Manager as used in the code shows that it does not
scale at all, as it constantly fork new process and lock during
contention when updating the shared dict (plus IPC overheads).
As a consequence, beyond a few processes, the actual running time is
constant or actually increase. All gains from parallel calculation is
lost on the shared objects management, which can take typically:
  - 60% for 4 processes 
  - 85% for 12 processes
  - and so on..

Using simply the regular returned value from pool.map is much more
scalable, in addition of being simple.

+ Add profile script in tools/benchmark + a `profpool.py` utility in tools/
to enable the ability to profile each process. To run, simply launch the
tools/benchmark/prof_agg_interp.py script.

### DIFF
--- a/src/harness/reference_models/common/mpool.py
+++ b/src/harness/reference_models/common/mpool.py
@@ -48,7 +48,7 @@ pool.apply_async(...)
 
 from functools import partial
 import multiprocessing
-
+import time
 
 class _DummyPool(object):
   """A dummy pool for replacement of `multiprocessing.Pool`
@@ -105,7 +105,11 @@ def GetNumWorkerProcesses():
 
 
 def _partial_fn(fn):
+  # sleep to avoid returning too fast so each worker
+  # gets one job in the RunOnEachWorkerProcess.
+  time.sleep(0.5)
   return fn()
+
 
 def RunOnEachWorkerProcess(fn, * args, **kwargs):
   """Runs a function on each of the pool process."""
--- a/src/harness/reference_models/common/mpool.py
+++ b/src/harness/reference_models/common/mpool.py
@@ -88,14 +88,20 @@ _pool = _DummyPool()
 _num_workers = 0
 
 # External interface
-def Pool():
+def Pool(reinit=False):
   """Returns the worker pool.
 
   It supports routines:
     map()
     apply_async()
   And other `multiprocessing.Pool` routines if not a dummy pool.
+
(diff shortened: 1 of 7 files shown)


src/harness/reference_models/common/mpool.py
```diff
@@ -105,8 +111,12 @@ def GetNumWorkerProcesses():
 
 
 def _partial_fn(fn):
+  # sleep to avoid returning too fast so each worker
+  # gets one job in the RunOnEachWorkerProcess.
+  time.sleep(0.5)
```
jni2000: should we consider using a random number 0.5 - 1.0 to be more reliable ?

src/harness/reference_models/iap/iap.py
```diff
@@ -134,12 +128,8 @@ def iapPointConstraint(protection_point, channels, low_freq, high_freq,
       protection_constraint)
 
   if not neighbor_grants:
-    for j in xrange(len(channels)):
-      asas_interference.UpdateAggregateInterferenceInfo(
-          protection_point[1], protection_point[0], 0)
-      aggr_interference.UpdateAggregateInterferenceInfo(
-          protection_point[1], protection_point[0], 0)
-    return
+    return (protection_point[1], protection_point[0],
```
jni2000: I understand that you are trying to return the original asas_interference and aggr_interference here and all other places in stead of updating them. I will have to true you that they are logically correct and not affect the calculation results. If you test results are correct, I have no objections, but do please make sure tested and verified.

src/harness/reference_models/common/mpool.py
sbdt-google: Not really in this case. This is just for the use of the little function nbelow which insure that you run  one function on each pool.
This is only used in the `test_main.py` script which allows to configure each pool geo data drivers for example.

src/harness/reference_models/iap/iap.py
sbdt-google: Yes I checked carefully and compared the examples before and after. They give exactly the same results on all 15 decimals printed

sbdt-google:
No algorithmic change. 
Just programmatic.

PR #732, opened 2018-05-26
Add memoizing propag cache around repeated calculations.

sbdt-google:
Only for the Hybrid model (used for PPA and GWPZ) as these dominates the calculation.
This avoids redoing the propag calculation for parts of grant crossing
the 5MHz channels, and the IAP loop.

As an example if we process grants of 20MHz in agg_interference, we decrease by 4x the lengthy calculation. For IAP it is even worse as there are several loops, and the computeInterference is 
recomputed each time that a grant is not satisfied and we retry with a different threshold or power.

2 lines added, everything else is just indentation.
See the "Hide whitespace change" in the button Diff Settings, in the Files Changed pane, to see the actual change of the 2 lines.


src/harness/reference_models/iap/iap.py
```diff
@@ -198,75 +200,77 @@ def iapPointConstraint(protection_point, channels, low_freq, high_freq,
   grants_overlap_flag = {}
   grants_interference = {}
 
-  while num_unsatisfied_grants > 0:
-    for g_idx, grant in enumerate(neighbor_grants):
+  with cache.CacheManager(wf_hybrid.CalcHybridPropagationLoss):
```
jni2000: even though diff showed significant differences, I guess this is the only addition added, right ? Same comment on the other file.

kate-harrison: Check https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/732/files?w=1

jni2000: Good, thanks.

PR #733, opened 2018-05-29
Update with latest DPA and exclusion zone.

sbdt-google:
Hawaii6 is now fixed, so reenable the invalidity test for coastal DPA.
Note: neighbor distances are still not fully defined, so using default
values.


sbdt-google:
Thanks all for checking.
Thao: yes only Hawaii 6 was invalid, and Nickconverted it to a polygon with hole instead of a "horseshoe" type polygon with self-touching. 
Others Hawai seems to have changed but looking quickly it only looks like a rotation of the points (meaning there are shifted but kept unchanged).

Thao-Nguyen:
I see. Thanks!

PR #734, opened 2018-05-31
Updates the cert generation scripts:

harinineelu:
(1) Fixes the problem where root/ was deleted in generate_fake_certs.sh but required by generate_sas_cert.sh.
(2) Adds SAN support to generate_sas_cert.sh.
(3) Now also generates an ECC cert for SAS UUT.
(4) Fixes the filenames of the SAS UUT certs so that they do not overwrite certs generated by generate_fake_certs.sh.


harinineelu:
Thanks. Merged #734

PR #735, opened 2018-05-31
Fix IPR2 issue caused by temporary DPA KML

sbdt-google:
DPA KML are not yet finalized in the sense that:
  - all mandatory OOB neighbor distances are currently set as NaN
  - the catA optional distance is also set to NaN instead of either being specified or left blank

This second point means that the code was using a NaN distance for CatA, and this breaks IPR2: 
you then have CatA CBSD in Oregon being in neighborhood of east coast DPA. 
Running such long range path loss calculation is extremely slow obviously as a lot of terrain tiles need to be loaded.

In addition fixed the channelization of the DPA channel for the case where the DPA bandwidth are not multiple of 10MHz. Although they are all in the current DPA KML, the original list of inland DPAs had radar operating only in small frequency band of few MHz. This change will support these cases if they are provided back in future portal DPA update.

### DIFF
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -418,7 +418,7 @@ def GetDpaProtectedChannels(freq_ranges_mhz, is_portal_dpa=False):
   channels = set()
   for freq_range in freq_ranges_mhz:
     min_freq = int(freq_range[0] / DPA_CHANNEL_BANDWIDTH) * DPA_CHANNEL_BANDWIDTH
-    max_freq = int(freq_range[1] / DPA_CHANNEL_BANDWIDTH) * DPA_CHANNEL_BANDWIDTH
+    max_freq = freq_range[1]
     freqs = np.arange(min_freq, max_freq, DPA_CHANNEL_BANDWIDTH)
     for freq in freqs:
       channels.add((freq, freq + DPA_CHANNEL_BANDWIDTH))
--- a/src/harness/reference_models/dpa/dpa_mgr_test.py
+++ b/src/harness/reference_models/dpa/dpa_mgr_test.py
@@ -1,0 +1,56 @@
+#    Copyright 2018 SAS Project Authors. All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -418,7 +418,7 @@ def GetDpaProtectedChannels(freq_ranges_mhz, is_portal_dpa=False):
   channels = set()
   for freq_range in freq_ranges_mhz:
     min_freq = int(freq_range[0] / DPA_CHANNEL_BANDWIDTH) * DPA_CHANNEL_BANDWIDTH
-    max_freq = int(freq_range[1] / DPA_CHANNEL_BANDWIDTH) * DPA_CHANNEL_BANDWIDTH
+    max_freq = freq_range[1]
```
jni2000: why we do not need to do the same for min_freq, does this assume that min_freq  is always on the DPA_CHANNEL_BANDWIDTH boundary ?

jni2000:
Yes I agreed with what you explained. My question was that why didn't you make the same change to the freq_min as you did to freq_max.

On Jun 1, 2018 2:24 AM, Serge Barbosa <notifications@github.com> wrote:

@sbdt-google commented on this pull request.

________________________________

In src/harness/reference_models/dpa/dpa_mgr.py<https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/735#discussion_r192306937>:

> @@ -418,7 +418,7 @@ def GetDpaProtectedChannels(freq_ranges_mhz, is_portal_dpa=False):
   channels = set()
   for freq_range in freq_ranges_mhz:
     min_freq = int(freq_range[0] / DPA_CHANNEL_BANDWIDTH) * DPA_CHANNEL_BANDWIDTH
-    max_freq = int(freq_range[1] / DPA_CHANNEL_BANDWIDTH) * DPA_CHANNEL_BANDWIDTH
+    max_freq = freq_range[1]


The code does not assume that the radar bandwidth starts at a multiple of 10MHz. In fact in the original list provided, there was some radar with frequency starting at things like 3582 (for experimental radar). But the goal here is to compute the channels covering the radar bandwidth.
As an example, if you have a radar covering 3582-3597, then you want one DPA channel between 3580-3590, and one for 3590-3600. This was not the case with previous code which would miss the last one.
Note that for now the DPA KML provided only have bandwidth as a multiple of 10MHz, but it is best not to rely on this fact for future update.

kate-harrison:
James, take a careful look at line 422 and think through the example that Serge gave. I initially had the same questions but Serge has convinced me that the code is correct.

jni2000:
OK, got it. Thanks. Approved.

On Jun 1, 2018 11:42 AM, Kate <notifications@github.com> wrote:

James, take a careful look at line 422 and think through the example that Serge gave. I initially had the same questions but Serge has convinced me that the code is correct.

PR #737, opened 2018-06-02
Fix xPR tests for interference check when no peer SAS.

sbdt-google:
The absolute threshold need to be scaled to be relative to same
bandwidth as the one used for aggregate interference calculation,
ie IAP_BW (5MHz) in most cases, except for FSS blocking where it
is a general threshold.


sbdt-google:
Thanks

PR #739, opened 2018-06-04
Fix flakiness in test cases which inject PPAs

harinineelu:
...by ensuring that the PAL records and the PPA record have the same time range.


harinineelu:
Thanks. Merged #739

PR #741, opened 2018-06-04
MCP updates

harinineelu:
These are the changes agreed upon in this morning's WG4 call (see the WG4 mailing list for the latest draft).


kate-harrison:
Yup, definitely tested :) Thanks for the quick review!

PR #744, opened 2018-06-05
Update IPR test cases according to the latest spec changes

harinineelu:
... and add logging to existing test cases.

### DIFF
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -640,3 +640,4 @@ def BuildDpa(dpa_name, protection_points_method=None, portal_dpa_filename=None):
              freq_ranges_mhz=freq_ranges_mhz,
              neighbor_distances=neighbor_distances,
              monitor_type=monitor_type)
+
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -568,7 +568,7 @@ def _CalcTestPointInterfDiff(point,
 
 
 
-def BuildDpa(dpa_name, protection_points_method=None):
+def BuildDpa(dpa_name, protection_points_method=None, portal_dpa_filename=None):
   """Builds a DPA parameterized correctly.
 
   The DPA special parameters are obtained from the DPA database.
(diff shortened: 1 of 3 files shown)


src/harness/testcases/WINNF_FT_S_IPR_testcase.py
```diff
@@ -905,3 +989,294 @@ def test_WINNF_FT_S_IPR_6(self, config_filename):
...
+        })
+    using_esc_dpa = 'escDpa' in config
+    using_portal_dpa = 'portalDpa' in config
+    self.assertTrue(
+        using_esc_dpa ^ using_portal_dpa,
+        msg=
+        'Invalid config: must use exactly one ESC-monitored DPA OR one portal-controlled DPA.'
+    )
+    if using_esc_dpa:
+      self.assertLessEqual(
```
credtech: Shoudn't it be **assertEqual** instead of assertLessEqual here? (i.e raise an error if the low frequency is not 3540 MHz)

kate-harrison: I think assertLessEqual is correct since we need to permit activation of DPAs in 3550-3650 (whereas using assertEqual would restrict only to checking aggregate interference to always-active DPAs).

credtech: You're right, thanks for the clarification!

kate-harrison: Thanks for checking! Does the rest of the PR look good?

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -612,7 +612,7 @@ def BuildDpa(dpa_name, protection_points_method=None):
     monitor_type = 'esc'
   except KeyError:
```
redtechnologies: This error won't be raised if the name of the portal dpa provided in the config file (field dpaId) happens to exist in the default ESC NTIA file.

=> to avoid ending up with the wrong portal DPA points, it may be safer not to call GetCoastalDpaZones when building a portal DPA (e.g, by adding an optional parameter "isPortalDpa" to BuildDpa)

redtechnologies: ok to leave the code unchanged. I agree this situation is not supposed to happen & can be easily detected. Thanks!

harinineelu:
Thanks. Merged #744

PR #748, opened 2018-06-08
Bug fix on the ESC neighborhood for CatA during IAP.

sbdt-google:
The special CatA with ESC case was done with the implicit assumption
that the channels are the 5MHz IAP channels. The code works in that
case (for example: aggregateInterference check).
But the IAP runs the overlap selection using the whole ESC band
3550-3680 to filter the grants in the neighborhood. The code was not
working in that condition and CatA beyond 3660 were kept in the
neighborhood, preventing the IAP loop to ever converge.


sbdt-google:
Thanks for the quick review.

Issue #749, opened 2018-06-08
generate_sas_cert.sh malforms x509 extensions


bradeales:
`openssl` incorrectly permits the entry of multiple `-extensions` options in the command line, but it only uses the last option provided. As a result, the current `generate_sas_cert.sh` script only creates the SAN x509 extension.  

Correct x509 extensions from `generate_fake_certs.sh` (but this does not provide ability to add SAN, nor generate ECC SAS certs):
```bash
        X509v3 extensions:
            X509v3 Subject Key Identifier:
                62:A6:8A:FC:96:15:DC:A7:B8:4A:B5:4D:6C:3C:0F:5C:26:A5:FF:EE
            X509v3 Authority Key Identifier:
                keyid:C4:AB:A9:C0:A6:21:68:F4:11:21:97:CA:A5:BD:8C:32:65:46:E0:28

            X509v3 Basic Constraints:
                CA:FALSE
            X509v3 Key Usage: critical
                Digital Signature, CRL Sign
            X509v3 Extended Key Usage:
                TLS Web Server Authentication, TLS Web Client Authentication
            X509v3 Certificate Policies:
                Policy: 2.16.840.1.114412.2.1
                  CPS: https://www.digicert.com/CPS
                Policy: 1.3.6.1.4.1.46609.1.1.1

            X509v3 CRL Distribution Points:

                Full Name:
                  URI:http://localhost:9007/ca.crl
```

x509 extensions when no SAN is provided through `generate_sas_cert.sh`:  
```bash
[no extensions whatsoever]
```

x509 extensions when a SAN is provided through `generate_sas_cert.sh`:
```bash
        X509v3 extensions:
            X509v3 Subject Alternative Name:
                DNS:foo.bar.com, DNS:bar.com
```

kate-harrison:
Thanks for the thorough report and analysis! We'll have a PR up for this very soon. When it's merged will depend on when another SAS Admin can review it, but of course you're always welcome to use the draft.

kate-harrison:
Fixed in #750.

PR #750, opened 2018-06-08
Use all extensions during SAS cert generation.

harinineelu:
Fixes Issue #749


jni2000:
The script changes look good to me, we are in process of verifying it. Will approve asap if no concerns.

PR #752, opened 2018-06-11
add simple MCP check and update FAD.2 device 4

RedaHossino:
- add simple MCP check for initial cbsdRequests
- update FAD.2 device 4, so the Eirp of its Grant is reduced after CPAS(the reference model result now for this grant is 12 dBm)

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_MCP_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_MCP_testcase.py
@@ -46,13 +46,13 @@ ONE_MHZ = 1000000
 
 class McpXprCommonTestcase(sas_testcase.SasTestCase):
 
-  def getEmptyCbsdRequestsWithDomainProxies(numberOfElements):
+  def getEmptyCbsdRequestsWithDomainProxies(self, number_of_elements):
     empty_cbsd_records_domain_proxy = {
         'registrationRequests': [],
         'grantRequests': [],
         'conditionalRegistrationData': []
     }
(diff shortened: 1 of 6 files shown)


src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -80,6 +80,10 @@ def checkFadConfiguration(fad_config):
           len(config['domainProxyConfigs']),
           'Mismatch in the number of domain proxies and the configuration for this iteration: %s.'
           % str(iteration_data))
+      self.assertLessEqual(
```
kate-harrison: I think we want these to be equal, right?

RedaHossino: I agree, I chose the easiest way to have this check, but I agree that it is better to have more strict check, please check the update.

src/harness/testcases/WINNF_FT_S_MCP_testcase.py
```diff
@@ -46,6 +46,14 @@
 
 class McpXprCommonTestcase(sas_testcase.SasTestCase):
 
+  def getEmptyCbsdRequestsWithDomainProxies(numberOfElements):
```
kate-harrison: I think you need to have `self` as the first arg so that you can call as `self.getEmptyCbsdRequestsWithDomainProxies(2)` later.

When you make that change, can you also change to `number_of_elements` so that the var naming style is more consistent?

PR #755, opened 2018-06-12
Fix device_c heightType and location for a contour with a testable signal level.

harinineelu:
1. Change device_c heightType to 'AGL' in PCR_1, PCR_2 and PCR_4.
2. Change device_c location in PCR_1


harinineelu:
Thanks. Merged #755

PR #756, opened 2018-06-14
Trigger DPA activation correctly in MCP.1

TaranFurnell:
- Check for a DPA being already active (and skip its activation if that is the case, will log a warning but continue running the test)
- Deactivate DPAs in the dpaDeactivationList


TaranFurnell:
Hi @jni2000,

Can you please re-review? We needed to move some things around to support the case of having no peer sases.

Thanks

kate-harrison:
James, can you please take another look at this PR? I think it's really important to include it in this week's release.

PR #757, opened 2018-06-14
Improve GridPolygon.

sbdt-google:
- Reduce the floating point accumulation errors.
- Make sure border points are included in falling exactly on multiple of resolution, even given the small floating point errors.


sbdt-google:
Thanks.

PR #758, opened 2018-06-15
Remove erroneous line from IPR.7.

harinineelu:


kate-harrison:
It doesn't actually have a shutdown() method. Deletion should be sufficient.

harinineelu:
Thanks. Merged #758

PR #759, opened 2018-06-15
By default, allow NLCD voting

harinineelu:
...even when the NLCD data isn't available (uses the default value instead). Without this change, GWPZs in e.g. Hawaii will cause the test harness to crash.


harinineelu:
Thanks, Merged #759

PR #760, opened 2018-06-16
PCR: avoid multipolygons in default configurations.

sbdt-google:
PCR: New routine to shrink properly the PPA polygon.

By shrinking a polygon with shapely buffer(), spurious micro polygons
may be created, and the resulting PPA is then a MultiPolygon which
is forbidden by the spec. This new routine allows to more simply
shrink a PPA polygon while insuriung it is "clean" single polygon.


src/harness/reference_models/geo/utils.py
```diff
@@ -372,6 +372,45 @@ def PolygonsAlmostEqual(poly_ref, poly, tol_perc=10):
...
+    raise ValueError('Input is not a polygon.')
+  poly = poly.buffer(shrink_deg)
+  if isinstance(poly, sgeo.MultiPolygon):
+    # Cleanup multi-polygon.
+    max_poly_area = max(p.area for p in poly)
+    polys = [p for p in poly if p.area > min_poly_ratio * max_poly_area]
+    if len(polys) > 1:
+      raise ValueError('Impossible to shrink polygon '
+                       'into a proper single polygon')
+    poly = polys[0]
```
jni2000: do we always return the first polygon in multi-polygon case ? or should we just simply raise exception ?

sbdt-google: After trying to remove spurious polygons with small area, if we still have a MultiPolygon, then we raise an exception and stop: the configuration is bad, as the spec speficies that a PPA shall be a single polygon.
It is the task of the SAS (and of the ITS for the test cases) to feed single Polygon.
In fact  we might make another PR with some validation code, such as if a PPA config contains something else than a single Polygon, it stops with an error to avoid such occurence

jni2000: thanks. I agree that we shall check at the config time.

PR #763, opened 2018-06-21
Fix  ShrinkAndCleanPolygon to actually shrink and not expand.

sbdt-google:
A minus was missing. As a consequence the default PPA for PCR3 and PCR7 are expanded instead of shrinked (and vice versa for PCR6)


RedaHossino:
thanks for the change, looks good :)

RedaHossino:
so submitting this PR will depend on the decision of the SAS admins for V&V and delivery process to ITS in the related mail thread.

kate-harrison:
As discussed in the email thread, I will submit this PR now which will restart the V&V period. I will also send an email notifying the rest of test dev of the restart.

PR #764, opened 2018-06-25
Fix BPR.2 Issue by adding Multi_lineString case

RedaHossino:
the two cases treated in the intersection with canadian border are linestring & multi linestring.

the other cases are not treated and I don't think they can happen e.x.if the intersection is only one point.

### DIFF
--- a/src/harness/reference_models/geo/utils.py
+++ b/src/harness/reference_models/geo/utils.py
@@ -446,11 +446,11 @@ def GetClosestCanadianBorderPoint(latitude, longitude, max_dist_km):
     return None
   points_dists = []
   # Find closest point
-  if border_cap.type == 'LineString':
-    points_dists.extend(_distancesOfLineString(latitude, longitude, border_cap))
-  elif border_cap.type == 'MultiLineString':
-    for line in border_cap.geoms:
-      points_dists.extend(_distancesOfLineString(latitude, longitude, line))
+  if border_cap.type in ['LineString', 'Point']:
+       points_dists.extend(_distancesOfPoints(latitude, longitude, border_cap))
+  elif border_cap.type in ['MultiLineString', 'MultiPoint']:
+        for element in border_cap.geoms:
+          points_dists.extend(_distancesOfPoints(latitude, longitude, element))
   else:
     raise ValueError(border_cap.type + ', not treated type of intersection with canadian border')             
   closest = min(points_dists)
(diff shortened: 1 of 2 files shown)


src/harness/reference_models/geo/utils.py
```diff
@@ -442,19 +442,28 @@ def GetClosestCanadianBorderPoint(latitude, longitude, max_dist_km):
...
     return None
+  points_dists = []
   # Find closest point
-  points_dists = [(vincenty.GeodesicDistanceBearing(latitude, longitude, point[1], point[0]),
-                   point)
-                  for point in zip(*border_cap.xy)]
+  if border_cap.type == 'LineString':
+    points_dists.extend(_distancesOfLineString(latitude, longitude, border_cap))
+  elif border_cap.type == 'MultiLineString':
+    for line in border_cap.geoms:
```
sbdt-google: Could be sufficient to do:
```for line in border_cap:```

All the MultiXXX thing are iteratating on the non Multi version

File: src/harness/reference_models/geo/utils.py [hunk 2]
```diff
@@ -442,19 +442,28 @@ def GetClosestCanadianBorderPoint(latitude, longitude, max_dist_km):
...
       sgeo.Point(longitude, latitude).buffer(max_delta))
-  if not border_cap or 'LineString' not in border_cap.type:
+  if not border_cap:
     return None
+  points_dists = []
   # Find closest point
-  points_dists = [(vincenty.GeodesicDistanceBearing(latitude, longitude, point[1], point[0]),
-                   point)
-                  for point in zip(*border_cap.xy)]
+  if border_cap.type == 'LineString':
```
sbdt-google: Since we are fixing this, it could be in theory possible that the intersection returns a Point or MultiPoint for edge cases.
So why not extending your fix for those cases:
``` if border_cap.type in ['LineString', 'Point']:
       points_dists.extend(_distancesOfLineString(latitude, longitude, border_cap))
    elif border_cap.type in ['MultiLineString', 'MultiPoint']:
        for line in border_cap.geoms:
          points_dists.extend(_distancesOfLineString(latitude, longitude, line))
```
Eventually rename the _distancesOfLineString to _distancesOfPoints
or something like that since it works on LineString and Point.

RedaHossino: thanks for you comment, I added the cases of point&multui-points but I couldn't add unit tests for that,
actually I tried few tests as the following(to get only one point):
    res3 = utils.GetClosestCanadianBorderPoint(45.1165, -71.35, border_point[2])
however it didn't create one point of intersection.
 if you think of another way to test that, I can do it.

sbdt-google: It is really a edge case, and difficult to reproduce when you are dealing with floating points.
Synthetically you can reproduce the case as following:
```
ls = sgeo.LineString([(0,-2), (0,0), (0,2)])
cap = sgeo.Point(1,0).buffer(1)
pt = cap.intersection(ls)
```
==> The intersection is a Point.

Since the candian border coorrdinates are double produced by a script, difficult to repoduce this case. You can mock it if you wish by reproducing the previous example.
Or since the code is simple, just submit it without relying on unit test for the Point case.

PR #766, opened 2018-06-25
Fix the non-standard layout/encoding of CRL files

pmarks-net:
I recommend looking at the commits individually, to separate the signal (CRL-related fixes) from the noise (formatting cleanups).

This pull request gives each CA its own database and RFC-compliant CRL file.

In terms of ITS using this code, there is one key functional difference:

- Before: There is a single CRL Distribution Point **URL**, stored in `simple_crl_server.py` and `openssl.cnf`.
  http://crl.example.com:1234/ca.crl
  
- After: There is a CRL Distribution Point **URL prefix**, stored in `simple_crl_server.py` and `include_crldp_base.sh`. Each CA generates its own suffix.

  `CRLDP_BASE = "http://crl.example.com:1234/"`
  ↓
  http://crl.example.com:1234/root_ca.crl
  http://crl.example.com:1234/cbsd_ca.crl
  http://crl.example.com:1234/sas_ca.crl
  [ etc. ]

---

**Use a distinct index.txt database for each CA.**

Prior to this change, all CAs stored their revoked certificates in the
same database.  Now, all openssl commands that depend on a database must
specify which CA to modify.

After running generate_fake_certs.sh, you can browse the db/ directory
to see certs (and revocations) belonging to each CA.

---

**Generate one CRL per CA, in DER format.**

Previously, we concatenated several PEM-formatted CRLs into a single
ca.crl file, but https://tools.ietf.org/html/rfc5280#section-4.2.1.13
specifies that each URI "MUST point to a single DER encoded CRL".

Changes include:

- The generate scripts emit .crl files in DER format, and pass
  CA-specific CRLDP values to openssl.

- fake_sas accepts a crl_index_*.txt file, containing a list of
  DER-encoded CRLs.  While the files are now in the correct format,
  the distribution method is trivial compared to a real SAS.

- simple_crl_server writes the CRLDP to a one-line include file instead
  of openssl.cnf, and serves multiple *.crl files over HTTP.

- openssl.cnf obtains CRLDPs from its environment, and provides default
  values for its environment variables (because the error triggered by a
  missing variable was pretty cryptic.)

### DIFF
--- a/cert/openssl.cnf
+++ b/cert/openssl.cnf
@@ -44,11 +44,11 @@ TEST = 1.3.6.1.4.1.46609.1.999
 default_ca = CA_default  # The default ca section
 ####################################################################
 [ CA_default ]
-dir = ./root
+dir = ${ENV::OPENSSL_CNF_CA_DIR}
 certs = $dir/certs
 crl_dir = $dir/crl
-new_certs_dir = .
-database = index.txt
+new_certs_dir = $dir
+database = $dir/index.txt
 serial = $dir/serial
 RANDFILE = $dir/private/.rand
 
--- a/cert/openssl.cnf
+++ b/cert/openssl.cnf
@@ -18,6 +18,11 @@
 # =============                                                    #
 ####################################################################
 
+# Default ENV values.  If these strings appear in a generated certificate or
+# error message, then a script should've set the environment variable prior
+# to invoking openssl.
+OPENSSL_CNF_CA_DIR = _ca_dir_not_set_by_environment_
+OPENSSL_CNF_CRLDP = _crldp_not_set_by_environment_
 
 ####################################################################
 # Defining OIDS registered by The Wireless Innovation Forum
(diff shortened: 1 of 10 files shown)


pmarks-net:
I agree that it would make sense for every generated certificate (except the roots) to contain a CRL DP; the question is whether it's important enough to expand the scope of this PR.

When a SAS encounters a client certificate without a CRL DP, it could either:
1. Fail
2. Treat the certificate as non-revokable

So far, I've been assuming 2.  Do we have a specific reason to prefer 1?

plerebourg:
I agree assuming 2 as WINNF-TS-0022 specification specifies that CRL needs to be checked but cRLDistributionPoint extension is optional. So we don't have to add this extension on all configurations.

But for CA certificates they should have cRLDistributionPoint extension to allow their revocation as tested in SCS_16, SDS_16 and SSS_16.

pmarks-net:
Let's see, the following CRLs actually contain revoked certificates:
- cbsd_ca.crl 1
- proxy_ca.crl 1
- root_ca.crl 3
- sas_ca.crl 1

And [this paste](https://gist.github.com/pmarks-net/cd6365140d30672af5ad9f9d22e355df) shows which certificates have CRL DPs.

I see that there are no CRL DPs pointing to `root_ca.crl`, which will certainly break SxS_16, as you've noted.  I'll prepare a commit to fix this.

pmarks-net:
[This paste](https://gist.github.com/pmarks-net/b367f7f0d71fb7c882f628139f518bde) shows the CRL DPs after the above commit.

PR #774, opened 2018-06-28
Marking Grant Object as Not Authorized if it is terminated

sushantparkhi:
getAuthorizedGrants and hasAuthorizedGrant might cause an issue(Still give or create the Grants which are not Authorized) if the Grant is Terminated instead of Suspended. Updating the authorized_in_last_heartbeat flag might solve the issue.


sushantparkhi:
Thanks!

PR #779, opened 2018-07-10
Update the IPR.7 testcase to not require a portal DPA DB config...

harinineelu:
…when using an E-DPA.


harinineelu:
Thanks. Merged #779

PR #780, opened 2018-07-11
Some cleanup on data scripts

sbdt-google:
- Improve and cleanup the documentation of the `src/data/` scripts directory and data/fcc/ directories.

- Fix the usborder.py script to properly split the polygon crossing anti-meridian and regenerate the usborder.kml. Note that this has no effect on test harness since the usborder.kml is only used for simulation purposes. It is also used in the "default" method for DPA protection points generation for the splitting of front vs back zones, but the proposed change should not affect this in the CONUS part of US (only in alaska western islands).
Note that the KMZ does not display the conus border in Google Earth as that software is limited to about 60k vertices per pol;ygon, and we exceed this. Other GIS tool should show it properly though.


src/data/usborder.py
```diff
@@ -470,10 +471,41 @@ def AddBorderSegments(mexicoDoc, canadaDoc, coordinateLists):
...
+    r = list(reversed(ls))
+    lineStrings[k] = r
+
+# Cleanup the coordinates to be rounded correctly with 9 precision numbers
+# (ie about 0.1mm)
+for k, ls in enumerate(lineStrings):
+  new_ls = []
+  for c in ls:
+    xy = c.split(',')
+    new_ls.append('%.9f,%.9f,0' % (float(xy[0]), float(xy[1])))
```
KJaara: I wonder if there is a small risk of getting unvalid geometry by doing this.
May be, it's better to test the gemetry validity before writing to kml

Thank you
Kusay Jaare from RED Technologies

PR #782, opened 2018-07-12
pre-IAP filtering and iap shall be performed even when there is no peer SASs

jni2000:
In the current MCP codes, for step 11, pre-IAP filtering and IAP will only be performed when there is at least one peer SAS. If no peer SAS, the functions will be skipped. This will fail all the single SAS incumbent protection test cases such as EPR.1, PPR.1 and so on. This fix is to remove the number of peer SAS check to allow the functions be called in all cases including single SAS case.


kate-harrison:
There is a note in the MCP test case (in the test spec) which specifies that the aggregate interference should be compared against the absolute threshold -- i.e. pre-IAP filtering and IAP reference model should not be executed -- in the case where there are no peer SASes:

> Note 1: if NS = 0 (i.e. there are no peer SASs), Ap_IAP,ref is equal to the protection threshold and the IAP Reference Model need not be used.

Can you please explain further why you think the current code is incorrect?

jni2000:
after carefully review the test spec and the test codes, we agreed that the original codes are correct. No change is needed anymore. Thanks.

PR #785, opened 2018-07-18
Update BPR.2 to only check Grants which at least partially overlap 36…

harinineelu:
…50-3700

Update BPR.2 to only check Grants which at least partially overlap 3650-3700 (consistent with Arrangement R).


harinineelu:
Thanks. Merged #785

PR #786, opened 2018-07-20
FDB.6: Skip Relinquishment in case of Interference Response in Grant

sushantparkhi:
When SAS UUT Responded with Interference in Grant Response, there's no need for Relinquishment.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_FDB_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_FDB_testcase.py
@@ -1257,7 +1257,7 @@ class FederalGovernmentDatabaseUpdateTestcase(sas_testcase.SasTestCase):
         cbsd_ids_g1.append(resp['cbsdId'])
 
     # If SAS UUT Responded with SUCCESS for Grant then Proceed with Relinquishment or else skip it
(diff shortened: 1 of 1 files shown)


src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1256,22 +1256,25 @@ def test_WINNF_FT_S_FDB_6(self, config_filename):
...
-    for cbsd_id, grant_id in zip(cbsd_ids_g1, grant_ids_g1):
-      # Update Relinquishment Request content
-      relq_requests['relinquishmentRequest'].append({
-        'cbsdId': cbsd_id,
-        'grantId': grant_id
-      })
-    relq_responses = self._sas.Relinquishment(relq_requests)['relinquishmentResponse']
-    self.assertEqual(len(relq_responses), len(relq_requests['relinquishmentRequest']))
+    # If SAS UUT Responded with SUCCESS for Grant then Proceed with Relinquishment or else skip it
+    if len(grant_ids_g1):
```
kate-harrison: I think you can simplify with just `if grant_ids_g1` but of course this is also correct.

PR #789, opened 2018-07-27
Minor numerical correction in QPR testcase

harinineelu:
p/10 is integer division. Change to floating point.


harinineelu:
Thanks! Merged #789

PR #792, opened 2018-08-01
Fix Server Shutdown between Tests

sushantparkhi:
Added Server Shutdown as a part of Teardown. While running tests in Bulk Ports get exhausted, this fix will stop the servers if they are running in between tests.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_EPR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_EPR_testcase.py
@@ -15,8 +15,6 @@
 import json
 import os
 import sas
-import logging
-import threading
 import sas_testcase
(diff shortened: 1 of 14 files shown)


src/harness/testcases/WINNF_FT_S_EPR_testcase.py
```diff
@@ -31,7 +33,11 @@ def setUp(self):
     self._sas_admin.Reset()
 
   def tearDown(self):
-    pass
+    logging.info('Stopping all running servers, if any')
```
kate-harrison: Maybe it makes sense to add this as a method to SasTestcase so that the code does not diverge later?

PR #793, opened 2018-08-02
Implement changes to IPR.7 requested by WG4.

harinineelu:
Implement changes to IPR.7 requested by WG4.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_IPR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_IPR_testcase.py
@@ -1177,7 +1177,7 @@ class FederalIncumbentProtectionTestcase(sas_testcase.SasTestCase):
               config['dpaDatabaseConfig']['fileUrl']
       })
 
-    if 'runEarlyCpas' in config and config['runEarlyCpas']:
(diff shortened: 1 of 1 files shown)


src/harness/testcases/WINNF_FT_S_IPR_testcase.py
```diff
@@ -1098,6 +1098,7 @@ def generate_IPR_7_default_config(self, filename):
         'sasTestHarnessConfigs': [sas_test_harness_config],
         'domainProxies': [domain_proxy],
         'portalDpa': portal_dpa,
+        'runEarlyCpas': False
```
jcereced: Should we make the default True?  I run it both ways and it work and the impact (time) is minimal if the flag is set to True.  I could not tell from the recommended WG4 changes what the plan was for ITS to do first

kate-harrison: As far as I know, there is no WG4 preference. It was a bit cleaner to code it this way:

```if flag exists and flag=True...```

vs.

```if flag doesn't exist or (flag exists and flag=True)...```

hence I went with the first one. (If I'm honest, it's that plus personal preference for the default to be 'false.')

kate-harrison:
Thank you!

PR #795, opened 2018-08-07
Use the right grant request & maxEirp in the interference calculation.

harinineelu:
Use the right grant request (and corresponding maxEirp) in the interference calculation. The existing code uses the wrong index into the grant requests array and hence there is a mismatch between the EIRP used in the SAS calculation vs. the EIRP used in the test harness calculation.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_BPR_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_BPR_testcase.py
@@ -285,7 +285,7 @@ class BorderProtectionTestcase(sas_testcase.SasTestCase):
       logging.info('Closest point in the border: Lat is %f', closest_point_lat)
       logging.info('Closest point in the border: Long is %f', closest_point_lon)
       # requested_eirp (p)
-      p = config['grantRequests'][i]['operationParam']['maxEirp']
+      p = config['grantRequests'][index]['operationParam']['maxEirp']
       # Calculate PL
       cbsd_height = cbsd_information['installationParam']['height']
       cbsd_height_type = cbsd_information['installationParam']['heightType']


RedaHossino:
I think we should do the same change for the lines(we don't need to use the "i" variable of this loop):

254-       logging.info('Looking at Grant response number: %d', i)

323-       self.assertTrue(responses[i]['response']['responseCode'] == 400)

right?

kate-harrison:
I think that the current code is correct, although I can see an argument that using 'i' instead of 'index' on line 254 is misleading. We definitely need to use 'i' for line 323, though.

The problem is that only some of the registration requests are approved. Let's take the scenario where we have four registration requests: 0 and 2 are rejected, 1 and 3 are accepted. Now `index_of_devices_with_success_registration_response` is equal to [1, 3]. So far, so easy.

To be a valid config, `config['grantRequests']` needed to have four grant requests -- just in case all of the registrations were accepted. Moreover, the Nth grant request is meant to go with the Nth registration request. However we only want to send the requests corresponding to the devices which were actually registered, so we pull those out into `grant_request`. So `len(config['grantRequests']) = 4` but `len(grant_request) = 2` and they're using two different indexing schemes.

In line 288, it's correct to use either of `config['grantRequests'][index]` or `grant_request[i]`, and we went with the first one because it was slightly simpler. In line 323, you're indexing into the grant response which uses the same indexing scheme as `grant_request` and therefore `i` is the correct choice. Do you agree?

I'm OK to change line 254 if you really prefer, but I think `i` is slightly better here since it's again indexing into the grant response which corresponds to `grant_request`, not `config['grantRequests']`. (But on the other hand, debugging is easier if you can just index into the config to find the offending grant request rather than having to first figure out which registration requests were successful.) Let me know what you prefer, given the explanation above.

redtechnologies:
Kate, thank you for this detailed explanation, I agree with your points, and I think it is better to let line 254  as it is.

kate-harrison:
Thank you!

PR #802, opened 2018-09-05
Fix DPA manager to have coherent state always.

sbdt-google:
Wen one does not call ComputeMoveLists(), the internal state of DPA manager is not coherent.
MCP type harness test only perform CheckInterference when no peer SAS causes a test crash.

The fix is to :
  - always have a coherent list of neighbor_list and move_list with the number of channels, and do not depend on ComputeMoveLIst being called
  - make sure the self.channels is a private variable (renaming it to self._channels) as client code shall not manipulate this directly.


sbdt-google:
Thanks.

PR #805, opened 2018-09-13
Fix miscellaneous bugs in domain proxy

TaranFurnell:
- Fix call to _relinquishmentRequestWithMaximumBatchSize to use self (without it the code will fail if a grant is relinquished on heartbeat).
- Remove incorrect error message when a heartbeat terminates a grant but does not provide operationParam. This is a valid response as a SAS is not required to provide operationParam, the current error message suggests such a response is invalid.


kate-harrison:
Approving on behalf of Sony, who approved via email earlier today. Thanks!

PR #807, opened 2018-09-14
Domain proxy batching of registration and heartbeat

TaranFurnell:
Batch domain proxy registration and heartbeat requests with maximum batch size.

### DIFF
--- a/src/harness/test_harness_objects.py
+++ b/src/harness/test_harness_objects.py
@@ -168,11 +168,8 @@ class DomainProxy(object):
     # There should be exactly one grant request per registration request.
     self.testcase.assertEqual(len(grant_requests), len(registration_requests))
     try:
-      cbsd_ids = self.testcase.assertRegistered(
-          registration_requests,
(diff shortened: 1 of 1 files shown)


src/harness/test_harness_objects.py
```diff
@@ -356,6 +349,56 @@ def getCbsdsWithAtLeastOneAuthorizedGrant(self):
...
+      conditional_registration_data: Optional, any conditional registration data
+        for the CBSDs.
+    Returns:
+      The cbsd_ids from each registration request.
+    Raises:
+      AssertionError: If any registration responses are invalid.
+    """
+    # Make sure FccId and userId are injected before conditionals. The
+    # optimization in the SAS interface will prevent duplicate requests from
+    # being sent.
```
RedaHossino: the code looks good, this comment was not that clear to me(namely the text "in the SAS interface" is a bit confusing, because it is not the interface which forces for this optimization, but it is internally in the class "SasAdminImpl" in "sas.py" ),  so I think  this comment could be more clear.

PR #808, opened 2018-09-17
Fix FSS purge list: MCBSD should be for the 5MHz target band.

sbdt-google:
In Purge list, the threshold is established according to a 5MHz band at 3700-3705.
However the generated interference in that band is counted only for a 1MHz bandwidth
(since the MCBSD is for 1 MHz reference). This PR fixes the MCBSD to count for
the full 5MHz channel.


sbdt-google:
If you agree, can you approve so that I submit in time for V&V ? Thanks

Issue #799, opened 2018-08-16
test_WINNF_FT_S_FPR: fccIds used in SAS-THs are not whitelisted in SAS UUT before getting FAD from SAS-THs to SAS UUT


chandanahs:
test_WINNF_FT_S_FPR tests: 
fccIds used in SAS-THs are not whitelisted in SAS UUT before getting FAD from SAS-THs to SAS UUT

    # SAS Test Harnesses configurations,
    # Following configurations are for two SAS test harnesses
    sas_test_harness_device_1 = json.load(
        open(os.path.join('testcases', 'testdata', 'device_a.json')))
    sas_test_harness_device_1['fccId'] = 'test_fcc_id_1'
    sas_test_harness_device_1['userId'] = 'test_user_id_1'

    sas_test_harness_device_2 = json.load(
        open(os.path.join('testcases', 'testdata', 'device_b.json')))
    sas_test_harness_device_2['fccId'] = 'test_fcc_id_2'
    sas_test_harness_device_2['userId'] = 'test_user_id_2'

    sas_test_harness_device_3 = json.load(
        open(os.path.join('testcases', 'testdata', 'device_c.json')))
    sas_test_harness_device_3['fccId'] = 'test_fcc_id_3'
    sas_test_harness_device_3['userId'] = 'test_user_id_3'

    # Generate Cbsd FAD Records for SAS Test Harness 0
    cbsd_fad_records_sas_test_harness_0 = generateCbsdRecords(
        [sas_test_harness_device_1], [[grant_request_1]])

    # Generate Cbsd FAD Records for SAS Test Harness 1
    cbsd_fad_records_sas_test_harness_1 = generateCbsdRecords(
        [sas_test_harness_device_2, sas_test_harness_device_3],
        [[grant_request_4], [grant_request_6]])

    # Generate SAS Test Harnesses dump records
    dump_records_sas_test_harness_0 = {
        'cbsdRecords': cbsd_fad_records_sas_test_harness_0
    }
    dump_records_sas_test_harness_1 = {
        'cbsdRecords': cbsd_fad_records_sas_test_harness_1
    }

=====================
Is it not necessary to whitelist fccIds used in SAS THs ('test_fcc_id_1', 'test_fcc_id_2' and 'test_fcc_id_3') in SAS UUT before "logging.info('Collecting SAS TH FADs.')" and "logging.info('Step 10: Triggering CPAS.')" in executeSingleMCPIteration?

[...]

PR #810, opened 2018-09-20
Whitelist fccIds by injecting before FAD

arhannan:
Change based on Issue 799
https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues/799#issuecomment-421357895

### DIFF
--- a/src/harness/sas_testcase.py
+++ b/src/harness/sas_testcase.py
@@ -515,3 +515,8 @@ class SasTestCase(unittest.TestCase):
       if 'shutdown' in dir(thread):
         logging.info('Stopping %s' % thread.name)
         thread.shutdown()
+
+  def InjectTestHarnessFccIds(self, cbsd_records):
+    logging.info('Injecting FCC IDs for CBSDs in the SAS test harness into the SAS UUT.')
+    for cbsd_record in cbsd_records:
+        self._sas_admin.InjectFccId({'fccId': cbsd_record['registration']['fccId']}) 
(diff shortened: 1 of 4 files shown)


arhannan:
We can make the change inside sas_test_harness.writeFadRecord. That however will require calling sas_admin from sas_test_harness. Either we have to setup the Admin API in sas_test_harness or pass it into sas_test_harness. Won’t it be easier to call sas_admin.InjectFccId whenever writeFadRecord is called? 
I.e make changes in these files WINNF_FT_S_FAD_testcase.py WINNF_FT_S_GRA_testcase.py WINNF_FT_S_IPR_testcase.py WINNF_FT_S_MCP_testcase.py

kate-harrison:
You have a good point that the Admin API is not currently accessible to the SAS Test Harness. I think it should be relatively simple to add it, but I agree that it could also be quite confusing since the SAS TH is not logically connected to the SAS UUT Admin API.

My concern is that the code proposed here will have to be repeated in a number of places (e.g. there are 7 IPR tests, each with the potential for SAS THs) and that leaves a lot of opportunity for error -- especially because the current implementation is a little more complicated than necessary. I would feel more comfortable with your proposal if we could at least encapsulate the code here into a function (living in sas_testcase.py) which can be called from each of the necessary locations:

```
def InjectTestHarnessFccIds(self, cbsd_records):
  logging.info('Injecting FCC IDs for CBSDs in the SAS test harness into the SAS UUT.')
  for cbsd_record in cbsd_records:
    self._sas_admin.InjectFccId({'fccId': cbsd_record['registration']['fccId']})
```

then e.g. the code in MCP looks like:

```
...
self.InjectTestHarnessFccIds(test_harness_data['cbsdRecords'])
test_harness.writeFadRecords([test_harness_data['cbsdRecords']]) # no change here
...
```
and it would be a similarly small change to the other test cases. What do you think about this?

arhannan:
I agree with you. Your proposal is a good idea. We'll make the changes

kate-harrison:
Sounds great, thank you! I'll be OOO soon, so Taran from Google will make sure that your PR is reviewed quickly if I'm unable to do so myself.

arhannan:
Thanks @kate-harrison . We have checked in the new changes. We modified MCP, GRA and FAD.

PR #811, opened 2018-09-20
Update MCP testcase to ignore failed registrations

kate-harrison:
ITS has requested this change to account for differences in SAS behavior during registration. The change allows the test harness to ignore registration failures and continue test execution, whereas previously a registration failure would immediately cause a test case failure. Registration failures will be logged so that unexpected or widespread registration failures can be detected. This change is consistent with the test spec.


kate-harrison:
Thank you! (And don't worry, I saw the Pantheon today :-) )

PR #812, opened 2018-09-21
Introduce a flag to turn on/off early-exit in PAT

kate-harrison:
The current test code causes the PAT test to fail as soon as the failure threshold is reached. While this can save time if you want only a pass/fail result for a SAS, it makes debugging difficult: the debugger typically wants to get the maximum amount of data from each test run, and in that case the early-exit feature is counterproductive. ITS recently noted the drawbacks of our overly-eager optimization, so this PR introduces a flag which lets them toggle between the two modes (with the default being to disable early-exit).


kate-harrison:
Thank you!

PR #813, opened 2018-09-25
Update FDB 5 and 6 expiration dates to 2028-09-23

TaranFurnell:
Currently the FDB.5 and FDB.6 default GWBLs have expired. This PR moves the expiration date into the future by a decade (2028-09-23). By that point this testcase should well and truly be deprecated, and if not the test will also print a log to make it easier on anyone trying to debug why this test could be failing.


src/harness/testcases/WINNF_FT_S_FDB_testcase.py
```diff
@@ -1465,3 +1469,10 @@ def test_WINNF_FT_S_FDB_8(self, config_filename):
     # Check the heartbeat response code is 500(TERMINATED_GRANT)
     for resp in heartbeat_responses:
       self.assertEqual(resp['response']['responseCode'], 500)
+
+  def check_default_test_not_expired(self):
+    """Check the expiration date in the default configs has not been reached."""
+    if datetime.now() > datetime.strptime('2028-09-23', '%Y-%m-%d') - timedelta(days=1):
+      logging.error('Test is being run near or past the expiration date of the '
```
sushantparkhi: How about putting this as `logging.warn` instead of `error`?

jni2000:
approved, thanks.

src/harness/testcases/WINNF_FT_S_FDB_testcase.py
TaranFurnell: If this condition is ever true it's a very strong indication that you should likely not be running this test at all anymore hence the error level logging.

PR #817, opened 2018-10-02
DRG.6 remove registered CBSDs from second preload conditional data

RedaHossino:
Update DRG 6 so the test doesn't send preload conditional data for registrated CBSD(not successfully deregistered CBSDs)

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_DRG_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_DRG_testcase.py
@@ -429,8 +429,8 @@ class DeregistrationTestcase(sas_testcase.SasTestCase):
       if response['response']['responseCode'] != 0:
         cbsd_to_be_removed_from_next_reg = registration_request[i]
         indexes_conditional_to_remove = [index for index, c in enumerate(conditional_registration_data) if c['cbsdSerialNumber'] == cbsd_to_be_removed_from_next_reg['cbsdSerialNumber'] and c['fccId'] == cbsd_to_be_removed_from_next_reg['fccId']]
-        if indexes_conditional_to_remove:
-          del conditional_registration_data[indexes_conditional_to_remove[0]]
+        for index in reversed(indexes_conditional_to_remove):
+          del conditional_registration_data[index]
     # Step 6: Send the Deregistration request from Step 5 again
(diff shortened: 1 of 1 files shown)


src/harness/testcases/WINNF_FT_S_DRG_testcase.py
```diff
@@ -423,7 +425,12 @@ def test_WINNF_FT_S_DRG_6(self, config_filename):
           self.assertEqual(response['cbsdId'], deregister_request[i]['cbsdId'])
         else:
           self.assertFalse('cbsdId' in response)
-
+      # remove still registered CBSDs from the next preload conditional registration
+      if response['response']['responseCode'] != 0:
+        cbsd_to_be_removed_from_next_reg = registration_request[i]
+        indexes_conditional_to_remove = [index for index, c in enumerate(conditional_registration_data) if c['cbsdSerialNumber'] == cbsd_to_be_removed_from_next_reg['cbsdSerialNumber'] and c['fccId'] == cbsd_to_be_removed_from_next_reg['fccId']]
+        if indexes_conditional_to_remove:
+          del conditional_registration_data[indexes_conditional_to_remove[0]]
```
kate-harrison: If for some reason the config file has multiple conditional reg entries for the same (FCC ID, SN) pair, this will only remove the first. You probably want something like

```
for index in reversed(indexes_conditional_to_remove):
  del conditional_registration_data[index]
```
Please double-check the logic though to make sure that it works correctly. Note specifically the reversal of the list (taking advantage of the fact that it's already sorted in ascending order) so that we don't change the meaning of the index while using it.

RedaHossino:
thank you for catching this case, I did the suggested changes and I tested it.
on another hand, I don't know if it will be a "correct" config file if it  contains twice or more the same preloaded conditional data for the same CBSD.
I cann't think of another reason except a mistake in filling the config file for why this case might happen, right?
if yes then this fix may  hide this kind of mistakes.

kate-harrison:
Hm, that's a good point: I agree that that kind of config would be invalid. In that case, it's probably better to catch the problem with a separate tool since even checking that `len(indexes_conditional_to_remove) <= 1` will not catch all cases of that kind of bad config (and a larger change is probably too much at this point).

PR #819, opened 2018-10-04
EHata: Fix Fine Rolling Hill correction.

sbdt-google:
When the rolling hill is below 10m, the correction should be zero.
Otherwise it does not follow the EHata spec on one hand, plus it
introduces some randomness and discontinuity in the correction of
2dB or 4dB when the terrain is very flat near the mobile (ie a
microscopic change of altitude can cause up to 4dB variation in
pathloss).

This change is make to reflect the proposed change in the ITS original
code base, which was integrated recently at:
NTIA/ehata@b54b71f

Further documentation on the root issue.
 + See Ticket on NTIA GitHub: https://github.com/NTIA/ehata/issues/7
 + Issue #333


krahnavardy:
Approved!

PR #820, opened 2018-10-04
EHata: Cleanup of the float -> double switch.

sbdt-google:
Previously to avoid having the code base differing too much from
the original ITS reference implementation, we defined a global
  #define float double
in the ehata.h header file.
Obviously this is very hack and can fail depending on the order of
inclusion of ehata.h, or if t is forgotten in one of the cpp file.

The proper way is now implemented by directly changing all `float`
occurence with `double`.
This is in line with the change made by ITS on their reference
implementation at:
https://github.com/NTIA/ehata/commit/df2b5802f1444b77ffadf4a4301c5b2fd7be691c


krahnavardy:
Approved!

sbdt-google:
Thanks.

PR #821, opened 2018-10-04
EHata: Cleanup Average terrain code for the NTIA compatibility mode.

sbdt-google:
Note: no effect on the code used by test harness, this is a cleanup.

The compatibility mode (option WinnforumExtension=False) replicating 
ITS implementation has been upgraded to latest version found in NTIA 
reference version (as was done in NTIA/ehata@80954b1).

This has no effect on test harness which used WinnforumExtensions=True,
but removes confusion when comparing the Winnforum version versus the
NTIA version (like was reported recently by ITS). 

In some sense, it also documents precisely what variation we have now with 
the NTIA ref code: just the use of ceil() in this average terrain calculation (in addition
of the clipping of the effective height in ExtendedHata.cpp which differs as per the spec).


krahnavardy:
Approved

PR #826, opened 2018-10-10
Add additional debugging information to some of the protocol tests.

kate-harrison:
There are no functional changes in this PR. Instead, it proposes to collect information on multiple failures before halting the test, allowing for faster debugging in the presence of multiple errors. It also logs additional information which may be useful during debugging.


sbdt-google:
Approving on behalf of Sony (since they cannot use GitHub).

PR #829, opened 2018-10-17
Update DPAs KML from NTIA

sbdt-google:
Update DPAs KMLs:
  E-DPA to v1.07
  P-DPA to v1.06

As provided by Nick from NTIA
The difference with previous DPAs are:
 - using the default value (instead of NaN) for catA distance (150km),
 and OOB CatB (25km) and OOB CatA (0km, ie deactivated)
 - portal DPA "Barking Sands" has now its CatB distance set properly
 - CatA distance are set to not exceed CatB distances for:
    + West9: 144km
    + Alameda: 128km
    + Bremerton-Everett: 128km

Also put the PDF file describing the DPA KML format.


sushantparkhi:
Looks Good!

PR #830, opened 2018-10-17
Add more logging of A_DPA, A_DPA_ref, and keep lists to DPA manager

kate-harrison:
There are no functional changes in this PR. Instead, it adds logging for A_DPA and A_DPA_ref for every DPA point at every azimuth. It also adds logging of statistics on their difference.

### DIFF
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -38,6 +38,7 @@ Example usage:
   # Check the interference according to Winnforum IPR tests
   status = dpa.CheckInterference(sas_uut_keep_list, margin_db=2)
 """
+from collections import namedtuple
 import functools
 import os
 import logging
@@ -65,6 +66,20 @@ class _EmptyFad(object):
   """Helper class for representing an empty FAD."""
   def getCbsdRecords(self):
     return []
+
+class DpaInterferenceResult(namedtuple('DpaInterferenceResult',
+                                       ['max_difference',
+                                        'A_DPA',
+                                        'A_DPA_ref',
+                                        ])):
+  """Holds detailed information on a DPA aggregate interference result.
+
+  Attributes:
+    max_difference: maximum difference between A_DPA and A_DPA_ref.
+    A_DPA: SAS UUT's interference to the DPA point.
+    A_DPA_ref: reference value for SAS UUT's interference to the DPA point.
+  """
+  __slots__ = ()
 
 
 # The main Dpa class representing a Dpa zone
@@ -358,11 +373,12 @@ class Dpa(object):
       (ie if at least one combined protection point / azimuth fails the test).
     """
     if channel is None:
+      test_passed = True
       for chan in self._channels:
         if not self.CheckInterference(sas_uut_active_grants, margin_db, chan, num_iter,
                                       do_abs_check_single_uut):
-          return False
-      return True
+          test_passed = False
+      return test_passed
 
     if num_iter is None:
       num_iter = Dpa.num_iteration
@@ -416,12 +432,13 @@ class Dpa(object):
         neighbor_distances=self.neighbor_distances,
         threshold=hard_threshold
     )
-    # TODO(sbdt): could do early stop as soon as one fails, although I would expect
-    # the criteria to be changed into checking 99.x% of success instead of 100%.
     pool = mpool.Pool()
     result = pool.map(checkPointInterf, self.protected_points)
-    max_diff_interf = max(result)
 
+    logging.info('Printing stats for DPA %s, channel %s', self.name, channel)
+    self.__PrintStatistics(result, margin_db)
+
+    max_diff_interf = max([r.max_difference for r in result])
     if max_diff_interf > margin_db:
       logging.warning('DPA Check Fail `%s`- channel %s thresh %s max_diff %s',
                       self.name, channel,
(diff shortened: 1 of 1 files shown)


kate-harrison:
A sample of the new output file format is here (remove .txt extension before opening): [2018-10-18 11:54:06 DPA=BATH channel=(3620, 3630).csv.txt](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/2493243/2018-10-18.11.54.06.DPA.BATH.channel.3620.3630.csv.txt)

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -432,6 +448,39 @@ def CheckInterference(self, sas_uut_active_grants, margin_db,
 
     return max_diff_interf <= margin_db
 
+  def __PrintStatistics(self, results, margin_db, dpa_name, channel):
+    timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+    filename='%s DPA=%s channel=%s.csv' % (timestamp, self.name, channel)
+    logging.info('Saving stats for DPA %s, channel %s to file: %s', dpa_name, channel, filename)
+    with open(filename, 'w') as f:
+      # CSV header.
+      f.write('Latitude,Longitude,Azimuth (degrees),A_DPA (dBm),A_DPA_ref (dBm),A_DPA - A_DPA_ref,A_DPA + 144 dBm\n')
```
redtechnologies: May be we could replace 144 by the value from the kml if available (self.threshold?)

kate-harrison: This is what ITS suggested so I would prefer to keep it as is. Perhaps you can ask Joel from ITS if he would prefer your approach?

kate-harrison:
I just added the ability to output the keep lists, as requested by ITS. (It was such a small change, it was easier to add it to the existing PR than to create a new PR and deal with merge conflicts later.)

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -416,12 +435,12 @@ def CheckInterference(self, sas_uut_active_grants, margin_db,
...
     )
-    # TODO(sbdt): could do early stop as soon as one fails, although I would expect
-    # the criteria to be changed into checking 99.x% of success instead of 100%.
     pool = mpool.Pool()
     result = pool.map(checkPointInterf, self.protected_points)
-    max_diff_interf = max(result)
 
+    self.__PrintStatistics(result, margin_db, self.name, channel)
+
+    max_diff_interf = max([r.max_difference for r in result])
```
sbdt-google: use : max_diff_interf = max(r.max_difference for r in result)

This will avoid creation of the list, as it will use iterator.

src/harness/reference_models/dpa/dpa_mgr.py
sbdt-google: It is true it could create confusion if a test is done on one of the DPA with a different threshold, then they will see big negative number which leads to the belief the test is badly failed.
The title could be named: "A_DPA+Prot_thresh" and use self.threshold in the addition below

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -432,6 +451,71 @@ def CheckInterference(self, sas_uut_active_grants, margin_db,
...
+          'latitude', 'longitude', 'height_agl', 'indoor_deployment',
+          'cbsd_category', 'antenna_azimuth', 'antenna_gain',
+          'antenna_beamwidth', 'max_eirp', 'low_frequency', 'high_frequency',
+          'is_managed_grant'
+      ]
+      with open(filename, 'w') as f:
+        f.write(','.join(fields) + '\n')
+        for cbsd_grant_info in keep_list:
+          cbsd_grant_dict = cbsd_grant_info._asdict()
+          f.write(','.join([str(cbsd_grant_dict[key]) for key in fields]) + '\n')
```
sbdt-google: If you want you can get as following:
```
for cbsd_grant_info  in ...:
  f.write(','.join(str(getattr(cbsd_grant_info, key)) for key in fields) + '\n')
```
to avoid the conversion to dict and temp creation of a list

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -432,6 +451,70 @@ def CheckInterference(self, sas_uut_active_grants, margin_db,
 
     return max_diff_interf <= margin_db
 
+  def __PrintStatistics(self, results, margin_db, dpa_name, channel):
+    timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
+    filename='%s DPA=%s channel=%s.csv' % (timestamp, self.name, channel)
```
redtechnologies: Both "dpa_name" are "self.name" are used in the function. For consistency, it could  be better to use just one of them.

redtechnologies: Both "dpa_name" are "self.name" are used in the function. For consistency, it could  be better to use just one of them.

redtechnologies: Both  **dpa_name** are **self.name** are used in the function. For consistency, it could  be better to use just one of them.

redtechnologies: Both  **dpa_name** are **self.name** are used in the function. For consistency, it could  be better to use just one of them.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -432,6 +451,70 @@ def CheckInterference(self, sas_uut_active_grants, margin_db,
 
     return max_diff_interf <= margin_db
 
+  def __PrintStatistics(self, results, margin_db, dpa_name, channel):
+    timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
```
redtechnologies: we have an issue with files containing ":". 
Could we replace ':' by another character, for example "_"?

src/harness/reference_models/dpa/dpa_mgr.py
kate-harrison: Joel concurs, so I've changed the filename to be (e.g.) `2018-10-22 07_11_12 DPA=BATH channel=(3620, 3630) threshold=-144.0.csv` and the format fo the file to be

```
Latitude,Longitude,Azimuth (degrees),A_DPA (dBm),A_DPA_ref (dBm),A_DPA - A_DPA_ref,A_DPA - threshold
43.9055555600,-69.8138888900,0.0000000000,-179.7476485399,-144.0000000000,-35.7476485399,-35.7476485399
43.9055555600,-69.8138888900,1.5000000000,-179.7476485399,-144.0000000000,-35.7476485399,-35.7476485399
...
```

src/harness/reference_models/dpa/dpa_mgr.py
kate-harrison: Thanks for catching this, fixed!

src/harness/reference_models/dpa/dpa_mgr.py
kate-harrison: No problem, done.

kate-harrison:
I have addressed all review comments -- please take another look.

Javi/Navin, do you still plan to look at this before 9am Pacific?

nasrinivasa:
Kate, We looked at the code and it looks good. The Keep list that is being stored is not the per protection point Keep list right.

kate-harrison:
Thanks, Navin! Correct, the keep list being stored is the overall (union) keep list for all DPA points.

Can whoever is reviewing from RED please take a quick look and approve or give comments? I'd really like to include this change in this week's V&V.

redtechnologies:
All changes look good. I approve this PR. 
Thanks
Christophe

kate-harrison:
Thank you!!

kate-harrison:
It looks like you accidentally closed it rather than approve, though. I re-opened but it seems to have lost the approval. Can you please re-approve?

PR #833, opened 2018-10-25
Simulation only: DPA margin script & simulation utilities

sbdt-google:
A script demonstrating the test harness SUCCESS check issue related to margin.
Replicate latest test between Reda and Michael showing the fundamental issue: Failure with 12dB difference for a single point DPA with 1000 CBSDs.

The script can be extended or used as a starter for other type of simulation. Functions which are generic
to simulation have been put in the tools/sim_utils.py


sbdt-google:
Thanks for the review

PR #835, opened 2018-10-26
Batch conditional reg data

kate-harrison:
Without this change, the injection of conditional registration data does not follow the batching methodology we've established for other APIs.


kate-harrison:
Thank you!

Issue #834, opened 2018-10-26
Formatting Error in makePalRecordsConsistent


nkentNTIA:
Hey all,

The following currently occupies lines 208 and 209 of util.py:

```python
    pal_rec['license']['licenseAreaExtent'] = \
      'zone/census_tract/census/%d/%d' % (pal_census_year, pal_fips_code)
```

This produces an error if the user provides pal records that have string fips codes, which are necessary, as leading zeroes aren't allowed in Python ints or json numbers, and there are many census tracts for which the fips codes contain leading zeroes.

This could be fixed with the following: 

```python
    pal_rec['license']['licenseAreaExtent'] = \
      'zone/census_tract/census/{}/{}'.format(pal_census_year, pal_fips_code)
```

There are other pitfalls to using the old `%` syntax here, so i think format is preferable, but obviously this is up to the discretion of the TH dev team.

Cheers!

kate-harrison:
Thanks! I put your fix into #836 after double-checking that it worked.

nkentNTIA:
Thanks a bunch, Kate!

kate-harrison:
Fixed in #836.

PR #836, opened 2018-10-26
Fix error creating PAL record for FIPS codes with leading zeroes

kate-harrison:
Fixes issue #834.


kate-harrison:
FYI the easiest way to verify the fix is to modify pal_record_0.json to use `"fipsCode": "00001234"`, run GRA.10 (default config), and search for "1234" in the logs.

PR #837, opened 2018-10-27
DPA: Add new modes for DPA margin.

sbdt-google:
As discussed in latest sas-test-dev meeting, two modes are added:
  + 'target (margin_db)': using the absolute target + margin_db
  + 'linear (margin_db)': using a linear margin (in mW) computed
    from the equivalent margin_db at the threshold. Sucess if interference
    below ref_model_interference_mw + margin_mw

This corresponds respectively to option 1 and option 2. 
To configure these modes, simply replace the ''movelistMargin" tag in the test config, for example:  
   "movelistMargin" : "linear (1.5)"
 
This PR also fixed a few bug in the _PrintStatistics() routine:
  - was failing if an input threshold is integer (normally not the case in test harness, but possibly)
  - was failing in one condition (peer sas with no grants in neighborhood) as the test was using the
  wrong A_DPA.


kate-harrison:
Thank you! I ran it myself to double-check and it worked as expected.

PR #838, opened 2018-10-27
DPA interf check: Add support for 'Best-of-N' reference move list.

sbdt-google:
By default OFF. 
The computeMoveList can now optionally compute several
move list and take the best, best being either:
 -  the smallest one, ie the one with biggest keep list, and so producing the largest interference
hopefully.
 - the intersection one, ie the intersection of N move list

This change is fully integrated with the test harness (IPR/MCP tests)
via an optional config parameter 'bestOfN': positive value will use the smallest method and negative value withl use the intersection method.

Note that this proposal slow down the reference model by factor N.
The idea is to avoid "outliers" by using a small N, even something as small as 2 to 5.

### Explanation:
All previous DPA margin studies are demonstrating the checkInterference issue 
by comparing the SAS UUT creating a "best" move list (not a crazy assumption for an
efficient SAS), to the "worst case" move list reference model across many runs.
This can happen in practice in a random fashion. But if we repeat the
move list calculation a few time and take the smallest one, the chance
for it to happen is much lower. For example if we want to avoid the 10%
worst move list:
 - nominally we have 1 chance over 10 to fall in this scenario
 - by choosing the best of 3, the probability fall to 1 over 1000.

The intersection method is somewhat  is related to Michael Souryal "move list intersection" proposal which involves more move list calculation and the intersection of the XX% middle ones (in terms of move list length).

### DIFF
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -261,9 +261,6 @@ class Dpa(object):
         will be the best of those, ie the smallest one. If number is negative,
         intersection between the move list is performed instead.
     """
-    # TODO: add eventually the NTIA proposal which computes the intersection move list.
-    #       it is more involved as it requires more single move list calculation, but
-    #       likely produces a more stable 'composite' move list.
     logging.info('DPA Compute movelist `%s`- channels %s thresh %s bw %s height %s '
                  'iter %s azi_range %s nbor_dists %s best_of_n %s',
                  self.name, self._channels, self.threshold, self.beamwidth,
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -264,10 +264,10 @@ class Dpa(object):
     #       it is more involved as it requires more single move list calculation, but
     #       likely produces a more stable 'composite' move list.
     logging.info('DPA Compute movelist `%s`- channels %s thresh %s bw %s height %s '
-                 'iter %s azi_range %s nbor_dists %s',
+                 'iter %s azi_range %s nbor_dists %s best_of_n %s',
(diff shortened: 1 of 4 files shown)


msouryal:
(Copying my comment from the reflector.)
Keep list size is not a perfect predictor of aggregate interference; the content of the keep list matters.  Another approach would be to iterate the move-list computation and aggregate check, together, up to N times.  In IPR.7, this would entail executing Steps 10, 13, and the Check up to N times.  Each iteration generates a valid reference move list, and the SAS UUT would only need to pass one Check.  Not all N iterations would necessarily be needed; as soon as the SAS UUT passes, the test would end.

sbdt-google:
@msouryal @kate-harrison  

You are right that it is not the best predictor, so let's think more about this. The issue about doing the full check several times seems to be an integration issue within the test harness framework, according to Kate. This is why a pure standalone solution to come up with a single "better" move list would have been easier to integrate within the rest of test harness.

Perhaps going to your intersection method instead is the right way... Do you think it makes sense for a small number of move lists (let's say 10 and not 100) for the intersection to work well enough ??
I can easily upgrade to that methodology instead.

msouryal:
@sbdt-google @kate-harrison @msouryal 

We can at least say that the intersection of N move lists would yield no less aggregate interference than the smallest of N, and probably more.  The choice of N can be left to the tester, with the understanding that the larger N is, the larger the reference aggregate will be, in general.  But as you note above, the intersection move list is not necessarily a valid move list.

sbdt-google:
@sbdt-google @kate-harrison @msouryal

I am not sure if this PR is obsolete yet. I updated it with "intersection of N" capabilities.
Now if the 'BestOfN' flag is set with 
 - positive value: minimum of N
 - negative value: intersection of N

sbdt-google:
I think this PR is not to be included, so I will close it. It can be reinitiated in the future if required

PR #842, opened 2018-10-31
Improve portability of JSON schema checking code

### LINKED ISSUE #840: assertContainsRequiredFields breaks in Windows
Hey again,

In getting the TH set up on a Windows user's machine, a bug turned up that causes an unhandled exception when checking schema.

Line 270 in `assertContainsRequiredFields` currently reads as follows:

```python
resolver = RefResolver(referrer=schema, base_uri='file://' + schema_dir + '/')
```

The bug can be avoided with this guard:

```python
if os.name == "nt":
    os_base_uri = "file:///"
else:
    os_base_uri = "file:"

resolver = RefResolver(referrer=schema, base_uri=os_base_uri + schema_dir + "/")
```

Thanks again for your help!

kate-harrison:
Fixes #840.

### DIFF
--- a/src/harness/util.py
+++ b/src/harness/util.py
@@ -270,7 +270,7 @@ def assertContainsRequiredFields(schema_filename, response):
   if os.name == 'nt':
       os_base_uri = 'file:///'
   else:
-      os_base_uri = 'file:'
+      os_base_uri = 'file://'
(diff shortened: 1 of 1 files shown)


kate-harrison:
I tested on Linux -- the only OS I have available -- and it works as expected. I assume based on the Nick's comments in #840 that this change will work on Windows as well, but it would be ideal if someone could double-check.

PR #845, opened 2018-11-07
fix MCP injection of gwbl records

redtechnologies:


redtechnologies:
yes I think ir is the one type.
Thank you for the review!

PR #847, opened 2018-11-11
Fix mem Leak on ITM

sbdt-google:
Affects the calculations with array or reliabilities (like DPA),
and average computed value.

The mem leak is expected to be about 16kB for a DPA CBSD-Point link,
and 800 bytes for a CBSD-point link in other protection type.
It is thus quite bug for DPA and affect negatively any test of the test harness

### DIFF
--- a/src/harness/reference_models/propagation/itm/itm_its_py.cpp
+++ b/src/harness/reference_models/propagation/itm/itm_its_py.cpp
@@ -158,8 +158,9 @@ static PyObject* itm_point_to_point_rels(PyObject* self, PyObject* args) {
     PyList_SET_ITEM(loss_obj, k, Py_BuildValue("d", db_losses[k]));
   }
   delete[] db_losses;
-
-  return Py_BuildValue("Oddsi", loss_obj, ver0, ver1, strmode, errnum);
+  // Note: use N instead of O here to avoid increment of ref count of loss_obj.
+  // Otherwise would have a mem leak, since it would never be released.
+  return Py_BuildValue("Nddsi", loss_obj, ver0, ver1, strmode, errnum);
 }
 
 static PyMethodDef ITMMethods[] = {


sbdt-google:
Further explanation for reviewer:

As you may know, CPython use "reference counting" to manage its memory, meaning it keeps
internally a counter on each object, that is being incremented each time a new variable reference that object, and decremented when variable goes out of scope. With the goal that when the counter goes to zero, then the object is simply erased and the memory reclaimed by the system (since it is not referenced anymore by any variable).

All this is done automatically by the Python interpreter. This issue comes when doing C extension modules then you need to be careful to properly manage the reference count of new objects that you create.
In our case here, we want to create a list of float values to hold all the path loss for each required reliability.
We create a new list with:
`   PyObject* loss_obj = PyList_New(num_rels);
`
The reference count is set to 1 by this function.
We fill up the list with newly create Python float using the 
`    PyList_SET_ITEM(loss_obj, k, Py_BuildValue("d", db_losses[k]));
`
Note that Py_BuildValue creates the float object and set the ref count to 1. PyList_SetItem does not
increment the reference counting.
Finally we want to return this to the caller in a tuple.
We build the tuple again wiuth Py_BuildValue:
`  return Py_BuildValue("Oddsi", loss_obj, ver0, ver1, strmode, errnum);
`
The issue here is that using the 'O' qualifier means that the first element of the tuple is the newly created list, and its reference counting is incremented. So now we have it at value 2. 
Let's imagine the calling function to do:
```

def test():
   path_losses, angle0, angle1, _, _ = point_to_point_rels(...)
 
```
the variable path_losses has reference count 2.
When we go out of scope of function test(), the reference count goes to 1. This object in memory
is no more referenced by anyone but its memory has never been reclaimed since it did not go to zero.
Then we have the memory leak, which ultimately will take down the test server if the test is too big
(too many CBSD and/or protection entities).

The solution is to make sure that the reference count of the the returned lis is one and not two.
Using the identifier "N" instead of "O" in the Py_BuildValue() avoids the increment of the reference counting as explained in [this link.](https://docs.python.org/2/c-api/arg.html#c.Py_BuildValue)

arhannan:
Thanks for finding the memory and the detailed explanation for the fix.

sbdt-google:
Thanks for the review in time for the V&V start.

PR #853, opened 2018-11-14
DPA margin simulator: few improvements

sbdt-google:
Mostly on the reported statistics and histogram (log-density).
And ability to save simulation data in pickled file for replay later on, or more detailed analysis.


sbdt-google:
To Reda: last PR I think. Will not submit until absolutely sure of that, but please approve if you can.

sbdt-google:
Thanks Reda. I will let it open a bit more, so as not to bother you if I still need a few adjustments.

PR #854, opened 2018-11-14
DPA: improve logs for enabling post processing eval

sbdt-google:
Some logs were introduced in DPA CheckInterference for better post analysis of result in #830.

This PR improves this logging according to various report received since then:
 
 - output of CSV logs put into harness/testcases/output/
 - output the neighbor list in addition of the keep list, to allow full post processing analysis (for example redo random movelist with ref model and anlayse agg interference variation)
 - fix issues with SAS UUT keep list to be the keep list per channel and not all SAS active grants in whole US for all channels. This means we run the active grants through a neighbor filter, just used for that. 
 - main test harness script can now be run in debug mode by adding option: --log_level=debug

### DIFF
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -79,10 +79,7 @@ def Db2Lin(x):
   return 10**(x / 10.)
 
 def Lin2Db(x):
-  if not np.isscalar(x):
-    return 10 * np.log10(x.clip(min=1e-100))
-  return 10 * np.log10(max(x, 1e-100))
-
+  return 10 * np.log10(np.asarray(x).clip(min=1e-100))
 
 class DpaInterferenceResult(
     namedtuple('DpaInterferenceResult',
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -62,6 +62,13 @@ DPA_DEFAULT_DISTANCES = (150, 200, 0, 25)
 # The channel bandwidth
 DPA_CHANNEL_BANDWIDTH = 10
 
+# The logging path
+def GetDpaLogDir():
+  dpa_log_dir = os.path.join(os.path.dirname(__file__),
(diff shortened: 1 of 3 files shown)


sbdt-google:
Reda, can you have a look since you reported the Keep List being too big.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -72,7 +79,7 @@ def Db2Lin(x):
   return 10**(x / 10.)
 
 def Lin2Db(x):
-  return 10 * np.log10(x)
+  return 10 * np.log10(x.clip(min=1e-100))
```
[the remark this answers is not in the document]
sbdt-google: You are right. Let me change it.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -72,7 +79,9 @@ def Db2Lin(x):
   return 10**(x / 10.)
 
 def Lin2Db(x):
-  return 10 * np.log10(x)
+  if not np.isscalar(x):
+    return 10 * np.log10(x.clip(min=1e-100))
```
RedaHossino: could you check that this line work correctly when x is a list and not an np array?

sbdt-google: I dont think we use this Lin2Db for a list in this file, and this function is not designed to work with any Python sequence (list, dict, set,), only with  numerical container: (scalar and ndarray) as used in the module. 
But since it is easy to extend it, please check the new form.

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -79,10 +79,7 @@ def Db2Lin(x):
   return 10**(x / 10.)
 
 def Lin2Db(x):
-  if not np.isscalar(x):
-    return 10 * np.log10(x.clip(min=1e-100))
-  return 10 * np.log10(max(x, 1e-100))
-
+  return 10 * np.log10(np.asarray(x).clip(min=1e-100))
```
RedaHossino: looks good, thanks!

Issue #814, opened 2018-10-01
Small holes in the created polygon with 2 adjacent census tracts, causes PCR_1 fail


chandanahs:
when the service area is created in PCR_1 using 2 PALs with adjacent census tracts (pal_record_a['fipsCode'] = 20063955100 & pal_record_b['fipsCode'] = 20063955200), created Polygon after cascading is not the outer polygon, but has very small gap between census tracts.

When adjacent sides of adjacent census tracts are not having the exact same co-ordinates, we see very small gap which is considered as holes and PPA created during PCR_1 is failing with "AssertionError: PPA Zone is not within service area" as there is tiny hole in cascaded polygon using 2 given PALs with adjacent census tracts.

the similar scenario is discussed here: [https://gis.stackexchange.com/questions/277334/shapely-polygon-union-results-in-strange-artifacts-of-tiny-non-overlapping-area](url)

So Pls check if we have to change the test code as well as reference model code to create a outer polygon without any tiny holes!

Have a look at the attachments.

[raw_map.zip](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/2435248/raw_map.zip)

![adjacentcensustractscascadedpolygonwithtinyholes](https://user-images.githubusercontent.com/42067597/46309897-c15ca380-c58b-11e8-8725-ef724d9a742c.PNG)

chandanahs:
@kate-harrison : Could you have a look? or redirect to the right person?

chandanahs:
i have observed holes with less than 6 meter width holes between 2 adjacent census tracts.

attaching the screen shot for the same. Yes, Buffering would solve this problem.

![tinyholebtw2censustracts](https://user-images.githubusercontent.com/42067597/46692904-66076280-cbd6-11e8-83ae-89523aa433fd.PNG)

def isPpaWithinServiceArea(pal_records, ppa_zone_geometry):
   census_tracts_for_pal = [
      utils.ToShapely(drive.census_tract_driver.GetCensusTract(
          pal['license']['licenseAreaIdentifier'])
                      ['features'][0]['geometry']) for pal in pal_records]
  pal_service_area = ops.cascaded_union(census_tracts_for_pal)

  **pal_service_area = geometry.Polygon(pal_service_area.exterior)
[...]

PR #856, opened 2018-11-15
Remove obsolete census file.

sbdt-google:
This file is obsolete, was here to support default config in PCR and WDB while the SAS-Data database was being built. And should have been removed once that one became active.

Because it stayed there, it now creates confusion such as expressed in issue #814, since it was generated by an old process with some defects.


sbdt-google:
Thanks

PR #857, opened 2018-11-15
for FAD_1, consider lat and lng equal if same to 6 decimal places ( Fixes #822 )

### LINKED ISSUE #822: util.getRandomLatLongInPolygon returns lat long with 14 digits after decimal, But SAS-CBSD requirement says 6 digits after decimal
getRandomLatLongInPolygon returns lat long with 14 digits after decimal.

SAS-CBSD requirement says "Latitude of the CBSD antenna location in degrees relative to the WGS 84 datum [n.11]. The allowed range is from -90.000000 to +90.000000. Positive values represent latitudes north of the equator; negative values south of the equator. Values are specified using 6 digits to the right of the decimal point."

And FAD_1 test fails as lat/long sent in request (38.77693566715789/-97.20361504249222) are not equal to the registered CBSD coordinates (38.776936/-97.203615)

======================================================================
FAIL: test_WINNF_FT_S_FAD_1_0_default (WINNF_FT_S_FAD_testcase.FullActivityDumpTestcase)
This test verifies that a SAS UUT can successfully respond to a full
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/sasuser/code/Regression/Spectrum-Access-System/src/harness/util.py", line 79, in _func
    return func(*a, config_filename=config)
  File "/home/sasuser/code/Regression/Spectrum-Access-System/src/harness/testcases/WINNF_FT_S_FAD_testcase.py", line 478, in test_WINNF_FT_S_FAD_1
    self.assertCbsdRecord(config['registrationRequests'], grants, grant_responses, cbsd_dump_data, config['conditionalRegistrationData'])
  File "/home/sasuser/code/Regression/Spectrum-Access-System/src/harness/testcases/WINNF_FT_S_FAD_testcase.py", line 105, in assertCbsdRecord
    ['installationParam'], cbsd_record[0]['installationParam'])
  File "/home/sasuser/code/Regression/Spectrum-Access-System/src/harness/testcases/WINNF_FT_S_FAD_testcase.py", line 59, in assertEqualToDeviceOrPreloadedConditionalParam
    self.assertEqual(attr_value, record[attr_name])
AssertionError: 38.77693566715789 != 38.776936

Should we change the function to limit the number of digits to 6 post decimal point in getRandomLatLongInPolygon function in util.py?

nirajankeybridge:
Fixes https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues/822

Adhering to [SAS-CBSD specification WINNF-TS-0016, Version V1.2.2, 1 October 2018](https://workspace.winnforum.org/higherlogic/ws/public/download/7279/WINNF-TS-0016-V1.2.2%20SAS%20to%20CBSD%20Technical%20Specification.pdf) which says:
"Latitude of the CBSD antenna location in degrees relative to the WGS 84 datum [n.11]. The allowed range is from -90.000000 to +90.000000. Positive values represent latitudes north of the equator; negative values south of the equator. Values are specified using 6 digits to the right of the decimal point."

This should not affect the latitude or longitude precision elsewhere.

### DIFF
--- a/src/harness/testcases/WINNF_FT_S_FAD_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_FAD_testcase.py
@@ -57,7 +57,7 @@ class FullActivityDumpTestcase(sas_testcase.SasTestCase):
       attr_value = registration_request[attr_name] if attr_name in registration_request\
         else preloaded_conditionals[attr_name]
       if attr_name in ['latitude', 'longitude']:      
-        self.assertEqual(abs(attr_value-record[attr_name])<1e-6,True)
+        self.assertTrue(abs(attr_value-record[attr_name])<1e-6)
       else:
         self.assertEqual(attr_value, record[attr_name])
 
--- a/src/harness/testcases/WINNF_FT_S_FAD_testcase.py
+++ b/src/harness/testcases/WINNF_FT_S_FAD_testcase.py
@@ -56,7 +56,10 @@ class FullActivityDumpTestcase(sas_testcase.SasTestCase):
       """
       attr_value = registration_request[attr_name] if attr_name in registration_request\
         else preloaded_conditionals[attr_name]
-      self.assertEqual(attr_value, record[attr_name])
+      if attr_name in ['latitude', 'longitude']:      
+        self.assertEqual(abs(attr_value-record[attr_name])<1e-6,True)
+      else:
+        self.assertEqual(attr_value, record[attr_name])
 
 
   def assertCbsdRecord(self, registration_request, grant_request, grant_response, cbsd_dump_data, reg_conditional_data):


src/harness/testcases/WINNF_FT_S_FAD_testcase.py
```diff
@@ -56,7 +56,10 @@ def assertEqualToDeviceOrPreloadedConditionalParam(self, attr_name, registration
       """
       attr_value = registration_request[attr_name] if attr_name in registration_request\
         else preloaded_conditionals[attr_name]
-      self.assertEqual(attr_value, record[attr_name])
+      if attr_name in ['latitude', 'longitude']:      
+        self.assertEqual(abs(attr_value-record[attr_name])<1e-6,True)
```
kate-harrison: It's better to use assertTrue() instead of assertEqual() in this case (the error message is easier to understand), but this technically works as well.

nirajankeybridge: Agreed. I've switched to assertTrue(). Thanks Kate.

PR #852, opened 2018-11-13
PCR - add buffering to remove gaps between census tracts ( fixes #814 )

nirajankeybridge:
Adds buffering to remove gaps between census tracts for PCR
Fixes https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/issues/814
Based on work by @chandanahs

PR #858, opened 2018-11-15
Revert "PCR - add buffering to remove gaps between census tracts ( fixes #814 )"

### LINKED ISSUE #814: Small holes in the created polygon with 2 adjacent census tracts, causes PCR_1 fail

sbdt-google:
Reverts Wireless-Innovation-Forum/Spectrum-Access-System#852

As decided in the sas-test-dev group, this PR reverts the PR handling the issue #814 .
See email thread in sas_test_dev mail list.


kate-harrison:
I confirm that this matches what was decided in today's test dev meeting. Thanks!

PR #859, opened 2018-11-16
update if PPA cbsd_reference_id is empty array in sas_test_harness.py generatePpaRecords()

nirajankeybridge:
FAD.2 generates default.config with empty "cbsdReferenceId": [].

According to [WINNF-TS-0096 v1.3.0 SAS-SAS Protocol spec](https://workspace.winnforum.org/higherlogic/ws/public/document?document_id=6482), section 8.5.1 PPA Information Object definition mentions that _cbsdReferenceId_ field is a _"List of one or more CBSD Reference IDs in the cluster list of the PPA"_.

So _ppa_record['ppaInfo']['cbsdReferenceId']_ should also be updated if it's an empty array.

### DIFF
--- a/src/harness/sas_test_harness.py
+++ b/src/harness/sas_test_harness.py
@@ -449,11 +449,9 @@ def generatePpaRecords(ppa_records, cbsd_reference_ids):
   for ppa_record, cbsd_reference_id in \
               zip(ppa_records, cbsd_reference_ids):
 
-    # Check cbsdReferenceID in ppaInfo and add cbsdReferenceID if not present.
-    if not ppa_record['ppaInfo'].has_key('cbsdReferenceId'):
-      ppa_record['ppaInfo'].update({'cbsdReferenceId': cbsd_reference_id})
-    elif len(ppa_record['ppaInfo']['cbsdReferenceId']) == 0:
-      ppa_record['ppaInfo'].update({'cbsdReferenceId': cbsd_reference_id})    
+    # Update cbsdReferenceID in ppaInfo
+    ppa_record['ppaInfo'].update({'cbsdReferenceId': cbsd_reference_id})
+    
     ppa_records_list.append(ppa_record)
 
   return ppa_records_list
--- a/src/harness/sas_test_harness.py
+++ b/src/harness/sas_test_harness.py
@@ -452,6 +452,8 @@ def generatePpaRecords(ppa_records, cbsd_reference_ids):
     # Check cbsdReferenceID in ppaInfo and add cbsdReferenceID if not present.
     if not ppa_record['ppaInfo'].has_key('cbsdReferenceId'):
       ppa_record['ppaInfo'].update({'cbsdReferenceId': cbsd_reference_id})
+    elif len(ppa_record['ppaInfo']['cbsdReferenceId']) == 0:
+      ppa_record['ppaInfo'].update({'cbsdReferenceId': cbsd_reference_id})    
     ppa_records_list.append(ppa_record)
 
   return ppa_records_list


kate-harrison:
Our testing API specification explicitly allows the cluster list of an injected PPA to be empty, so this is working as intended. The SAS is not expected to infer that any particular CBSD is on the cluster list, as your PR appears to imply. Please let me know if I've misinterpreted something, though -- I'm working from my phone in the early morning and hence less observant than usual :)

diptivete:
Hi @kate-harrison, 
Thanks for your quick response :)

Yes, testing API spec allows for empty cluster list of an injected PPA for SIQ & GRA related functional tests to verify certain functionality as these groups are testing CBSD <-> SAS protocol.

But FAD testcases verify SAS <-> SAS protocol defined in [n.11] which is WINNF-TS-0096 SAS-SAS protocol spec, so they should follow the rule defined in protocol in our opinion and also like mentioned in test API spec in section 7.2.2 table 7.2-1 SAS-SAS Full Activity Dump Test Characteristics.

cbsdReferenceId field is currently empty in ppaInfo object of ppaRecords that is part of sasTestHarnessDumpRecords in FAD.2's default config where the test verifies that SAS UUT can successfully request a full activity dump from peer SAS with 1 PPA record & 1 ESC sensor record and utilize the retrieved data.

Searching for usages of generatePpaRecords() in test harness repo yields only 2 results WINNF_FT_S_FAD_testcase.py#L642 and full_activity_dump_integration_test.py#L62 so it makes sense to actually always populate cbsdReferenceId in generatePpaRecords() function but instead it currently checks for presence of the key to make an update instead of checking for an empty array.

Approach used in this PR or one mentioned below would fix the problem in FAD.2's default config that we noticed.

    # Check cbsdReferenceID in ppaInfo and add cbsdReferenceID if not present.
    if not ppa_record['ppaInfo']['cbsdReferenceId']:
      ppa_record['ppaInfo'].update({'cbsdReferenceId': cbsd_reference_id})
    ppa_records_list.append(ppa_record)

  "sasTestHarnessDumpRecords": {
…
    "ppaRecords": [
…
        "ppaInfo": {
          "ppaRegionType": "URBAN",
...
          "cbsdReferenceId": [],
          "palId": [
            "pal/11-2017/20041084200/1"
          ]
        },

syandun:
In addition, CBRS Operational and Functional Requirements WINNF-TS-0112-V1.6.0 document, specifies requirements on how a SAS is to create a PPA. See: R1-SPU-01 and R1-SPU-02.
The first requirement for a PPA boundary creation, according to R1-SPU-02:
"The SAS shall use: 1) the locations of the CBSDs on the Cluster List".

Thus having a PPA without a list of cbsdReferenceId's or an empty list is incorrect.

In this case the TestHarness is the "PPA Managing SAS" so it should follow the same rules to create PPAs as any other SAS.

We believe the intention to follow this logic was intended due to the comment in the code:
"# Check cbsdReferenceID in ppaInfo and add cbsdReferenceID if not present."
But it may have been missed that the cbsdReferenceId is an array that could be empty, which should be treated the same as when the actual array is not present.

src/harness/sas_test_harness.py
```diff
@@ -452,6 +452,8 @@ def generatePpaRecords(ppa_records, cbsd_reference_ids):
     # Check cbsdReferenceID in ppaInfo and add cbsdReferenceID if not present.
     if not ppa_record['ppaInfo'].has_key('cbsdReferenceId'):
```
kate-harrison: It seems like there is no case in which we want to keep the cluster list (if present) which is already in the ppa_record, so perhaps these four lines should actually just be 

```ppa_record['ppaInfo'].update({'cbsdReferenceId': cbsd_reference_id})```

This would also be more consistent with the documentation for this function, and with how someone would naturally expect this function to work. What do you think?

diptivete: I agree with your explanation. 
Always updating the ppa cluster list in this function is more consistent with spec.
Thank you Kate for checking this!

nirajankeybridge:
In FAD.2 default.config, it will only change one line i.e. for the _peer SAS_, instead of cbsdReferenceId": [] being empty array as it is now, it will populate the cluster list of the PPA. As Kate noted:

> The key change here is to actually populate the cluster list of the PPA in the peer SAS, not to add CBSDs which were critical to the test to that cluster list.

Following is an excerpt of FAD.2 default.config  with affected line in bold:
"ppaRecords": [
…
"ppaInfo": {
"ppaRegionType": "URBAN",
...
**"cbsdReferenceId": []**,
"palId": [
"pal/11-2017/20041084200/1"
]
},

src/harness/sas_test_harness.py
nirajankeybridge: Thanks Kate, Dipti. I've added a commit to the PR reflecting the suggestion. Please review:
https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/859/commits/e4e157f3b8fc9f29cb6fbb283acc3a7cbb1212e1

kate-harrison:
I just talked to Javi offline and he says he's OK :) Thanks!

PR #860, opened 2018-11-19
small fix of move list when CBSD position is at same location of the protected point

redtechnologies:
fix move list when CBSD position is at same location of the protected point, the calculation of pathloss return a value(0), so for this case we form an array(with same length as reliabilities) of interferencex of the same value.

### DIFF
--- a/src/harness/reference_models/dpa/move_list.py
+++ b/src/harness/reference_models/dpa/move_list.py
@@ -252,8 +252,6 @@ def computeInterference(grant, constraint, inc_ant_height, num_iteration, dpa_ty
 
   # Calculate the interference contributions
   interf = eirp_cbsd - path_loss
-  if np.isscalar(interf):
-     interf = np.full(num_iteration + 1, interf)
   median_interf = interf[-1]      # last element is the median interference
   K_interf = interf[:-1]          # first 'K' interference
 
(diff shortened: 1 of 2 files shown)


RedaHossino:
Serge, could you check if you agree with this small fix, or if you have another solution.

src/harness/reference_models/dpa/move_list.py
```diff
@@ -252,6 +252,8 @@ def computeInterference(grant, constraint, inc_ant_height, num_iteration, dpa_ty
 
   # Calculate the interference contributions
   interf = eirp_cbsd - path_loss
+  if np.isscalar(interf):
```
sbdt-google: I dont think it works, because above we have:
  `path_loss = np.array(results.db_loss)`

And if results.db_loss is scalar 0, then path_loss is an 'array scalar', not a scalar.
Try this in ipython for example:
```
In [25]: a= 0

In [26]: np.array(a)
Out[26]: array(0)

In [27]: np.isscalar(np.array(0))
Out[27]: False
```
So there are 2 easy solution to the issue:
1) change above in line 228:
```
   if np.isscalar(results.db_loss):
     path_loss = np.full(num_iteration + 1, results.db_loss)
   else:
     path_loss = np.array(results.db_loss)
``` 
2) fix it in wf_itm.py directly, which is more clean and robust on other client code using the lib:
```
  # Case of same points
  if (lat_cbsd == lat_rx and lon_cbsd == lon_rx):
    return _PropagResult(
        db_loss = 0 if np.isscalar(reliability) else [0] * len(reliability),
        incidence_angles = _IncidenceAngles(0,0,0,0),
        internals = None)  
```

RedaHossino: Serga, thank for your usefull comment!, I updated it, please check.

PR #864, opened 2018-12-06
Fix DPA builder to work with single point in one DPA area.

sbdt-google:
The check routine will raise exception as they assume a MultiPoint.

Note: this is usually not the case that one use a single point for a DPA, except for testing purpose. 
This fix was requested by ITS so they can more easily debug which point create some issue they see.


Thao-Nguyen:
Looks good! It works with a single point.

sbdt-google:
Thanks Thao for the review. Will wait to get another reviewer as well from SAS admins side.

sbdt-google:
Thanks to all - submitting.

PR #868, opened 2019-01-14
Fix PPA creation code for proper hole removal.

sergebdt:
The code was removing holes afer clipping with census data.
But doing this means that any small census data holes will be
removed, and the final PPA zone would be "bigger" than the census area.

Fixing this by reverting the process:
 - first we remove small holes
 - then we clip with census data
This is more in line with the spec as well (R2-PAL-04).


sbdt-google:
Has been approved by Takashi-san by email.

PR #869, opened 2019-01-15
DPA: properly support grants within  DPA

sbdt-google:
Simply make sure that the "inside DPA" grants are included in the proper move list
after the regular processing:
  - for inband channel, include all the grant which overlaps
  - for OOB, include all the grants 

This method insures that we would get same move list results as if we were adding 
protection points exactly under each CBSD. 

Note that initial implementation was proposing to first filter out the grants inside the DPA and then run
the move list on the remaining outside grants. This is somewhat a bit more optimal in theory, 
although we do not expect a major difference in practice. It has the disadvantage of changing 
a bit more the philosophy of the move list algorithm, and thus requires handling a special case.

Adding unit test on the logic implemented. 
And for sake of improving DPA unit test, added some more unit test which was part of
the dpa_mgr_example.py

### DIFF
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -263,8 +263,9 @@ class Dpa(object):
     # add them into move list for sure later on.
     inside_grants = set()
     if self.geometry and not isinstance(self.geometry, sgeo.Point):
+      dpa_geometry = self.geometry.buffer(1e-6)
       inside_grants = set(g for g in self._grants
-                          if sgeo.Point(g.longitude, g.latitude).within(self.geometry))
+                          if sgeo.Point(g.longitude, g.latitude).within(dpa_geometry))
 
     for chan_idx, (low_freq, high_freq) in enumerate(self._channels):
       moveListConstraint = functools.partial(
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -263,9 +263,8 @@ class Dpa(object):
     # add them into move list for sure later on.
     inside_grants = set()
     if self.geometry and not isinstance(self.geometry, sgeo.Point):
-      dpa_geometry = self.geometry.buffer(1e-6)
       inside_grants = set(g for g in self._grants
-                          if sgeo.Point(g.longitude, g.latitude).within(dpa_geometry))
+                          if sgeo.Point(g.longitude, g.latitude).intersects(self.geometry))
 
     for chan_idx, (low_freq, high_freq) in enumerate(self._channels):
       moveListConstraint = functools.partial(
--- a/src/harness/reference_models/dpa/dpa_mgr.py
+++ b/src/harness/reference_models/dpa/dpa_mgr.py
@@ -45,6 +45,7 @@ import os
 import logging
 
 import numpy as np
+import shapely.geometry as sgeo
 
 from reference_models.geo import zones
 from reference_models.common import data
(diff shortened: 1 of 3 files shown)


sbdt-google:
Note that the implementation and description has changed slightly from first proposal

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -254,6 +259,13 @@ def ComputeMoveLists(self):
     logging.debug('  protected points: %s', self.protected_points)
     pool = mpool.Pool()
     self.ResetLists()
+    # Detect the inside "inside grants", which will allow to
+    # add them into move list for sure later on.
+    inside_grants = set()
+    if self.geometry and not isinstance(self.geometry, sgeo.Point):
+      inside_grants = set(g for g in self._grants
+                          if sgeo.Point(g.longitude, g.latitude).within(self.geometry))
```
RedaHossino: it seems that this code will not handle the case when the CBSD is on one edge of the polygon.

object.within(other)¶
Returns True if the object’s boundary and interior intersect only with the interior of the other.

sbdt-google: Since the rest of the code distribute protection points along the edge, then you are right: we should consider the edge as being inside.
I have updated the code slightly to test for both inside and on the boundary, by using intersects() instead of within().

src/harness/reference_models/dpa/dpa_mgr.py
```diff
@@ -273,6 +285,10 @@ def ComputeMoveLists(self):
       # Combine the individual point move lists
       move_list = set().union(*move_list)
       nbor_list = set().union(*nbor_list)
+      include_grants = ml.filterGrantsForFreqRange(
```
nasrinivasa: Serge,
    Is this being done for all the Channels or just the OOB channels?

sbdt-google: This is done for all the channels. Otherwise you may end up with CBSD inside the DPA which are not in the move list.
This filterGrantsForFreqRange() function returns only the grant overlapping with the protection channel when inband (3550 < dpa_chan < 3650) and all of them when OOB (dpa_chan < 3550).

nasrinivasa: If we are not pre-filtering the Grants, then for the grants overlapping with the protection channel, whether inside or outside, will already be on the ML as they do not have a 0km neighborhood distance.

sbdt-google: I am sorry I do not fully follow you.
A grant inside the DPA can have a neighborhood distance of 0km (like the CatA OOB currently) and would not be detected as part of the move list in the prior implementation, because a move list is the aggregation of all the indivcidual point move list. 
This PR fixes this by adding all those grants to the move list.

Similarly a grant with a neighborhood bigger than 0km (for example all others, Cat A and CatB inband, CatB OOB) could have a big pathloss to the nearby protection points (think about China Lake with mountains where the CBSD is in a valley and the protected point outside the valley), and NOT BE in the move list as a consequence.
This PR fixes this by making sure that they are all in the move list, like if there was a protected point right under the CBSD.

Let me know if I misunderstood your comment.

sbdt-google:
Thanks.  I will wait for a second approval from Navini before submitting.
If you have time, could you also have a look at the other PR ? I think we need 2 approvals at this stage before submittng, and the PPA issue is being reviewed by Takashi, but we dont have a second reviewer as far as I know.

src/harness/reference_models/dpa/dpa_mgr.py
nasrinivasa: Serge,
   That is a good point. I agree with you that this would cover all the cases and not leave anything to be dependent on the interference to be large when CBSDs are within the DPA and hence will automatically be put on the ML.
I agree and approve the changes

sbdt-google:
Thanks to both for quick turn around. Submitting now that we have 2 approval

PR #870, opened 2019-01-18
DPA: Fix the CheckInterference loggers for "inside" grants.

sbdt-google:
Previous PR has introduced special code to include all inside grants
in move list and neighbor list.
Unfortunately this let the logger used for analysing DPA margin failures
not perfectly in line. The SAS_UUT active list of grants (reported to
the log only) would not show some of the inside grants.
This PR fixes that.

Note that this PR has no effect whatsoever on the DPA operation, just on
the logging that is used for post analysis (only when required because
of a failure), possibly with provided script:
tools/studies/dpa_margin_sim.py.


sbdt-google:
Thanks, submitted.

PR #872, opened 2019-02-05
Adding Retries in case of Curl Error

sushantparkhi:


src/harness/request_handler.py
```diff
@@ -127,12 +131,23 @@ def _Request(url, request, config, is_post_method):
...
+    except pycurl.error as e:
+      # e contains a tuple (libcurl_error_code, string_description).
+      # See https://curl.haxx.se/libcurl/c/libcurl-errors.html
+      error_message += ' %s\n' % str(CurlError(e.args[1], e.args[0]))
+      time.sleep(REQUEST_ATTEMPT_DELAY_SECOND)
+    except Exception as e:
+      error_message += ' %s\n' % str(e)
+      time.sleep(REQUEST_ATTEMPT_DELAY_SECOND)
+  if error_message:
+    raise Exception(error_message.strip())
```
kate-harrison: Can you please raise the exception/error which was caught? I'm worried about interference with things like [this](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/sas_testcase.py#L239).

Unfortunately saving `e` and re-raising (i.e. `raise e`) interferes with the traceback, so the preferred option is to use just `raise` from inside the `except` since this gives the right behavior (example [here](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/WINNF_FT_S_SIQ_testcase.py#L1016)).

src/harness/request_handler.py
```diff
@@ -127,12 +131,23 @@ def _Request(url, request, config, is_post_method):
...
+      # e contains a tuple (libcurl_error_code, string_description).
+      # See https://curl.haxx.se/libcurl/c/libcurl-errors.html
+      error_message += ' %s\n' % str(CurlError(e.args[1], e.args[0]))
+      time.sleep(REQUEST_ATTEMPT_DELAY_SECOND)
+    except Exception as e:
+      error_message += ' %s\n' % str(e)
+      time.sleep(REQUEST_ATTEMPT_DELAY_SECOND)
+  if error_message:
+    logging.error(error_message.strip())
+    raise
```
kate-harrison: Did you run this? I think you need to call `raise` from within the `except`. Here's an experiment I did:

```
In [5]: try:
   ...:     raise ImportError("test")
   ...: except Exception as e:
   ...:     print "Inside the except: ", e
   ...: raise
   ...: 
   ...: 
Inside the except:  test
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-7fa1da41e333> in <module>()
      3 except Exception as e:
      4     print "Inside the except: ", e
----> 5 raise
      6 

TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
```

kate-harrison: Strange -- I'm using 2.7.15 (from iPython) and I wouldn't have expected such a large difference between minor versions.

Depending on where you look, our repo says we use 2.7 [here](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System) or 2.7.12 [here](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/tree/master/src/harness/). I'm not sure what flavor ITS is using.

sushantparkhi: Tried on Python 2.7.15
`Python 2.7.15 (default, Aug 17 2018, 22:39:05)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin`

I am getting the same output 😬

sushantparkhi: No, it is not, maybe try creating a py file? If it is still not working, we may have to think about another logic to implement it.

kate-harrison: Hmm, very strange! That worked, and I ran an example test case (REG_6) to double-check. Thanks for the tip and for the PR itself!

src/harness/request_handler.py
```diff
@@ -127,12 +131,23 @@ def _Request(url, request, config, is_post_method):
...
+    try:
+      conn.perform()
+      error_message = ''
+      break
+    except pycurl.error as e:
+      # e contains a tuple (libcurl_error_code, string_description).
+      # See https://curl.haxx.se/libcurl/c/libcurl-errors.html
+      error_message += ' %s\n' % str(CurlError(e.args[1], e.args[0]))
+      time.sleep(REQUEST_ATTEMPT_DELAY_SECOND)
+    except Exception as e:
```
RedaHossino: Don't you think that it is good to log the error(as info or warning) even if the connection works the next try?
I think that may help for troubleshooting the connectivity problems.

kate-harrison: +1 This would be very helpful as it would allow ITS to keep track of how often this is happening and, if needed, to investigate alternative/additional remedies for this issue.

sushantparkhi:
Thanks for the quick review 😄. Should I wait for @RedaHossino as well to approve it, before merging it to master?

kate-harrison:
> Thanks for the quick review . Should I wait for @RedaHossino as well to approve it, before merging it to master?

I'd suggest waiting ~30 minutes and then submitting since (1) you addressed his comment, (2) he's hopefully not working right now since it's 9pm for him, and (3) I'd really like to get this submitted sooner rather than later so we can unblock ITS. If he brings a comment during WF V&V, we can all agree ahead of time not to be upset with him for it since we're submitting "early" :)

PR #873, opened 2019-02-14
Study: Adding the missing East CBSD deployment model.

sbdt-google:
As provided by Michael Souryal from NITS, based on the
Matlab deployement model of Nick Delasorte from NTIA.

Note that these have no effect whatsoever on test harness.


sbdt-google:
Thanks Michael.
On recommendation of Kate, I moved all your deployment models in a top level data/research/ directory instead of keeping them in the reference_models/tools/studies/data.
This will make more clear that they are not used by the test harness per say, and is a better location for any research oriented scripts to use.

Issue #874, opened 2019-03-29
Changes needed for MCP to support DPA OOB check


jcereced:
some changes are needed in order for MCP to support checking the OOB component on DPAs

PR #875, opened 2019-03-30
check for always activated DPA frequency range

arhannan:
Check for always activated DPA frequency range. Activation is not triggered for such DPAs


arhannan:
This addresses Issue #874 . Thanks

PR #876, opened 2019-04-02
Implement PPR.3

kate-harrison:
This PR implements the new PPR.3 test case which was approved last week in WG4. Although it makes changes to the DomainProxy object, the changes are strictly additions and hence no other test cases are affected.

### DIFF
--- a/src/harness/test_harness_objects.py
+++ b/src/harness/test_harness_objects.py
@@ -227,6 +227,9 @@ class DomainProxy(object):
         data that need to be preloaded into SAS (each of which is a dictionary).
         The dictionary is a RegistrationRequest object, the fccId and
         cbsdSerialNumber fields are required, other fields are optional.
+
+    Returns
+      ppa_record: the input PPA record with the cluster list added.
     """
     # Checking if the number of registration requests matches number of grant requests.
     # There should be exactly one grant request per registration request.
--- a/src/harness/test_harness_objects.py
+++ b/src/harness/test_harness_objects.py
@@ -14,6 +14,7 @@
 
 """Implementation of multiple objects (Grant, Cbsd and DomainProxy).
    Mainly used in MCP and related test cases."""
+import copy
 import common_strings
 import logging
 import sas
@@ -230,7 +231,10 @@ class DomainProxy(object):
         cbsdSerialNumber fields are required, other fields are optional.
 
     Returns
-      ppa_record: the input PPA record with the cluster list added.
+      ppa_record_with_cbsd_ids: the input PPA record with the cluster list (of
+        CBSD IDs) added.
+      ppa_record_with_reference_ids: the input PPA record with the cluster list
+        (of CBSD reference IDs) added.
     """
     # Checking if the number of registration requests matches number of grant requests.
     # There should be exactly one grant request per registration request.
@@ -248,10 +252,12 @@ class DomainProxy(object):
     cbsd_id_cluster_list = [cbsd_ids[i] for i in cluster_list if cbsd_ids[i]]
     ppa_record['ppaInfo']['cbsdReferenceId'] = cbsd_id_cluster_list
     self.testcase._sas_admin.InjectZoneData({'record': ppa_record})
+    ppa_record_with_cbsd_ids = ppa_record
     # But the reference model will be looking to compare with CBSD reference
     # IDs, which are not necessarily the same as the CBSD IDs. So now that the
     # PPA has been injected, we need to put the CBSD reference IDs into the PPA
     # record so that the reference model will do the right thing.
+    ppa_record_with_reference_ids = copy.deepcopy(ppa_record)
     def makeReferenceId(reg_request):
       return 'cbsd/' + generateCbsdReferenceId(reg_request['fccId'],
                                                reg_request['cbsdSerialNumber'])
@@ -260,8 +266,9 @@ class DomainProxy(object):
       if not cbsd_ids[i]:
         continue  # Skip if not registered.
       reference_id_cluster_list.append(makeReferenceId(registration_requests[i]))
-    ppa_record['ppaInfo']['cbsdReferenceId'] = reference_id_cluster_list
-    # END PPA-specific logic.
+    ppa_record_with_reference_ids['ppaInfo'][
+        'cbsdReferenceId'] = reference_id_cluster_list
+    # END PPA-specific logic. (Note: return statement also added.)
 
     # Copy the cbsdId from Registration response to grant requests, omitting
     # grant requests for CBSDs which were not successfully registered.
@@ -295,7 +302,7 @@ class DomainProxy(object):
                            [grant_response['grantId']], [grant_request])
         self.cbsd_objects[cbsd_id] = cbsd_object
 
-    return ppa_record
+    return ppa_record_with_cbsd_ids, ppa_record_with_reference_ids
 
 
   def _mergeConditionals(self, registration_request, conditionals):
--- a/src/harness/test_harness_objects.py
+++ b/src/harness/test_harness_objects.py
@@ -17,6 +17,7 @@
 import common_strings
 import logging
 import sas
+from sas_test_harness import generateCbsdReferenceId
 import math
 from common_types import ResponseCodes
 
@@ -244,9 +245,22 @@ class DomainProxy(object):
     # BEGIN PPA-specific logic.
     # Form the cluster list from the CBSD IDs corresponding to the input
     # indices, taking care to only include those which were actually registered.
-    actual_cluster_list = [cbsd_ids[i] for i in cluster_list if cbsd_ids[i]]
-    ppa_record['ppaInfo']['cbsdReferenceId'] = actual_cluster_list
+    cbsd_id_cluster_list = [cbsd_ids[i] for i in cluster_list if cbsd_ids[i]]
+    ppa_record['ppaInfo']['cbsdReferenceId'] = cbsd_id_cluster_list
     self.testcase._sas_admin.InjectZoneData({'record': ppa_record})
+    # But the reference model will be looking to compare with CBSD reference
+    # IDs, which are not necessarily the same as the CBSD IDs. So now that the
+    # PPA has been injected, we need to put the CBSD reference IDs into the PPA
+    # record so that the reference model will do the right thing.
+    def makeReferenceId(reg_request):
+      return 'cbsd/' + generateCbsdReferenceId(reg_request['fccId'],
+                                               reg_request['cbsdSerialNumber'])
+    reference_id_cluster_list = []
+    for i in cluster_list:
+      if not cbsd_ids[i]:
+        continue  # Skip if not registered.
+      reference_id_cluster_list.append(makeReferenceId(registration_requests[i]))
+    ppa_record['ppaInfo']['cbsdReferenceId'] = reference_id_cluster_list
     # END PPA-specific logic.
 
     # Copy the cbsdId from Registration response to grant requests, omitting
--- a/src/harness/test_harness_objects.py
+++ b/src/harness/test_harness_objects.py
@@ -215,7 +215,7 @@ class DomainProxy(object):
                                            cluster_list,
                                            conditional_registration_data=None):
     """Construct CBSD object based on the result of registration and
-    grant requests.
+    grant requests. Registers the CBSDs, injects the PPA, and requests grants.
 
     Args
       registration_requests: A list of dictionary elements, where each list
--- a/src/harness/test_harness_objects.py
+++ b/src/harness/test_harness_objects.py
@@ -206,6 +206,81 @@ class DomainProxy(object):
                            [grant_response['grantId']], [grant_request])
         self.cbsd_objects[cbsd_id] = cbsd_object
 
+  def registerCbsdsAndRequestGrantsWithPpa(self,
+                                           registration_requests,
+                                           grant_requests,
+                                           ppa_record,
+                                           cluster_list,
+                                           conditional_registration_data=None):
+    """Construct CBSD object based on the result of registration and
+    grant requests.
+
+    Args
+      registration_requests: A list of dictionary elements, where each list
+        element is an individual CBSD registration request (which is itself a
+        dictionary).
+      grant_requests: A list of dictionary elements, where each list element is
+        an individual CBSD grant request (which is itself a dictionary).
+      ppa_record: the PPA record to inject after registration.
+      cluster_list: indices indicating which CBSDs are on the cluster list.
+      conditional_registration_data: A list of individual CBSD registration
+        data that need to be preloaded into SAS (each of which is a dictionary).
+        The dictionary is a RegistrationRequest object, the fccId and
+        cbsdSerialNumber fields are required, other fields are optional.
+    """
+    # Checking if the number of registration requests matches number of grant requests.
+    # There should be exactly one grant request per registration request.
+    self.testcase.assertEqual(len(grant_requests), len(registration_requests))
+    try:
+      cbsd_ids = self._tryRegistrationWithMaximumBatchSize(
+          registration_requests, conditional_registration_data)
+    except Exception:
+      logging.error(common_strings.EXPECTED_SUCCESSFUL_REGISTRATION)
+      raise
+
+    # BEGIN PPA-specific logic.
+    # Form the cluster list from the CBSD IDs corresponding to the input
+    # indices, taking care to only include those which were actually registered.
+    actual_cluster_list = [cbsd_ids[i] for i in cluster_list if cbsd_ids[i]]
+    ppa_record['ppaInfo']['cbsdReferenceId'] = actual_cluster_list
+    self.testcase._sas_admin.InjectZoneData({'record': ppa_record})
+    # END PPA-specific logic.
+
+    # Copy the cbsdId from Registration response to grant requests, omitting
+    # grant requests for CBSDs which were not successfully registered.
+    actual_grant_requests = []
+    for cbsd_id, grant_request in zip(cbsd_ids, grant_requests):
+      if cbsd_id:  # cbsd_id is None if registration was unsuccessful.
+        grant_request['cbsdId'] = cbsd_id
+        actual_grant_requests.append(grant_request)
+
+    # Perform grant operation.
+    grant_responses = self._grantRequestWithMaximumBatchSize(actual_grant_requests)
+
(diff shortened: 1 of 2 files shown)


src/harness/testcases/WINNF_FT_S_PPR_testcase.py
```diff
@@ -401,3 +409,257 @@ def test_WINNF_FT_S_PPR_2(self, config_filename):
...
+    }
+
+    # One PPA in SAS UUT.
+    pal_low_frequency = 3620000000
+    pal_high_frequency = 3630000000
+    pal_record_0 = json.load(
+        open(os.path.join('testcases', 'testdata', 'pal_record_0.json')))
+    ppa_record_0 = json.load(
+        open(os.path.join('testcases', 'testdata', 'ppa_record_0.json')))
+    ppa_record_0, pal_records_0 = makePpaAndPalRecordsConsistent(
```
nasrinivasa: The fips code entry is missing for this record (pal_record_0)

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
```diff
@@ -401,3 +409,257 @@ def test_WINNF_FT_S_PPR_2(self, config_filename):
...
+    # Trigger CPAS in SAS UUT, and wait until completion.
+    logging.info('Step 8: trigger CPAS')
+    self.cpas = self.cpas_executor.submit(
+        self.TriggerDailyActivitiesImmediatelyAndWaitUntilComplete)
+
+    logging.info('Step 9: execute IAP reference model')
+    # Pre-IAP filtering.
+    pre_iap_filtering.preIapReferenceModel(self.protected_entity_records,
+                                           self.sas_uut_fad,
+                                           self.test_harness_fads)
```
RedaHossino: there is one problem in the file data.py in the method getCbsdsNotPartOfPpaCluster.
as the id of CBSD records in the dump start by "cbsd/" but  the 'cbsdReferenceId' doesn't, the follwing condition will not be satisfied:
  if cbsd['id'] not in ppa_record['ppaInfo']['cbsdReferenceId']:

so during the IAP prefiltering, the reference model will not take the CBSD from SAS UUT dump as part of the PPA cluster list for the PPA of SAS UUT.

right?

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
kate-harrison: It's included (see [here](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/master/src/harness/testcases/testdata/pal_record_0.json)), I just didn't need to change it from the default so I didn't set it here.

nasrinivasa: Yup see it now. Thanks

jcereced:
When I try to run the default config I'm getting this error when it's doing the reference model, is it because of something Reda mentions above...or maybe I'm doing something wrong?




util.py:79: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
testcases/WINNF_FT_S_PPR_testcase.py:651: in test_WINNF_FT_S_PPR_3
    self.test_harness_fads)
reference_models/pre_iap_filtering/pre_iap_filtering.py:63: in preIapReferenceModel
    list_of_fss_neighboring_gwbl)
reference_models/pre_iap_filtering/zone_purge.py:62: in zonePurgeReferenceModel
    cbsds_within_ppa = pre_iap_util.getCbsdsWithinPolygon(cbsds, ppa_record['zone'])
reference_models/pre_iap_filtering/pre_iap_util.py:71: in getCbsdsWithinPolygon
    polygon = utils.ToShapely(polygon['features'][0]['geometry'])
reference_models/geo/utils.py:168: in ToShapely
    return _GeoJsonToShapelyGeometry(geometry)
reference_models/geo/utils.py:145: in _GeoJsonToShapelyGeometry
    geometry = sgeo.shape(geometry)
/opt/harness_env/lib/python2.7/site-packages/shapely/geometry/geo.py:35: in shape
    return Polygon(ob["coordinates"][0], ob["coordinates"][1:])
/opt/harness_env/lib/python2.7/site-packages/shapely/geometry/polygon.py:240: in __init__
    ret = geos_polygon_from_py(shell, holes)
/opt/harness_env/lib/python2.7/site-packages/shapely/geometry/polygon.py:494: in geos_polygon_from_py
    ret = geos_linearring_from_py(shell)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   TypeError: object of type 'float' has no len()

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
```diff
@@ -401,3 +409,290 @@ def test_WINNF_FT_S_PPR_2(self, config_filename):
...
+    ppa_record_0 = json.load(
+        open(os.path.join('testcases', 'testdata', 'ppa_record_0.json')))
+    ppa_record_0['zone']['features'][0]['geometry']['coordinates'] = [[[
+        -97.155, 38.75
+    ], [-97.155, 38.85], [-97.165, 38.85], [-97.165, 38.75], [-97.155, 38.75]]]
+    ppa_record_0, pal_records_0 = makePpaAndPalRecordsConsistent(
+        ppa_record_0, [pal_record_0], pal_low_frequency, pal_high_frequency,
+        'test_user_1')
+    device_b['userId'] = 'test_user_1'
+    ppa_cluster_list = [0]  # The first device is device B.
```
jcereced: Assuming both deviceB and deviceD belong to the SUUT PPA and will get PAL grants we would need to add deviceD to the clusterlist, right? or are we trying to have deviceD fail to get a grant?

jcereced:
I thought I added a comment but can't find it.

For the two CBSDs that the SUUT is managing deviceB and deviceD, do we want both of them to get PAL grants?  If so we would need to add deviceD to the clusterList in the config

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
```diff
@@ -511,24 +515,34 @@ def generate_PPR_3_default_config(self, filename):
...
+    # the CBSDs themselves.
+    cbsd_reference_id_a = 'cbsd/' + generateCbsdReferenceId(
+        device_a['fccId'], device_a['cbsdSerialNumber'])
+    cbsd_reference_id_c = 'cbsd/' + generateCbsdReferenceId(
+        device_c['fccId'], device_c['cbsdSerialNumber'])
+    cbsd_reference_id_e = 'cbsd/' + generateCbsdReferenceId(
+        device_e['fccId'], device_e['cbsdSerialNumber'])
+    cbsd_reference_ids = [[
+        cbsd_reference_id_a, cbsd_reference_id_c, cbsd_reference_id_e
+    ]]
```
RedaHossino: this will not fix the problem :(
the main Id problem is between the SAS UUT injected PPA reference IDs and the ID of CBSD records in SAS UUT dump.
asthe reference model take the injected SAS UUT PPA record as protected intity, and the cbsd reference ids of this PPA don't contain 'cbsd/', butwhen TH pull the dump of SAS UUT, the id of CBSD records contain '/cbsd', so it cann't do the link between them.

I think we need to remove this '/cbsd' either from the cbsd records of the dump or from cbsd['id'] in the if statement of getCbsdsNotPartOfPpaCluster(data.py) :/

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
```diff
@@ -413,14 +413,13 @@ def generate_PPR_3_default_config(self, filename):
     """High-level description of the default config:
 
         SAS UUT has devices B, D; first gets a PAL grant and second is GAA.
-        SAS TH has devices A, C, E, none of which have a PAL grant.
+        SAS TH has devices A, C, E, all of which have a PAL grant.
 
         SAS UUT has one PPA, with device B as the only CBSD.
         SAS TH has one PPA, with only non-granted CBSDs.
```
RedaHossino: I see now your initial point :)
as you mentioned, you wanted the SAS TH grants to be GAA and not part of the PPA.
but as now they are PAL, this comment need an update.

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
RedaHossino: Javi,
as I read in the test comment (line 418):
SAS UUT has one PPA, with device B as the only CBSD.

so I think that  was done on purpose.

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
```diff
@@ -401,3 +409,257 @@ def test_WINNF_FT_S_PPR_2(self, config_filename):
...
+    """
+    # Load Devices
+    device_a = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_a.json')))
+    device_a['installationParam']['latitude'] = 38.842176
+    device_a['installationParam']['longitude'] = -97.092863
+    device_b = json.load(
+        open(os.path.join('testcases', 'testdata', 'device_b.json')))
+    device_b['installationParam']['latitude'] = 38.815291
+    device_b['installationParam']['longitude'] = -97.198805
```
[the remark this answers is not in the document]
RedaHossino: sorry for this coordinate mistake, I didn't check the service area.
thanks

src/harness/test_harness_objects.py
```diff
@@ -206,6 +206,81 @@ def registerCbsdsAndRequestGrants(self,
...
+
+    # Make one CBSD object per successful grant request.
+    # All arrays in the zip() will have the same length based on the filtering
+    # above, but we'll double-check to be sure.
+    self.testcase.assertEqual(len(actual_grant_requests), len(actual_registration_requests))
+    self.testcase.assertEqual(len(actual_grant_requests), len(actual_cbsd_ids))
+    for grant_request, grant_response, \
+        registration_request, cbsd_id in zip(actual_grant_requests,
+                                             grant_responses, actual_registration_requests, actual_cbsd_ids):
+      if grant_response['response']['responseCode'] == ResponseCodes.SUCCESS.value:
```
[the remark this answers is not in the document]
RedaHossino: I agree with that, thanks

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
kate-harrison: Thank you! I'm rushing too much...

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
kate-harrison: It was originally done on purpose, but now it's wrong. Thank you and fixed!

src/harness/testcases/WINNF_FT_S_PPR_testcase.py
kate-harrison: Ok, this time I think I properly fixed it :) The issue seems to have been in both SAS UUT and the peer SAS TH, and as you correctly pointed out, I only fixed the second one yesterday. This time I fixed the first one by making some changes in `test_harness_objects.py` (please double-check).

I double-checked myself by adding the following code to the end of the `getCbsdsNotPartOfPpaCluster()` function:

```
    logging.info('Examining PPA: %s' % ppa_record)
    logging.info('CBSDs: %s', cbsds)
    logging.info('CBSDs not part of cluster: %s', cbsds_not_part_of_ppa_cluster)
```

and the output showed that the peer SAS PPA has this as its cluster list:
```'cbsd/test_fcc_id_a/d7a9fe1be84243ebdd50c1359cf0630c3d273350', 'cbsd/test_fcc_id_c/33c81df2998eb5d587cea7dede0d35d034e79bcd', 'cbsd/test_fcc_id_e/1a57a6a304f29ccc1a5b814d661a8f8ef235bb9d'```

and although `cbsds` included these three CBSDs, `cbsds_not_part_of_ppa_cluster` was empty.

Similar for the SAS UUT PPA, which has this cluster list: `'cbsd/test_fcc_id_b/b8a0c47f2aed5f2b0222c35cd9b91cff745d0c46', 'cbsd/test_fcc_id_d/1584f2ceda10e1c0f29ad72cccdb339f90aecf40'`

Comment (sushantparkhi) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: Seems like this change does not match with WINNF-TS-0096. In 8.3 CBSD Data Message, the `id` field explicitly specifies to add cbsd/ before $CBSD_REFERENCE_ID but in 8.5.1 PPA Information the CBSDReferenceId  field does not say that.
I agree with @RedaHossino that we may have to modify the `data.py` which might have a bigger impact on other tests or we could just add one function at Line 679 which will remove `cbsd/` from CBSD Record.

Comment (kate-harrison) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: Bah, you're totally right! I took roughly the approach you suggested, but fixed the cluster list of the peer PPA (after writing out the FAD files) rather than removing "cbsd/" from the CBSD IDs.

Ideally we would fix data.py but doing so would be too disruptive at this point, so we have to live with it for now :/ I'll file a bug to track this issue so that we remember to fix it later.

Comment (sushantparkhi) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: I guess we have to do this for SAS UUT FAD as well because when we pull the FAD using 
`self.sas_uut_fad = getFullActivityDumpSasUut(self._sas, self._sas_admin,
                                                 self.fad_cert, self.fad_key)` this might also not have `cbsd/` in Cluster List of PPA but will have `cbsd/$REFERENCE_ID` in CBSD Record?

Comment (kate-harrison) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: I don't think so: the reference model will use the objects in `self.protected_entity_records` when computing protection -- it only looks at the FADs to figure out which CBSDs exist. I already added "cbsd/" to the reference IDs in `registerCbsdsAndRequestGrantsWithPpa()`, and it's that object which is added to `self.protected_entity_records`. Please double-check my work but I think it's correct?

Comment (sushantparkhi) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: Sorry, yes you are right!

Comment (kate-harrison) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: No problem -- I appreciate your very careful review!

Comment (RedaHossino) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: Kate, I think we have the same issue with the method getAuthorizedGrantsFromDomainProxies(data.py).
as a result during the calculationg of  aggregated intereferences of one PPA the reference model will not exclude the PAL grants of this PPA.
could you check that!?

Comment (kate-harrison) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: Oooof, this is a rough one. As far as I can tell, we need to use "cbsd/$REFERENCE_ID" when we call the pre-IAP and IAP ref models, but then we need to use "$CBSD_ID" when we actually do the aggregate interference check (based on the implementation of `getAuthorizedGrantsFromDomainProxies()`).

In the ideal world, we would just go fix these functions to be correct... but doing so would cause quite a stir right now.

So the hacky fix that I can think of is to actually have `registerCbsdsAndRequestGrantsWithPpa()` return two versions of the PPA, one for each ID type. Then we manage the values in `self.protected_entity_records` carefully so that the right version is used at each step.

This is a really ugly fix, but I don't see anything better which doesn't involve touching other code. Do you? And do you think this will be enough to solve the problem?

Comment (RedaHossino) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: yeah, I agree the simplest and most clean scenario is to update the code that impact all the xPR and MCP.
your solution is good if we shouldn't modify the MCP related code.
another idea, 
you don't think that after the CPAS step, in the PPR test code we can delete the 'cbsd/' from the  reference ids of PPAs in protected_entity_records and maybe from other record ids if necessary?

Comment (kate-harrison) [reply src/harness/testcases/WINNF_FT_S_PPR_testcase.py]: If $CBSD_ID == $REFERENCE_ID for all SASes, then your suggestion would work. But since the values *can* be different, I think we have to be a bit more careful. I'll update according to my original suggestion ASAP.

**Comment by kate-harrison** [comment]
Just updated -- please take another look!

File: src/harness/testcases/WINNF_FT_S_PPR_testcase.py [hunk 8]
```diff
@@ -401,3 +409,317 @@ def test_WINNF_FT_S_PPR_2(self, config_filename):
...
+            'palRecords': list,
+            'sasTestHarnessDumpRecords': dict,
+            'sasTestHarnessConfig': dict
+        })
+    self.assertEqual(
+        len(config['sasTestHarnessDumpRecords']['ppaRecords']), 1,
+        'Only one PPA is supported.')
+    # Make sure ID formats are correct.
+    ppa = config['sasTestHarnessDumpRecords']['ppaRecords'][0]
+    for cbsd_ref_id in ppa['ppaInfo']['cbsdReferenceId']:
```
sushantparkhi: Do you think we need an assertion for Length as well for `ppa['ppaInfo']['cbsdReferenceId']`?

kate-harrison: Done

src/harness/test_harness_objects.py
```diff
@@ -206,6 +208,103 @@ def registerCbsdsAndRequestGrants(self,
...
         self.cbsd_objects[cbsd_id] = cbsd_object
 
+  def registerCbsdsAndRequestGrantsWithPpa(self,
+                                           registration_requests,
+                                           grant_requests,
+                                           ppa_record,
+                                           cluster_list,
+                                           conditional_registration_data=None):
+    """Construct CBSD object based on the result of registration and
+    grant requests.
```
sushantparkhi: Maybe updating comment to specify that PPA Record is also Injected?

kate-harrison: Done, thanks!

kate-harrison:
Thank you! I'll wait another hour or so just to give Sushant and Javi some time to make any final comments, but at that time I'll submit and start the WF V&V.

sushantparkhi:
PR looks good to me as well. Thank you Kate for addressing the comments!

kate-harrison:
Thank you very, very much to all reviewers! I'll submit now and start V&V soon.

PR #881, opened 2019-05-07
Updated KML file format specification (1.0.7)

w4je:
Version 1.0.7 file format specification

[DPA KML v1.0.7.pdf](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/files/3152943/DPA.KML.v1.0.7.pdf)


w4je:
I am working on an overview document of all incumbents and wanted to point to the KML file specification in case the reader was interested. I noticed that the one that's there now is outdated. It's fine if you want to wait before posting the latest version. No big deal.

sbdt-google:
You can submit it I think from this PR if you wish.
The pending PR will be submit for the rest of the other stuff, when I get the test-dev GO.

PR #892, opened 2020-11-01
Fix header of the NationWide CSV deployement files.

sbdt-google:
Correctly name each column to something meaningful.


hsuany (approved):
LGTM

PR #896, opened 2021-01-25
xlsx output

glossner:
This adds excel xlsx output. It requires importing xlsxwriter.


sbdt-google (approved):
Tested and verified. LGTM. 
You can submit. 
Thanks

PR #899, opened 2021-07-01
Add v1.09 of P-DPAs.kml.

sbdt-google:
Take this occasion to add minor fix in one "study" file (missing "import os").


w4je (approved):
LGTM

PR #900, opened 2021-07-13
Support counties and update expired dates

RedaHossino:
this branche will be used to create a new test version. this version will be used by ITS as initial step to confirm that the tests, which were used on wave 1, are valid and that they pass successfully on the SASs of wave 2.

the changes are :
- update the expired dates for wave 2.
- add the counties and update the reference model and the tests T0
- remove the census tracts

another related pull request to add counties process&script is on the SAS common data repository.


src/data/README.md
```diff
@@ -3,9 +3,9 @@
 This folder holds all scripts used to retrieve and preprocess data used in the
 SAS reference implementation.
 
-## Geo Data for propagation models and census
+## Geo Data for propagation models and county
 
-The geo data used by propagation models (as well as the census tract data) are stored 
+The geo data used by propagation models are stored 
 in Git LFS (Large File Storage) on the [separate SAS-Data WinnForum repository](https://github.com/Wireless-Innovation-Forum/SAS-Data).
```
sbdt-google, line 9: Correct "SAS-Data" into "Common-Data", and the link

src/data/README.md
```diff
@@ -19,7 +19,6 @@ The process to plug the data into the environment is:
    + create soft links for:
       * `data/geo/ned` -> `SAS_Data/ned`
       * `data/geo/nlcd` -> SAS_Data/nlcd`
-      * `data/census_tracts` -> SAS_Data/census
```
sbdt-google, line 22: Same here:  SAS_Data ==> Common_Data

src/harness/reference_models/geo/README.md
```diff
@@ -9,7 +9,7 @@ models:
  - `vincenty.py`: precise methods for deriving geodesic between two points in the earth
  modeled as ellipsoid
  - `utils.py`: utility routines for computing polygon area, gridding a polygon, etc..
- - `census_tract.py`: Census tract driver to read JSON census tracts geometries.
+ - `county.py`: county driver to read JSON coutnies geometries.
```
sbdt-google, line 12: typo "Coutnies"

sbdt-google (commented):
Few minor comment.  Otherwise seems ok to me.

For the follow up PR in Common-Data, you can keep both driver code and keep the old data.

src/harness/reference_models/ppa/ppa_test.py
```diff
@@ -89,18 +89,18 @@ def test_ClippedPpaByCensus(self):
...
     wf_hybrid.CalcHybridPropagationLoss = testutils.FakePropagationPredictor(
         dist_type='REAL', factor=1.0, offset=(96+30-0.1) - 45.0)
 
     ppa_zone = ppa.PpaCreationModel(TestPpa.devices[1:], TestPpa.pal_records[1:])
     ppa_zone = json.loads(ppa_zone)
 
-    census_zone = utils.ToShapely(
-        drive.census_tract_driver.GetCensusTract('06027000100')
+    county_zone = utils.ToShapely(
+        drive.county_driver.GetCounty('06027000100')
```
diptivete, line 101: Guess fipsCode here should change to first 5 digits '06027' to get corresponding county shape file.

RedaHossino: for this unit test, the name can be any string as long as the file exists in \src\harness\reference_models\ppa\test_data, I don't think it is necessary to have a name  with 5 digits which reflects a real name of a county as it is only a fake name for this unit test, don't you agree?

diptivete: Agree test won't fail as long as file exists in the test_data directory. Don't you think it would be good to stay consistent and use real identifiers even for unit tests? Anyways, I leave the decision to you and other approvers as you see fit.

All other changes look good.

RedaHossino: thank you!  the update is done.

src/harness/reference_models/geo/README.md
RedaHossino: thanks Serge!
I already made the PR with the changes in Common-data, please feel free to check ;)
for your comment about the PR in Common-data ("you can keep both driver code and keep the old data"), did you mean census tracts data&scripts?
I don't know if we need to keep census tracts related data as they exist in the Github history.

diptivete (approved):

src/harness/reference_models/geo/README.md
sbdt-google: I was suggesting to keep the census tract data and reading driver (within Common-Data repo), since it is located in a different repo that is supposed to be of use for all Winnforum projects. SAS is one, 6GHz-AFC is another, ...
Why not let them since they are already there and perhaps of use for one of these future projects

diptivete (approved):

PR #906, opened 2024-02-06
Add calc P.2108

FW-ccaparanga:
Add implementation of P.2108 defined clutter loss to reference_models/propagation

### DIFF
--- a/src/harness/reference_models/propagation/p2108.py
+++ b/src/harness/reference_models/propagation/p2108.py
@@ -1,0 +1,63 @@
+#    Copyright 2024 SAS Project Authors. All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import numpy as np
+
+from reference_models.geo import drive
+from reference_models.geo import vincenty
+
+F_2108 = 3.6  # Desired frequency (GHz)
+P_2108 = 50  # Desired percentage
+sigma_l = 4
+sigma_s = 6
+P2108_LOSS_2KM = 30.50030179
+
+
+def calc_P2108(lat_cbsd, lon_cbsd, height_cbsd,
+               lat_rx, lon_rx, is_height_cbsd_amsl=False):
+  """Implements P.2108 defined clutter loss
+
+  Inputs:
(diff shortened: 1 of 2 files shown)


src/harness/reference_models/propagation/p2108.py
```diff
@@ -0,0 +1,63 @@
...
+from reference_models.geo import vincenty
+
+F_2108 = 3.6  # Desired frequency (GHz)
+P_2108 = 50  # Desired percentage
+sigma_l = 4
+sigma_s = 6
+P2108_LOSS_2KM = 30.50030179
+
+
+def calc_P2108(lat_cbsd, lon_cbsd, height_cbsd,
```
gprillgoogle, line 27: Is there a way to unit test this?

src/harness/reference_models/propagation/p2108.py
```diff
@@ -0,0 +1,63 @@
...
+      return 0
+    elif distance >= 2:
+      return P2108_LOSS_2KM
+    else:
+      L_l = -2 * np.log10(10 ** (-5 * np.log10(F_2108) - 12.5) + 10 ** (-16.5))
+      L_s = 32.98 + 23.9 * np.log10(distance) + 3 * np.log10(F_2108)
+      sigma_cb = np.sqrt((sigma_l ** 2 * 10 ** (-0.2 * L_l) + sigma_s ** 2 * 10 ** (-0.2 * L_s)) / (
+          10 ** (-0.2 * L_l) + 10 ** (-0.2 * L_s)))
+      # Q_inv = norm.ppf(P_2108 / 100)
+      Q_inv = 0
```
gprillgoogle, line 60: If this is 0, remove it and in the equation below remove sigma_db * Q_inv?

src/harness/reference_models/propagation/p2108.py
```diff
@@ -0,0 +1,63 @@
...
+      return -5 * np.log10(10 ** (-0.2 * L_l) + 10 ** (-0.2 * L_s)) - sigma_cb * Q_inv
```
gprillgoogle, line 61: Instead of calculating multiple lines of variables, you can simplify the equation first and then calculate it. For example to a certain degree of precision this is equal to (in C++) the following that lets you have everything on one line:

-5.0 * log10(7.8995e-7 + (1.1754e-7 * pow(distance_meters / 1000.0, -4.78)))

gprillgoogle (changes requested):

src/harness/reference_models/propagation/p2108.py
FW-ccaparanga: Removed Q_inv

src/harness/reference_models/propagation/p2108.py
FW-ccaparanga: I don't think simplifying is necessary here. We prefer the clarity over the simplification. It is nicer to see the full math.

src/harness/reference_models/propagation/p2108.py
FW-ccaparanga: Added p2108_test.py for unit tests

src/harness/reference_models/propagation/p2108.py
gprillgoogle: That is fine with me.

gprillgoogle (approved):

PR #910, opened 2024-03-06
fix Leapday bug - Just changing Year does not work for Feb 29

### LINKED ISSUE #912: Leap day bug - in util2.py changing only the Year does not work for Feb 29
[util2.py](https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/blob/df2c016a1b242590f027e7e25485f80f7c448001/src/harness/util2.py#L49)  increments and decrements just the Year for Start and End dates.

However, this fails for Leap Day (Feb. 29) since this date doesn't exist in the preceding and subsequent years.

Suggesting to increment and decrement the date by 365 days instead using timedelta.

nirajankeybridge:
`util2.py` increments and decrements just the Year for Start and End dates.
However, this fails for Leap Day (Feb. 29) since this date doesn't exist in the preceding and subsequent years.

Suggesting instead to increment and decrement the date by 365 days using timedelta.

fixes #912

### DIFF
--- a/src/harness/util2.py
+++ b/src/harness/util2.py
@@ -18,7 +18,7 @@ from __future__ import division
 from __future__ import print_function
 
 from collections import defaultdict
-from datetime import datetime
+from datetime import datetime, timedelta
 import inspect
 import json
 import os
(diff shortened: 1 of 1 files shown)


mliang-mitre (changes requested):
This fix works.
Suggest defining a variable days_in_year = 365 to avoid using magic number in the code

nirajankeybridge:
Thanks @mliang-mitre 
Implemented your suggestion in the latest commit ( https://github.com/Wireless-Innovation-Forum/Spectrum-Access-System/pull/910/commits/a52a31c8170127b06dab9bd8340a3eeffd87aec2 )

PR #913, opened 2024-03-06
do_Post bug fix and python3 related

mliang-mitre:
1. add requirements.txt for python venv
2. fix a bug in do_Post where a return statement is not indented correctly
3. numpy.int -> int to work with newer version of numpy
4. other changes to fix "empty response" issue when running fake-sas


requirements.txt
```diff
@@ -0,0 +1,28 @@
+attrs==23.2.0
+cffi==1.16.0
```
FW-ccaparanga, line 2: What version of Python was used for generating this? Looks like cffi 1.16.0 dropped support for python 3.7, which is the stated version in the readme

mliang-mitre: Python 3.11
Should I update readme to 3.11? Feels like that's the right thing to do, since 3.7 if end of life.

FW-ccaparanga: Yes, please update the readme. I'll test this branch with python 3.11 to verify

mliang-mitre: I've updated readme. Thank you for reviewing and testing.

FW-ccaparanga (approved):
I did some light validation to test out conda and venv + requirements.txt. Looks good.

mliang-mitre:
Thank you @FW-ccaparanga for reviewing & testing.

PR #918, opened 2024-04-04
PAT2 and PAT8 default.config, using Sony content

mliang-mitre:
Change the content of PAT2 default.config to match Sony provided file pat2.default.20240212-4.config


gprillgoogle:
It seems like a config this long belongs in a config file, not in the testcase file. Why can't we do it that way?

mliang-mitre:
@gprillgoogle good point. added two json files to hold the test data for PAT2 and IPR8

gprillgoogle (approved):

PR #929, opened 2024-05-16
add tranche-1 PAT2 test configs

mliang-mitre:
Move tranche-0 configs to a sub-folder; add tranche-1 configs


nirajankeybridge (approved):
I was able to run PAT.2 with these 12 new T1 configs so looks good.

PR #931, opened 2024-05-17
add tranche-1 ipr8 test configs

mliang-mitre:
a) add tranche-1 ipr8 test configs; b) move tranche-0 test configs to a subfolder


nirajankeybridge (approved):
New T1 configs seem to be read in correctly by the Test Harness.
I ran the new IPR.8 T1 configs against the Fake SAS and the tests ran to completion. 

Following config passed even with Fake SAS:
tranche-1/IPR8-MCBPendleton-2-all-20240508.config

Other 2 configs failed (as expected).

mliang-mitre:
For GB-DPA, not sure why fake-sas passes the test. This is the case with T0 GB-DPA as well

